PR #957, opened 2023-10-09
Feat: wt 1768 service unavailable error view

sharonsheah:
# Summary

https://immutable.atlassian.net/browse/WT-1768
Screenshot 2023-10-09 at 5 21 54 pm


deepti-imx (approved):

packages/checkout/widgets-lib/src/resources/text/textConfig.ts
```diff
@@ -67,6 +67,10 @@ export const text = {
       body: ['You can try again or contact', 'support', 'for help.'],
       actionText: 'Try again',
     },
+    [SharedViews.SERVICE_UNAVAILABLE_ERROR_VIEW]: {
+      heading: 'Service not available in your region',
+      body: 'We’re sorry we cannot provide this service in your region.',
```
ZacharyCouchman, line 72: @sharonsheah @deepti-imx 
Can we make this copy more specific?
For instance, instead of saying 'Service not available', we could at least say which service is not available. "Token swaps are not available in you region"

Also, do we have product direction to say why they are not available in the region, do we specify which regions are supported anywhere?

deepti-imx: > Can we make this copy more specific?
The error view is generic and the copy is provided by legal team (prettified by Cass) but can check with Cass again on the copy.

> Also, do we have product direction to say why they are not available in the region, do we specify which regions are supported anywhere?

We specify the geo-blocked regions in developer documentation.
For end-users, we don't have any documentation yet. It is something we have discussed to support the end-users but not prioritised the docs yet.

sharonsheah: Bringing this conversation to slack to include Harris and Cass https://imtbl.slack.com/archives/C0414KQ5162/p1696904344641679

deepti-imx (approved):

packages/checkout/widgets-lib/src/resources/text/textConfig.ts
```diff
@@ -67,6 +67,10 @@ export const text = {
       body: ['You can try again or contact', 'support', 'for help.'],
       actionText: 'Try again',
     },
+    [SharedViews.SERVICE_UNAVAILABLE_ERROR_VIEW]: {
+      heading: (service: string) => `${service} not available in your region`,
```
ZacharyCouchman, line 71: @sharonsheah @deepti-imx 
This solution means we need to hardcode a string in the props for the view, which also gets passed straight into the rendered text output. I think this what we are trying to get away from with the textConfig in general, so that we can define the text as constant. 
Perhaps you could pass a enum prop into the view which is sued to choose the specific constant text to be rendered? For example:
[SharedViews.SERVICE_UNAVAILABLE_ERROR_VIEW]: {
      swapHeading: 'Swapping is not available in your region',
      bridgeHeading: 'Bridging is not available in your region'
      onrampHeading: ...
 
This also means that we can change each of the service headings specifically for swap / bridge / other service, if we need to without having to decouple from `${service} is not available...`

ZacharyCouchman (approved):

PR #961, opened 2023-10-10
feat: wt 1781 create availability service

sharonsheah:
# Summary

https://immutable.atlassian.net/browse/WT-1781

# Things worth calling out

See comments


packages/checkout/sdk/src/availability/availability.ts
```diff
@@ -0,0 +1,43 @@
...
+  const checkDexAvailability = async (): Promise<boolean> => {
+    let availability;
+    try {
+      const response = await axios.post(`${postEndpoint()}/v1/availability/dex`);
+      if (response.status === 403) {
+        availability = false;
+      } else if (response.status === 204) {
+        availability = true;
+      } else {
+        throw new Error(`Error fetching from api: ${response.status} ${response.statusText}`);
```
sharonsheah, line 32: I've followed and kept consistent with the error that we have in config `makeHttpRequest` because we can't get the `error` like in the `catch` block as this `throw new Error` is in the `try` block.

packages/checkout/sdk/src/availability/availability.ts
```diff
@@ -0,0 +1,43 @@
+import { Environment } from '@imtbl/config';
+import axios from 'axios';
+import { ENV_DEVELOPMENT, IMMUTABLE_API_BASE_URL } from '../types';
+
+type AvailabilityService = (
+  isDevelopment: boolean,
+  isProduction: boolean
+) => {
+  checkDexAvailability: () => Promise<boolean>
+};
```
sharonsheah, line 10: Not very sure about how to define this type..

deepti-imx (approved):

packages/checkout/sdk/src/availability/availability.ts
```diff
@@ -0,0 +1,38 @@
...
+      const response = await axios.post(`${postEndpoint()}/v1/availability/checkout/swap`);
+      if (response.status === 403) {
+        return false;
+      }
+      if (response.status === 204) {
+        return true;
+      }
+      throw new Error(`Error fetching from api: ${response.status} ${response.statusText}`);
+    } catch (error: any) {
+      throw new Error(`Error fetching from api: ${error.message}`);
```
imx-mikhala, line 31: Should we be throwing `CheckoutError` here? I can see in the remote config fetcher we just return an `Error` there too but wondering why we arent using CheckoutError

dreamoftrees:
@deepti-imx @imx-mikhala @sharonsheah just curious about the thinking here with refactoring the AvailabilityService from a class to a factory function?

packages/checkout/sdk/src/availability/availability.ts
deepti-imx: Good point 
Why did we not return CheckoutError in remote config? 🤔

imx-mikhala:
> @deepti-imx @imx-mikhala @sharonsheah just curious about the thinking here with refactoring the AvailabilityService from a class to a factory function?

Was something I suggested. It feels a lot easier to read than a class - no need for `this` everywhere and other class overheads.

packages/checkout/sdk/src/availability/availability.ts
sharonsheah: @deepti-imx we were following config `makeHttpRequest` method which used `throw Error` and not `CheckoutError`

deepti-imx (approved):

PR #964, opened 2023-10-10
Move runners back to ubuntu-latest

timm088:
Since this repository was made public, all github actions have been stalled , unable to select a runner from our enterprise account.

We will need to figure out a strategy here (as switch back to ubuntu-latest will extend build times)

For the moment however, this temporary change will keep CI/CD moving.

Follow up PR will need to consider a runner group just for this repository, and ensure adequate security protections.


timm088:
@cajames @carleeto FYI - Switching back to ubuntu-latest to keep your builds moving.

carleeto (approved):

PR #1192, opened 2023-11-22
refactor: output build files to root of SDK folder

CodeSchwert:
# Summary

Change the rollup build to output build files to the root of the SDK package.

# Customer Impact

Will allow users to import via the code splitting method mentioned in the [developer docs](https://docs.immutable.com/docs/x/sdks/typescript#code-splitting) without needing TS5 and the `moduleResolution` set to `bundler` or `nodenext`.

## Changed
- Location of some build files has changed. Browser bundles are not affected.

# Things worth calling out

`exports` field in `package.json` still needs to be defined, even though it appears that the `exports` definitions are not being honoured! Looks like a bug with `package.json` `exports`.


sdk/package.json
```diff
@@ -75,46 +75,48 @@
...
     "./provider": {
-      "types": "./dist/provider.d.ts",
-      "import": "./dist/provider.js"
+      "types": "./provider.d.ts",
+      "import": "./provider.js"
     },
-    "./checkout": {
-      "types": "./dist/checkout.d.ts",
-      "import": "./dist/checkout.js"
+    "./checkout_sdk": {
```
wdinau, line 106: why this rename is important/ necessary? will it impact other team/ folders?

CodeSchwert: I'll need to fix this in the PR, it shouldn't be renamed. Checkout changed the package name after this PR was created!

sdk/.gitignore
```diff
@@ -1 +1,8 @@
-workspace-packages.json
\ No newline at end of file
+workspace-packages.json
+
+*.js
```
wdinau, line 3: i don't understand why all js and ts files are to be ignored under sdk. in this review we'd have rollup.config.js file under sdk folder, is that no longer needed?

PR #1210, opened 2023-11-28
WT-1946 - fix balances error due to change of network

andrearampin:
# Summary

Due to an error in the retry logic when the user changes network; the widget, at time, believes that there are no balances and therefore it fires a **no enough IMX screen** this also affected the bridge widget.

Additional follow up bug ticket https://immutable.atlassian.net/browse/WT-1948

# Customer Impact

## Security

# Things worth calling out


dreamoftrees (approved):
👍 Nice catch Andy. Is it possible to capture this impact via a cypress test in `XBridgeWidget.cy.tsx` ?

andrearampin:
https://github.com/immutable/ts-immutable-sdk/pull/1279

PR #1243, opened 2023-12-05
refactor: Immutable footer branding into SimpleLayout

dreamoftrees:
# Summary
Refactor to consolidate Immutable footer branding into main SimpleLayout component.
It can be disabled with `footerBrand=false` otherwise defaults to true.

OnRamp and Swap Widgets disable the footer branding.

Screenshot 2023-12-06 at 10 49 02 am

Screenshot 2023-12-06 at 10 41 36 am

Footer buttons now consistently use: ` padding: 'base.spacing.x4',`
Screenshot 2023-12-06 at 10 46 20 am

Updated view examples:

Screenshot 2023-12-06 at 12 52 28 am


packages/checkout/widgets-lib/src/widgets/x-bridge/views/ApproveTxn.tsx
ZacharyCouchman, line 1: @dreamoftrees how come this file is being added? I think this should be added as part of the bridge integration ticket

dreamoftrees: I might wait to the bridge integration and then rebase again .. just dont want to disrupt that at the moment.

PR #1268, opened 2023-12-08
fix: Swap Widget from amount truncation to next non-zero digit

dreamoftrees:
# Summary
The Swap Widget from amount was truncated to a maximum of 6 decimals. Some of the swap quotes required more decimal places, the from amount is now calculated dynamically expanding up to the next non-zero digit or to the maximum of the token decimal.
[WT-1977](https://immutable.atlassian.net/browse/WT-1977)

# Customer Impact
Swap Widget from amount now gives an accurate non-zero value up to the maximum token decimal amount.

Screenshot 2023-12-08 at 2 41 17 pm

[WT-1977]: https://immutable.atlassian.net/browse/WT-1977?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/checkout/widgets-lib/src/lib/utils.test.ts
```diff
@@ -435,8 +435,20 @@ describe('utils', () => {
...
+      expect(tokenValueFormat('0.00000012345')).toEqual('0.000000');
+    });
+
+    it('should format to custom maximum decimal places of 3', () => {
+      expect(tokenValueFormat('0.00000012345', 3)).toEqual('0.000');
+    });
+
+    it('should format to custom maximum decimal places of 18', () => {
+      expect(tokenValueFormat('0.000000000000000000', 18)).toEqual('0.000000000000000000');
+    });
```
ZacharyCouchman, line 448: Hey @dreamoftrees , should we make this case just return '0'?

dreamoftrees: @ZacharyCouchman there is another wrapping function called `formatZeroAmount` that handles this part in the UI.

ZacharyCouchman (approved):

imx-mikhala (approved):

PR #1269, opened 2023-12-08
Feat/wt 1967 bridge sdk integration

ZacharyCouchman:
# Summary

[WT-1967](https://immutable.atlassian.net/browse/WT-1967)

Integrate new Bridge SDK into Checkout.

Updating:
* X Bridge Widget
* Checkout SDK Gas estimator
* Smart Checkout

[WT-1967]: https://immutable.atlassian.net/browse/WT-1967?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/internal/bridge/sdk/rollup.config.js
```diff
@@ -15,7 +15,7 @@ export default {
   plugins: [
     json(),
     commonjs(),
-    nodeResolve(),
+    nodeResolve({ browser: true}),
```
ZacharyCouchman, line 18: @proletesseract we had to make the node Resolve for browser so this would work in Checkout.

The problem is the bridge sample app scripts which run in Node will break. You can get it to work by removing browser: true. 

Most likely we will need to just create a CommonJS build for the node scripts and an ESM build for Checkout

ZacharyCouchman: Or something along those lines

PR #1278, opened 2023-12-12
feat: Checkout i18next support

dreamoftrees:
# Summary
Adding support for i18next in the Checkout Widgets UI.
- [x] Converted existing translation strings into JSON locale files. Japanese, Korean, Chinese.
- [x] Refactored text bindings in UI components to use i18next-react useTranslation hooks.

# Customer Impact
Checkout Widgets support Japanese, Korean, and Chinese language.

Screenshot 2023-12-21 at 12 38 19 pm


packages/checkout/sdk/src/widgets/definitions/configurations/widget.ts
```diff
@@ -7,4 +7,5 @@ import { WidgetTheme } from './theme';
 export type WidgetConfiguration = {
   /** The theme of the Checkout Widget (default: "DARK") */
   theme?: WidgetTheme;
+  language?: 'en' | 'ja' | 'ko' | 'zh'
```
andrearampin, line 10: is this the same as `LanguageChanged`?

dreamoftrees: Yes good catch, I've improved that and mapped back to WidgetLanuage.

sharonsheah (approved):

CodeSchwert (approved):

PR #1279, opened 2023-12-12
WT-1946 - fix balances error due to change of network

andrearampin:
# Summary

Due to an error in the retry logic when the user changes network; the widget, at time, believes that there are no balances and therefore it fires a no enough IMX screen this also affected the bridge widget.

Additional follow up bug ticket https://immutable.atlassian.net/browse/WT-1948

# Customer Impact

## Security

# Things worth calling out


imx-mikhala (approved):

dreamoftrees (approved):
👍 Noice catch @andrearampin

PR #1281, opened 2023-12-13
WT-1996 Correctly handle token decimals when getting unsigned transactions from bridge

imx-mikhala:
# Summary

When getting the unsigned txns for the bridge we were sending the default token decimals of 18. We should use the decimals that come from the token to handle all token decimals. This fixes the issue when trying to bridge across the 6 decimal usdc contract.


dreamoftrees (approved):
Good find 👍

sharonsheah (approved):

PR #1283, opened 2023-12-13
feat: WT-1984 Rename xbridge to bridge

sharonsheah:
# Summary

Rename all xbridge files/folders/variable names to bridge

https://immutable.atlassian.net/browse/WT-1984


packages/checkout/widgets-lib/src/widgets/bridge/BridgeWidgetRoot.tsx
```diff
@@ -59,7 +59,7 @@ export class XBridge extends Base<WidgetType.BRIDGE> {
     this.reactRoot.render(
       <React.StrictMode>
         <WidgetContainer id="xbridge-container" config={this.strongConfig()}>
```
imx-mikhala, line 61: id still says `xbridge` here

sharonsheah: good catch! thanks 😄

imx-mikhala (approved):

dreamoftrees (approved):
🌮

PR #1284, opened 2023-12-13
WT-1934 Not enough gas drawer

imx-mikhala:
# Summary

Adds the not enough gas drawer and conditionally renders depending on whether the native token is ETH or IMX.

Also added the 'Add IMX coins' button, which is what we do in the swap widget. This button will take the user to the TopUpView.

# Demo

### ETH Selected

https://github.com/immutable/ts-immutable-sdk/assets/122326421/4c54a36a-b93b-4694-b1ea-b1af82ceab97

### Non-Native Token Selected

https://github.com/immutable/ts-immutable-sdk/assets/122326421/f17c7fd8-a225-4ecb-a101-0f00446697c4

### IMX Selected

https://github.com/immutable/ts-immutable-sdk/assets/122326421/5ffcb876-5c71-4362-9d77-372762b2b9f2

### Non-Native Token Selected

https://github.com/immutable/ts-immutable-sdk/assets/122326421/ba078630-f7e1-46e9-8a08-920a4781a963

### DIFF
--- a/packages/checkout/widgets-lib/src/views/top-up/TopUpView.tsx
+++ b/packages/checkout/widgets-lib/src/views/top-up/TopUpView.tsx
@@ -122,7 +122,7 @@ export function TopUpView({
       const onRampFees = getOnRampFeeEstimation(onRampFeesEstimate);
       setOnRampFeesPercentage(onRampFees);
     })();
-  }, [checkout]);
+  }, [checkout !== undefined]);
 
   // Check if swap is available
   useEffect(() => {
--- a/packages/checkout/widgets-lib/src/views/top-up/TopUpView.tsx
+++ b/packages/checkout/widgets-lib/src/views/top-up/TopUpView.tsx
@@ -120,7 +120,7 @@ export function TopUpView({
       const onRampFees = getOnRampFeeEstimation(onRampFeesEstimate);
       setOnRampFeesPercentage(onRampFees);
     })();
-  }, [checkout !== undefined]);
+  }, [checkout]);
 
   // Check if swap is available
   useEffect(() => {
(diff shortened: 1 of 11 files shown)


packages/checkout/widgets-lib/src/views/top-up/TopUpView.tsx
```diff
@@ -27,14 +29,15 @@ import {
...
-import { ConnectLoaderContext } from '../../context/connect-loader-context/ConnectLoaderContext';
 import { isPassportProvider } from '../../lib/providerUtils';
 import { OnRampWidgetViews } from '../../context/view-context/OnRampViewContextTypes';
 import { EventTargetContext } from '../../context/event-target-context/EventTargetContext';
 import { TopUpMenuItem } from './TopUpMenuItem';
 
 interface TopUpViewProps {
   widgetEvent: IMTBLWidgetEvents,
+  checkout?: Checkout,
+  provider?: Web3Provider,
```
imx-mikhala, line 40: Passing through via props now instead as the TopUpView was using the values from connect loader state, which the move widget does not have

packages/checkout/widgets-lib/src/widgets/wallet/views/WalletBalances.tsx
```diff
@@ -264,6 +264,7 @@ export function WalletBalances({
           onCloseDrawer={() => setShowNotEnoughGasDrawer(false)}
           walletAddress={walletAddress}
           showAdjustAmount={false}
+          tokenSymbol={ETH_TOKEN_SYMBOL}
```
imx-mikhala, line 267: Wallet widget shows a not enough gas drawer when `Move` is selected but the user has no ETH - so just adding the token symbol here

sharonsheah (approved):
LGTM! Nice work! 😄

dreamoftrees (approved):
Nice work @imx-mikhala 👍

PR #1285, opened 2023-12-14
feat: WT-1990 Rename contractAddress to tokenAddress

sharonsheah:
# Summary

- Rename `contractAddress` to `tokenAddress`
- Rename `fromContractAddress` to `fromTokenAddress`
- Rename `toContractAddress` to `toTokenAddress`
- JS docs updated

https://immutable.atlassian.net/browse/WT-1990

# Customer Impact

This will be a breaking change with the API endpoint change from contractAddress to tokenAddress. Customers would need to update params interface if they update sdk version.


jwhardwick (approved):

dreamoftrees (approved):
Nice work @sharonsheah I think maybe worth putting in the PR notes the section around Customer Impact and noting that this will be a breaking change with the API endpoint change from contractAddress to tokenAddress.

PR #1291, opened 2023-12-18
WT-2010 - Move Segment configurations to remote

andrearampin:
# Summary

Move Segment configurations to remote so that it is easier to change in the future.

# Customer Impact

## Security

# Things worth calling out


packages/checkout/sdk/src/config/remoteConfigFetcher.ts
```diff
@@ -90,6 +90,10 @@ export class RemoteConfigFetcher {
     return config[key];
   }
 
+  public isLoaded(): boolean {
+    return this.configCache !== undefined;
+  }
+
```
deepti-imx, line 96: Where is this fn used?

packages/checkout/widgets-lib/src/context/analytics-provider/SetupAnalytics.tsx
```diff
@@ -1,21 +1,35 @@
...
+  }, [checkout]);
+
   useEffect(() => {
-    const writeKey = getSegmentWriteKey(widgetConfig.environment);
-    updateWriteKey(writeKey);
-  }, [widgetConfig]);
+    (async () => {
+      const config = await telemetry() as TelemetryConfig;
+      if (!config) return;
+      updateWriteKey(config.publishableApiKey);
```
deepti-imx, line 30: Is `publishableApiKey` same as Segment key?

andrearampin: it is the segment key

deepti-imx: sorted in slack chat. Yes, it is a segment key.

deepti-imx (approved):

dreamoftrees (approved):
Looks good @andrearampin .. was just wondering about the name `CustomAnalyticsProvider` and if we should use something more contextual like `CheckoutAnalyticsProvider` not sure?

PR #1292, opened 2023-12-18
feat: Support for Publishable Key in Checkout SDK

dreamoftrees:
# Summary
Adds support for Checkout SDK to include `x-immutable-publishable-key` HTTP headers when HUB publishableKey is included in the Checkout configuration.
[WT-1992](https://immutable.atlassian.net/browse/WT-1992)

key

[WT-1992]: https://immutable.atlassian.net/browse/WT-1992?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/checkout/sdk/src/api/http/httpClient.ts
```diff
@@ -0,0 +1,110 @@
...
+          if (!publishableKey.startsWith(PUBLISHABLE_KEY_PREFIX)) {
+            throw new Error(
+              'Invalid Publishable key. Create your Publishable key in Immutable developer hub.'
+              + ' https://hub.immutable.com',
+            );
+          }
+
+          // Add publishable key header for whitelisted domains
+          // eslint-disable-next-line no-param-reassign
+          config.headers['x-immutable-publishable-key'] = publishableKey;
```
andrearampin, line 45: Is the reason for not using
```
axios.defaults.headers.common['x-immutable-publishable-key'] = publishableKey;
```
and therefore avoid to wrap the `get`, `post`, ... requests the fact that you are injecting the header only for some domains?

dreamoftrees: Yep exactly, otherwise those global headers would apply to OrderBook and anything else in the SDK that uses axios.

sharonsheah (approved):

packages/checkout/sdk/src/api/http/httpClient.ts
```diff
@@ -0,0 +1,110 @@
+import axios, { AxiosInstance, AxiosRequestConfig, AxiosResponse } from 'axios';
+import { CheckoutModuleConfiguration } from '../../types';
+import { CheckoutError, CheckoutErrorType } from '../../errors';
+
+const PUBLISHABLE_KEY_PREFIX = 'pk_imapik-';
+
+const publishableKeyWhitelistedDomain = [
```
imx-mikhala, line 7: can we use the term `allowlist` here, thats what we use throughout checkout

dreamoftrees: Yep. good call. done

sharonsheah (approved):

deepti-imx (approved):

PR #1302, opened 2023-12-20
Smr 2092 flowrate withdraw

proletesseract:
# Summary

This PR adds the `tokenBridge.getFlowRateWithdrawTx` which will generate the unsigned transaction required to withdraw from the flow rate limiter if the transaction can be withdrawn. Otherwise it will return the information about the transaction and when it can be withdrawn.

All unit tests and e2e tests passing.

Main code to review is the `getFlowRateWithdrawTx` method in `packages/internal/bridge/sdk/src/tokenBridge.ts`


packages/internal/bridge/sdk/src/tokenBridge.ts
```diff
@@ -849,21 +849,21 @@ export class TokenBridge {
...
 
       let flowRatePromiseIndex: number;
       if (metaStatus === StatusResponse.COMPLETE
-        && isWithdraw && txItem.receiver) {
-        // consolidate the calls we have to make to the flow rate by receiver
-        if (!flowRatePromisesReceivers.includes(txItem.receiver)) {
-          flowRatePromises.push(rootBridge!.getPendingWithdrawals(txItem.receiver, [0]));
-          flowRatePromisesReceivers.push(txItem.receiver);
+        && isWithdraw && txItem.recipient) {
+        // consolodate the calls we have to make to the flow rate by receiver
```
dreamoftrees, line 860:
```suggestion
        // consolidate the calls we have to make to the flow rate by receiver
```

dreamoftrees (dismissed):
Great work @proletesseract 👍 
Just a minor spelling.

packages/internal/bridge/sdk/src/tokenBridge.ts
```diff
@@ -979,12 +979,86 @@ export class TokenBridge {
...
+          this.config.bridgeContracts.rootERC20BridgeFlowRate,
+          ROOT_ERC20_BRIDGE_FLOW_RATE,
+          this.config.rootProvider,
+        );
+        return contract;
+      },
+      BridgeErrorType.INTERNAL_ERROR,
+    );
+
+    const pending:Array<RootBridgePendingWithdrawal> = await rootBridge.getPendingWithdrawals(req.recipient, [0]);
```
ermyas, line 1009: I guess this only supports single withdrawal. 
- Is there a reason we don't take the index of the withdrawal as an argument? 
- Can I confirm that there will be a subsequent PR to support multiple withdrawals, or is that not in scope just yet?

ermyas (changes requested):

packages/internal/bridge/bridge-sample-app/src/withdrawTx.ts
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable no-console */
+import 'dotenv/config';
+
+import { ImmutableConfiguration, Environment } from '@imtbl/config';
+
+import { 
+    TokenBridge, 
+    BridgeConfiguration, 
+    ETH_SEPOLIA_TO_ZKEVM_TESTNET,
```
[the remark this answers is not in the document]
proletesseract: the way the bridge was previously coded is that when the bridge is instantiated, based on the environment you pass in the config pairs two networks together as Root & Child and sets the bridge to one of these constants. You're right though, the sample app should be configurable to testnet or mainnet. i will add an environment flag in the .env file and set these chain pairings based on that.

packages/internal/bridge/bridge-sample-app/src/withdrawTx.ts
```diff
@@ -0,0 +1,60 @@
...
+// @ts-ignore
+import { setupForBridge } from './lib/utils.ts';
+
+async function getFlowRateWithdrawTx() {
+
+  const params = await setupForBridge();
+
+  if (!process.env.FLOW_RATE_INDEX) {
+    throw new Error('FLOW_RATE_INDEX not set');
+  }
```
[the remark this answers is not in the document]
proletesseract: the index is part of the SDK method params, but this is the sample app, so we need a way to set the index and address we are passing into the SDK method to generate the unsigned transaction for, which is why its part of the env file.

packages/internal/bridge/sdk/src/tokenBridge.ts
proletesseract: `rootBridge.getPendingWithdrawals` just gets all the pending withdrawals for an address starting at the index that's passed in. We want to fetch all of the pending withdrawals so we pass in `[0]` to this contract call. we then check there are pending withdrawals and that the `req.index` exists before trying to generate the unsigned transaction for the requested withdrawal index a bit further down.

There is the capability on the contract to sign a TX to execute all pending withdrawals but this is not required by Checkout yet so it's not in scope for this ticket.

packages/internal/bridge/bridge-sample-app/src/withdrawTx.ts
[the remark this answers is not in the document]
proletesseract: ive also switched out the rootChainID and childChainID params to be based on the bridge instance config. Everything should be now based on the ENVIRONMENT which is set in the env with no hard coded chainIDs or bridge instance constants.

packages/internal/bridge/sdk/src/tokenBridge.ts
ermyas: > starting at the index that's passed in.

are you sure? my recollection was that it takes in an array of indices which are the offsets in the withdrawal array to fetch? so just passing an array with a single element [0] just fetches the first element?

proletesseract: let me double check but pretty sure:

https://github.com/immutable/zkevm-bridge-contracts/blob/main/src/root/flowrate/FlowRateWithdrawalQueue.sol#L154

proletesseract: hmm actually it looks like you are right now i read the code, from the comment i thought it was just a starting offset. okay ill have to change how this works a bit.

proletesseract: i've made the changes, so we now query for the pendingWithdrawalsLength then create an array of indices based on the length and fetch those. I have actually made these changes in the other PR though since the same issue existed over there as well.

dreamoftrees (approved):
👍

ermyas (approved):

PR #1304, opened 2023-12-20
WT-1980 Update smart checkout fee types

imx-mikhala:
# Summary

Adds a fee type to clearly differentiate between gas, bridge/swap & the immutable fee

Also this PR renames the following, as they are just a singular fee and not part of an array:
- `approvalGasFees` -> `approvalGasFee`
- `swapGasFees` -> `swapGasFee`
- `bridgeGasFees` -> `bridgeGasFee`


dreamoftrees (approved):
Great work @imx-mikhala 👍

sharonsheah (approved):

PR #1307, opened 2023-12-21
feat: WT-1991 enable move option for balance item on L2

sharonsheah:
# Summary

https://immutable.atlassian.net/browse/WT-1991


dreamoftrees (approved):
👍 Noice

deepti-imx (approved):

PR #1310, opened 2023-12-21
fix: tokenId is uint256 so must be a BigNumber

keithbro-imx:
# Summary

tokenId is uint256 so must be a BigNumber

Test in checkout sdk-sample-app

# Customer Impact

When listing NFTs with a number that is too big to fit in a JS number, it will succeed.

## Security

# Things worth calling out


ZacharyCouchman (dismissed):
Have rebased this on latest main now and tested in checkout sdk-sample-app.

All looks good.

ckailash (approved):

PR #1315, opened 2023-12-22
Update testnet swap proxy address

keithbro-imx:
# Summary

Update the swap proxy address to the one we just deployed, and update a few more old references to secondary fees.

# Customer Impact

None.

## Security

# Things worth calling out


pano-skylakis (approved):

packages/internal/dex/sdk/src/types/index.ts
```diff
@@ -6,7 +6,7 @@ export type ExchangeContracts = {
   coreFactory: string;
   quoterV2: string;
   peripheryRouter: string;
-  secondaryFee: string;
```
pano-skylakis, line 9: Will this break the overrides in the sample app?

Benjimmutable (approved):

packages/internal/dex/sdk/src/types/index.ts
keithbro-imx: GOOD POINT. A job for next year 🖖

PR #1322, opened 2024-01-08
WT-1957 Ignore casing on address check when finding balance in smart checkout

imx-mikhala:
# Summary

Lower cases the addresses when doing comparison check for finding the users balance of a given token item requirement


sharonsheah (approved):

dreamoftrees (dismissed):
@imx-mikhala nice one, the fix looks good. Just curious what the scenario would be where this could happen? I worry if we need to compare addresses like this there are a few different places where comparison happens and we might get varying results?

packages/checkout/sdk/src/smartCheckout/routing/indexer/fetchL1Representation.ts
```diff
@@ -46,9 +47,8 @@ export const fetchL1Representation = async (
     contractAddress: l2address,
   });
 
-  // TODO: When bridge is ready we need to understand how L2 ETH will be mapped back to L1 ETH
```
imx-mikhala, line 49: Is this still something we need to look into @dreamoftrees ?

dreamoftrees: All good, I think this is sorted now.

dreamoftrees (approved):
Nice one @imx-mikhala 👍

dreamoftrees (dismissed):
👍

dreamoftrees (approved):
👍

sharonsheah (approved):

PR #1330, opened 2024-01-09
WT-2019 - migrate to bridge configs per chain

andrearampin:
# Summary

**Context**
The remote config had a ‘bridge’ section on the /config endpoint, but we removed this. Smart checkout is still using this bridge key to lookup the allowed bridgeable tokens so now smart checkout isnt suggesting bridge route as it thinks there are no allowed tokens.

**Description**
Based on the review from https://immutable.atlassian.net/browse/WT-2025 the bridge should have an isolated allowlist that Smart Checkout can also reference.

# Customer Impact

## Security

# Things worth calling out


packages/checkout/sdk/src/smartCheckout/allowList/allowListCheck.ts
```diff
@@ -59,7 +60,7 @@ export const allowListCheckForSwap = async (
   availableRoutingOptions: AvailableRoutingOptions,
 ) : Promise<TokenInfo[]> => {
   if (availableRoutingOptions.swap) {
-    const allowedTokens = ((await config.remote.getConfig('dex')) as DexConfig)?.tokens ?? [];
+    const allowedTokens = (await getTokenAllowList(config, { type: TokenFilterTypes.SWAP })).tokens;
```
dreamoftrees, line 63: Nice cleanup!

dreamoftrees (dismissed):
Nice work looks great!

packages/checkout/sdk/jest.config.ts
```diff
@@ -12,7 +12,7 @@ const config: Config = {
     '@imtbl/blockchain-data': '<rootDir>../../blockchain-data/sdk/src',
     '@imtbl/generated-clients': '<rootDir>../../internal/generated-clients/src'
   },
-  testEnvironment: 'node',
```
andrearampin, line 15: needed to make the `window.navigator` works -- this is a must to correctly run the tests

dreamoftrees (approved):
Done

packages/checkout/sdk/jest.config.ts
ZacharyCouchman: Can you not just add `/** @jest-environment jsdom */` to the top of the file instead of changing this for all of the tests. I think we have done this in other files as well to support jsdom testEnvironment.

PR #1336, opened 2024-01-10
refactor: WT-2041 - better axios response handling

andrearampin:
# Summary

In the current implementation we are only checking if the response is 

```if (response.status === 204)  return true;```

however we should return true if we get any 2xx responses.

# Customer Impact

n/a

## Changed

Axios response handling while checking for DEX availability

# Things worth calling out

### CODE REVIEW

**Review by ZacharyCouchman — APPROVED** [review]

File: packages/checkout/sdk/src/availability/availability.test.ts [hunk 1]
```diff
@@ -26,7 +24,7 @@ describe('availabilityService', () => {
       const mockResponse = {
         status: 403,
       };
-      mockedAxios.post.mockResolvedValueOnce(mockResponse);
+      mockedAxios.post.mockRejectedValueOnce({ response: mockResponse });
```
andrearampin, line 27: axios reject in case of status <> 2xx, it does not resolve

dreamoftrees (approved):
Nice fix @andrearampin

PR #1339, opened 2024-01-11
feat: WT-2032 - Warm up CDN as soon as a new NPM package has been released

andrearampin:
# Summary

Checkout JS bundle is fetched from JS Deliver. JS Deliver will cache that bundle at the first hit; however given the size of the bundle, the time to fetch and cache it is not negligible; therefore on release we can pre-warm the CDN by making the first request.

# Customer Impact

Reduce the load time of the Checkout bundle when a new NPM version is released.

## Changed

Hit JS Deliver when a new NPM version is released.

# Things worth calling out


dreamoftrees (approved):
Nice little boost! 👍

CodeSchwert (approved):

PR #1353, opened 2024-01-15
[GPR-354] Sale Widget | Handle Default Error and Transaction Error

jhesgodi:
# Summary
- When transaction fails with txn hash display error view with View Details ( link to block explorer )
- If not hash, display Dimiss ( closes widget )

- When any other error, display default error view

image

### DIFF
--- a/packages/checkout/sdk/src/blockExplorer/blockExplorer.ts
+++ b/packages/checkout/sdk/src/blockExplorer/blockExplorer.ts
@@ -1,0 +1,16 @@
+import { BLOCKSCOUT_CHAIN_URL_MAP } from '../env';
+import { ChainId } from '../types';
+
+export class BlockExplorerService {
+  /**
+   * getTransationLink returns the link to the transaction on blockscout
+   * @param hash transaction hash
+   * @returns link to the transaction on blockscout
+   */
+  public static getTransactionLink(chainId: ChainId, hash: string): string | undefined {
+    const url = BLOCKSCOUT_CHAIN_URL_MAP?.[chainId]?.url;
+    if (!url || !hash) return undefined;
+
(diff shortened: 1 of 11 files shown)


packages/checkout/sdk/src/blockExplorer/blockExplorer.ts
```diff
@@ -0,0 +1,16 @@
+import { BLOCKSCOUT_CHAIN_URL_MAP } from '../env';
+import { ChainId } from '../types';
+
+export class BlockExplorerService {
```
jhesgodi, line 4: Adding this service since the BLOCKSCOUT_CHAIN_URL_MAP is not exported from sdk.
Please let me know if we better do it a different way

mimi-imtbl (dismissed):

jwhardwick (approved):

packages/checkout/sdk/src/blockExplorer/blockExplorer.ts
dreamoftrees: Hmm yep this makes sense. Not sure if we want to expose the Blockscout instance but we could expose the Blockscout instance URL if you pass in the chainId. Would be better then reconstructing it maybe?

dreamoftrees (approved):
👍

jhesgodi:
@dreamoftrees Thanks for your comment [here](https://github.com/immutable/ts-immutable-sdk/pull/1353#discussion_r1456891459)

Thats actually why this is just a different block explorer class, as I wasnt sure how much of the blockscout api/config we want to expose in the future, hopefully this should be enough for now. and easy to refactor if we want to move this to the existing Blockscout class.

PR #1357, opened 2024-01-16
WT-2034 Lower case addresses when comparing balances to allowlist

imx-mikhala:
# Summary

Lower cases the addresses when checking if balance in allowlist


dreamoftrees (dismissed):
Nice catch @imx-mikhala 👍

sharonsheah (dismissed):

sharonsheah (approved):

dreamoftrees (approved):
👍

PR #1359, opened 2024-01-16
WT-1948 fix: Only show loading shimmer once balance API request started

dreamoftrees:
# Summary
When switching networks in the Wallet Widget we only want to show the loading shimmer once API request started, not while the switch network confirmation is requested in MetaMask.

Fix made as part of https://immutable.atlassian.net/browse/WT-1948

Screenshot 2024-01-17 at 7 19 24 am

### DIFF
--- a/packages/checkout/widgets-lib/src/widgets/wallet/components/NetworkMenu/NetworkMenu.tsx
+++ b/packages/checkout/widgets-lib/src/widgets/wallet/components/NetworkMenu/NetworkMenu.tsx
@@ -34,11 +34,7 @@ import {
 import { EventTargetContext } from '../../../../context/event-target-context/EventTargetContext';
 import { UserJourney, useAnalytics } from '../../../../context/analytics-provider/SegmentAnalyticsProvider';
 
-export interface NetworkMenuProps {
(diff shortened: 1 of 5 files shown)


packages/checkout/widgets-lib/src/widgets/wallet/components/NetworkMenu/NetworkMenu.tsx
```diff
@@ -65,8 +65,6 @@ export function NetworkMenu({ setBalancesLoading }: NetworkMenuProps) {
       });
 
       try {
-        setBalancesLoading(true);
```
imx-mikhala, line 68: is the `setBalancesLoading(false);` needed in the catch below still?

sharonsheah (approved):

packages/checkout/widgets-lib/src/widgets/wallet/components/NetworkMenu/NetworkMenu.tsx
dreamoftrees: No good catch, I've removed it.

imx-mikhala (approved):

PR #1360, opened 2024-01-17
feat: changed zkevm transaction url to be the new generic screen

scastleman-immutable:
# Summary

SDK now takes the user to the new generic transaction confirmation screen when confirming a zkevm transaction.

# Customer Impact

## Changed
URL for zkevm transaction confirmation

# Things worth calling out


carmen0208 (changes requested):
Block it for now

carmen0208 (approved):

haydenfowler (approved):

packages/passport/sdk/src/confirmation/confirmation.ts
```diff
@@ -82,8 +82,8 @@ export default class ConfirmationScreen {
       if (chainType === TransactionApprovalRequestChainTypeEnum.Starkex) {
         href = this.getHref('transaction', { transactionId, etherAddress, chainType });
       } else {
-        href = this.getHref('zkevm', {
-          transactionId, etherAddress, chainType, chainId,
+        href = this.getHref('zkevm/transaction', {
+          transactionID: transactionId, etherAddress, chainType, chainID: chainId,
```
haydenfowler, line 86: Does `chainID` need to be uppercase here?

carmen0208: Yes, although we have accepted both **ID or Id** on the passport screen. It is better to set to ID as the guardian API needs to be set to **chainID**. And uppercase **ID** seems like a standard in backend API.

PR #1362, opened 2024-01-17
WT-1899 Fix shared history array for view state

imx-mikhala:
# Summary

Sets up each widget with their own history array so they are not mutating a shared viewed history.

The history array being shared across multi-widgets was causing widgets to be removed from the view when back buttons were clicked due to the widget trying to render a view it was not familiar with how to handle

Also removed the back buttons from swap/bridge coming from top up view as this does not work as expected. This was only partially working as the view history happened to be getting mutated by multiple widgets, which was an unintended side effect. This would still break if someone went to top up view -> selected swap/bridge -> added something to history from another widget -> tried to go back


dreamoftrees (approved):
👍 Nice work @imx-mikhala

sharonsheah (approved):

PR #1367, opened 2024-01-17
SMR-2280 Optimise getTransactionStatus function

proletesseract:
# Summary
The getTransactionStatus function was quite long and complicated. It has now been broken down into a series of private methods which can be more easily tested and extended.

# Customer Impact
None, functionality remains the same.


packages/internal/bridge/sdk/src/types/index.ts
```diff
@@ -274,6 +274,16 @@ export interface TxStatusResponse {
   transactions: Array<TxStatusResponseItem>
 }
 
+/**
+ * @typedef {Object} TxStatusResponse
```
imx-mikhala, line 278: Should this be:

`* @typedef {Object} AxelarStatusResponse`

proletesseract: yes! thanks

imx-mikhala (dismissed):

packages/internal/bridge/bridge-sample-app/src/deposit.ts
```diff
@@ -38,22 +38,24 @@ async function deposit() {
...
+  
+    if (childBridgeChildAddress === ethers.constants.AddressZero) {
+      console.log('token mapping incomplete, please wait for token to map to childBridge before depositing');
+      return;
+    }
+  
+    if (rootBridgeChildAddress !== childBridgeChildAddress) {
+      console.log(`token mappings mismatch on rootBridge (${rootBridgeChildAddress}) & childBridge (${childBridgeChildAddress}).`, );
+      return;
+    }
```
ermyas, line 58: should we be return an error or throw if these conditions are true, given these are failure cases. Line 182 would also print "Exiting successfully" otherwise, which might be misleading?

packages/internal/bridge/bridge-sample-app/src/deposit.ts
```diff
@@ -38,22 +38,24 @@ async function deposit() {
...
+      console.log('token not mapped, please map token before depositing');
+      return;
+    }
+  
+    if (childBridgeChildAddress === ethers.constants.AddressZero) {
+      console.log('token mapping incomplete, please wait for token to map to childBridge before depositing');
+      return;
+    }
+  
+    if (rootBridgeChildAddress !== childBridgeChildAddress) {
```
ermyas, line 55: This would signal a serious issue with the bridge. Any way we can more clearly surface this error? (see other comment)

packages/internal/bridge/bridge-sample-app/src/deposit.ts
proletesseract: sure im throwing an error now here and in the same checks we do in the withdraw script

packages/internal/bridge/bridge-sample-app/src/deposit.ts
proletesseract: not sure what else to do, perhaps some of these scripts could be run as part of CI/CD or some kind of ongoing monitoring system?

ermyas (approved):

wcgcyx (approved):
LGTM

PR #1369, opened 2024-01-18
feat: WT-2057 Force SmartCheckout to fetch uncached user balance

jwhardwick:
# Summary

* expose a flag on `getAllBalances` `forceFetch`
* when `true`, reset Blockscout clients so the next Blockscout call will be uncached

https://immutable.atlassian.net/browse/WT-2057

# Customer Impact

## Security

# Things worth calling out


dreamoftrees (dismissed):
Looks great @jwhardwick thanks for jumping in and adding your support!

packages/checkout/sdk/src/smartCheckout/balanceCheck/balanceCheck.ts
```diff
@@ -32,6 +32,7 @@ const getTokenBalances = async (
   provider: Web3Provider,
   ownerAddress: string,
   itemRequirements: ItemRequirement[],
+  forceFetch?: boolean,
```
imx-mikhala, line 35: Could we just pass a default param so 

`forceFetch: boolean = false`

instead of needing the `!!forceFetch`

imx-mikhala:
Should this PR be passing `true` to where smart checkout is calling the `balanceCheck()`? Looks like the set up has been done but I cant see where smart checkout is forcing to true. the `balanceCheck` for smart checkout is in `smartCheckout.ts`

packages/checkout/sdk/src/smartCheckout/balanceCheck/balanceCheck.ts
jwhardwick: Good idea, added.

jwhardwick:
Thanks @imx-mikhala yeah good catch, added to `smartCheckout.ts`

imx-mikhala (approved):

dreamoftrees (approved):
👍

PR #1370, opened 2024-01-18
retrieve chain id from api instead of hardcoding

scastleman-immutable:
# Summary
Chain ID is now retrieved from the API rather than being hardcoded

# Customer Impact
Slightly increases time taken to register a new zkevm user due to another API call


packages/passport/sdk/src/zkEvm/user/registerZkEvmUser.ts
```diff
@@ -34,9 +34,11 @@ export async function registerZkEvmUser({
   const headers = { Authorization: `Bearer ${accessToken}` };
 
   const { chainId } = await jsonRpcProvider.ready;
+  const eipChainId = getEip155ChainId(chainId);
+  const chainList = (await multiRollupApiClients.chainsApi.listChains()).data.result;
```
carmen0208, line 38: Do/ How we want to handle the error

scastleman-immutable: We can probably just move it down within the existing try/catch block. Does that work for you?

carmen0208 (approved):

haydenfowler (approved):

PR #1374, opened 2024-01-18
feat: SMR-2094 Get Token Mapping

proletesseract:
# Summary
For convenience and completeness we should provide a way to fetch the token mappings for a token across a particular chain pairing.

This endpoint getTokenMapping and it’s interfaces are already in the code base and detailed here; https://immutable.atlassian.net/wiki/spaces/PRIM/pages/2301591596/Axelar+Bridge+JS+SDK#Get-Token-Mapping-%F0%9F%9A%A7


wcgcyx (approved):

packages/internal/bridge/bridge-sample-app/src/getMapping.ts
```diff
@@ -0,0 +1,92 @@
...
+    rootToken: bridgeConfig.bridgeContracts.rootChainIMX,
+    rootChainId: bridgeConfig.bridgeInstance.rootChainID,
+    childChainId: bridgeConfig.bridgeInstance.childChainID,
+  }
+
+  console.log('tokenMappingIMXReq', tokenMappingIMXReq);
+  
+  const tokenMappingIMXRes: TokenMappingResponse = await tokenBridge.getTokenMapping(tokenMappingIMXReq);
+  console.log('tokenMappingIMXRes');
+  console.log(util.inspect(tokenMappingIMXRes, {showHidden: false, depth: null, colors: true}));
```
ermyas, line 81: I think there is a fair bit of repeated code here that might merit refactoring out into a separate function to improve readability.

ermyas (approved):

PR #1376, opened 2024-01-19
feat: [WT-2070] Checkout buy endpoint support for wait fulfillment transactions options

dreamoftrees:
# Summary
Checkout buy endpoint currently waits for transactions to be fulfilled, adding an override to the buy function adds support to return the transactions without waiting for the transaction settlement.
https://immutable.atlassian.net/browse/WT-2070

# Customer Impact
The Checkout SDK buy endpoint includes a new overrides parameter `waitFulfillmentSettlements` that returns the transactions without waiting for confirmation.


packages/checkout/sdk/src/types/smartCheckout.ts
```diff
@@ -38,6 +43,21 @@ export type BuyResultFailed = {
...
+ * @property {SmartCheckoutSufficient} smartCheckoutResult
+ * @property {SendTransactionResult[]} transactions
+ */
+export type BuyResultFulfillmentsUnsettled = {
+  /** The status to indicate success */
+  status: CheckoutStatus.FULFILLMENTS_UNSETTLED,
+  /** The sufficient result of smart checkout */
+  smartCheckoutResult: SmartCheckoutSufficient,
+  /** Array of transaction results */
+  transactions: SendTransactionResult[],
```
imx-mikhala, line 58: `SendTransactionResult` is the type for the return of `sendTransaction`. These param/result types are specific for that sdk functions return type. I think because of this that type shouldn't be used with the `BuyResultFulfillmentsUnsettled`

Thinking about this and imo its probably fine to have this an array of `TransactionResponse` types as I dont think we would need any additional info in each of those transactions, but if we think its better for extensibility to have this on a new type then we should create a new one instead of using `SendTransactionResult` for the reason above.

Main downside with having this as a new object type instead of `TransactionResponse` is thats another layer (eg `transactions[i].transactionResponse.wait()` instead of `transactions[i].wait()`

Thoughts?

packages/checkout/sdk/src/types/smartCheckout.ts
```diff
@@ -135,11 +155,13 @@ export type CancelResultFailed = {
  * @property {string} SUCCESS - If checkout succeeded as the transactions were able to be processed
  * @property {string} FAILED - If checkout failed due to transactions not settling on chain
  * @property {string} INSUFFICIENT_FUNDS - If checkout failed due to insufficient funds
+ * @property {string} FULFILLMENTS_UNSETTLED - If checkout succeeded but the fulfillments are not yet settled
```
imx-mikhala, line 158: lets say `fulfilment transactions` here in the comment

packages/checkout/sdk/src/smartCheckout/buy/buy.ts
```diff
@@ -87,6 +89,11 @@ export const buy = async (
   config: CheckoutConfiguration,
   provider: Web3Provider,
   orders: Array<BuyOrder>,
+  overrides: {
+    waitFulfillmentSettlements?: boolean;
+  } = {
+    waitFulfillmentSettlements: true,
```
imx-mikhala, line 95: wondering if something like `waitOnFulfillmentSettlements` or `waitForFulfillmentSettlments` reads better - i know its already a pretty long name though

imx-mikhala:
We will need to expose the overrides on the `BuyParams` type for the buy endpoint that is on the top level part of the checkout sdk

Type is in `buy.ts` in types folder

```
export interface BuyParams {
  provider: Web3Provider;
  orders: Array<BuyOrder>;
  overrides: ... // need a type for the overrides
}
```

packages/checkout/sdk/src/smartCheckout/buy/buy.ts
andrearampin: I don't see the need for adding the additional `On` / `For` -- @dreamoftrees please make the call.

packages/checkout/sdk/src/types/smartCheckout.ts
andrearampin: Initial thought here was to not expose when possible etherjs specific types. Beside the web3 provider we use etherjs in other exposed types so perhaps returning `TransactionResponse` is just fine. Given that we have a `BuyResultFulfillmentsUnsettled` specific type, it shouldn't be too hard in the future to extend this response.

packages/checkout/sdk/src/smartCheckout/buy/buy.ts
dreamoftrees: Going to keep it short so it doesn't exceed my internal 3 word rule for variables :)

packages/checkout/sdk/src/types/smartCheckout.ts
dreamoftrees: I've updated this to return `TransactionResponse`. I think using more direct structure makes sense to me and it's not tied to the sendTransaction result as you mentioned @imx-mikhala

packages/checkout/sdk/src/types/smartCheckout.ts
```diff
@@ -50,6 +69,15 @@ export type BuyResultInsufficientFunds = {
   smartCheckoutResult: SmartCheckoutInsufficient
 };
 
+/**
+ * Represents the overrides available for {@link Checkout.buy}
+ * @property {boolean} waitFulfillmentSettlements
+ */
+export type BuyOverrides = {
+  /** If the buy should wait for the fulfillment transactions to settle */
```
imx-mikhala, line 77: should we mention its true by default in the description

dreamoftrees: I thought about this but decided not to as I didn't want the comment and code to become out of sync if that default ever changes. I guess the public facing endpoint isn't overly obvious though as it doesn't have a default. Would you like me to add it in here?

imx-mikhala: The type itself cant have a default value but could the exposed function show there is a default value? i think we should communicate there is a default value, otherwise as an optional override from an integration perspective it might look like or be assumed that its false by default

imx-mikhala: Lets just add this to the actual docs instead

imx-mikhala (approved):

packages/checkout/sdk/src/types/smartCheckout.ts
dreamoftrees: As discussed I'll update with description around defaults in the docs.

andrearampin (approved):

PR #1382, opened 2024-01-22
WT-1973 Get quote for 1 of the from token for conversion display

imx-mikhala:
# Summary

Current implementation was doing division which has issues with large numbers in js.
Rather than doing the conversion manually this PR fetches a quote of 1 of the from token to get the amount of to token that would be received

https://github.com/immutable/ts-immutable-sdk/assets/122326421/958d4c61-cb1b-465b-9d19-581102b60ee6


dreamoftrees (approved):
Nice find and fix @imx-mikhala 👍

andrearampin (approved):

PR #1385, opened 2024-01-22
feat: WT-2076 - expose root raw error

andrearampin:
# Summary

It is currently extremely hard to debug the root cause of a CheckoutError because the raw error isn’t always exposed.

https://imtbl.slack.com/archives/C02UK3GC2N4/p1705887314679759 

Add the root error to the CheckoutError payload. Example:
```ts
    throw new CheckoutError(
      'An error occurred while signing the message',
      CheckoutErrorType.SIGN_MESSAGE_ERROR,
      { error: err },
    );
```

# Customer Impact

Allow partners to access to root error and stack trace.

## Changed

`CheckoutError` `data` payload
# Things worth calling out


dreamoftrees (approved):
Nice work @andrearampin! 👍 
Just a minor nit I'm not a fan of abbreviations like `err` but I can see it's how the code was originally written.

andrearampin:
> Nice work @andrearampin! 👍 Just a minor nit I'm not a fan of abbreviations like `err` but I can see it's how the code was originally written.

yep me too, I just tried to avoid as many changes as possible in this PR

imx-mikhala (approved):

PR #1386, opened 2024-01-22
feat: [WT-2070] Cancel endpoint to support waitFulfillmentSettlements

dreamoftrees:
# Summary
Updates to Checkout SDK cancel endpoint to support the `waitFulfillmentSettlements` override and return without waiting for the fulfillment transactions to be settled.


packages/checkout/sdk/src/smartCheckout/cancel/cancel.ts
```diff
@@ -54,16 +59,38 @@ export const cancel = async (
     );
   }
 
-  const result = await signFulfillmentTransactions(provider, [unsignedCancelOrderTransaction]);
-  if (result.type === SignTransactionStatusType.FAILED) {
+  if (overrides.waitFulfillmentSettlements) {
+    const result = await signFulfillmentTransactions(provider, [unsignedCancelOrderTransaction]);
```
andrearampin, line 63: Sorry I should have caught it in the prev PR -- why aren't we passing a `waitSettlements` flag to `signFulfillmentTransactions` and let `signFulfillmentTransactions` deal with the logic so you don't have to duplicating the `signFulfillmentTransactions`'s logic on like 79?

dreamoftrees: Yep good suggestion, I was wondering if we should be sending back the transactions on success as well.. I'll refactor to include that.

imx-mikhala (approved):

andrearampin (approved):

PR #1387, opened 2024-01-22
Tp 1903/wrap unwrap imx

Benjimmutable:
# Summary
DEX SDK can now properly handle wrapping IMX and unwrapping WIMX.

# Customer Impact
This is required functionality for a token swap product.

## Added 
- Wrapping IMX
- Unwrapping WIMX

# Things worth calling out
Regarding gas estimates for wrapping unwrapping:
TL;DR is the difference between most expensive and least expensive is marginal assuming 10gwei IMX gas price, so we hard-code the upper bound.

45038 is the upper bound of the gas estimate for the deposit function. This upper bound is reached when a user is wrapping for the first time (cold + unused storage slot). The lower bound is 27,938, so the difference is 17,100 gas. If transactions are 10 gwei IMX, this is about 0.000171 IMX difference. In order to avoid an extra remote call, we just use the upper bound instead.

Please feel free to challenge this ^


packages/internal/dex/sdk/src/exchange.ts
```diff
@@ -143,6 +145,133 @@ export class Exchange {
...
+  private async getUnwrapTransaction(
+    fromAddress: string,
+    tokenAmount: CoinAmount<ERC20>,
+    wimxInterface: ethers.utils.Interface,
+    gasPrice: CoinAmount<Native> | null,
+  ) : Promise<WrapUnwrapTransactionDetails> {
+    const calldata = wimxInterface.encodeFunctionData('withdraw', [tokenAmount.value]);
+    // 35216 is the upper bound of the gas estimate for the withdraw function.
+    // The differenfce between upper and normal is less than 0.0001 IMX, so we just use the upper bound
+    // to avoid an extra remote call.
```
keithbro-imx, line 157: Nice optimization

packages/internal/dex/sdk/src/exchange.ts
```diff
@@ -143,6 +145,133 @@ export class Exchange {
...
+    // this is about 0.000171 IMX difference. In order to avoid an extra remote call, we just use the upper bound instead.
+    const gasEstimate = ethers.BigNumber.from('45038');
+
+    const gasFeeEstimate = gasPrice ? toPublicAmount(calculateGasFee(gasPrice, gasEstimate)) : null;
+    // Get the transaction details. This is calling `deposit` or `withdraw` on the WETH/WIMX contract.
+    const transactionDetails: TransactionDetails = {
+      transaction: {
+        data: calldata,
+        to: this.wrappedNativeToken.address, // wrapping and unwrapping is done on the WETH/WIMX contract itself
+        value: ethers.BigNumber.from(tokenAmount.value), // Wrapping involves sending the native asset as TX.value
```
keithbro-imx, line 205: `tokenAmount.value` is a bignumber already so shouldn't need to use `.from` here

packages/internal/dex/sdk/src/exchange.ts
```diff
@@ -143,6 +145,133 @@ export class Exchange {
...
+      fees: [],
+    };
+
+    const wimxInterface = WIMX__factory.createInterface();
+    let transactionDetails;
+
+    if (isWrap) {
+      transactionDetails = this.getWrapTransaction(
+        fromAddress,
+        otherTokenCoinAmount as CoinAmount<ERC20>,
```
keithbro-imx, line 244: I would try to avoid using `as` here, as it means we're overriding typescipt's smarts.

Easiest way to do this would be to pass in the BigNumber instead since we don't actually need the token info inside the function

packages/internal/dex/sdk/src/exchange.ts
```diff
@@ -143,6 +145,133 @@ export class Exchange {
...
+    }
+
+    return { quote, approval: transactionDetails.approval, swap: transactionDetails.transaction };
+  }
+
+  private isWrapOrUnwrap(tokenIn: Coin, tokenOut: Coin): boolean {
+    return (
+      (
+        tokenIn === this.nativeToken && (
+          tokenOut.type === 'erc20' && tokenOut.address === this.wrappedNativeToken.address
```
keithbro-imx, line 264: Knowing whether a given token is the wrapped native seems like the responsibility of the NativeTokenService. How about we add a method there to do this check? It would make this less verbose and more readable

packages/internal/dex/sdk/src/types/index.ts
```diff
@@ -97,6 +97,16 @@ export type TransactionResponse = {
...
 };
 
+/**
+ * Type representing the details of a wrap or unwrap transaction
+ * @property {@link TransactionDetails} transaction - The wrap or unwrap transaction
+ * @property {@link TransactionDetails | null} approval - The approval transaction or null if it is not required
+ */
+export type WrapUnwrapTransactionDetails = {
+  transaction: TransactionDetails,
+  approval: TransactionDetails | null,
```
keithbro-imx, line 107: Does not providing a quote mean Checkout will need to handle this case differently on their side?

pano-skylakis: I think so - we should be able to provide a Quote object still, and IMO a Quote still makes sense, even if it is fixed.

Without these details I don't think the consumer of this SDK can provide a breakdown of the transaction prior to the user initiating

pano-skylakis: In the context of a wrap/unwrap, it could just look like:

```typescript
const wrapQuote = {
  amount: 100,
  amountWithMaxSlippage: 100,
  slippage: 0, // No slippage gets applied
  fees: [] // No fees get applied
}
```

keithbro-imx: agree

Benjimmutable: We still provide a quote at the top level. This type was made for the internal details where we split off for wrapping/unwrapping.

Benjimmutable: @pano-skylakis yep that's what we are doing already

packages/internal/dex/sdk/src/exchange.ts
Benjimmutable: Good catch ty

keithbro-imx (approved):

packages/internal/dex/sdk/src/lib/nativeTokenService.ts
```diff
@@ -28,4 +28,19 @@ export class NativeTokenService {
   maybeWrapAmount(amount: CoinAmount<Coin>): CoinAmount<ERC20> {
     return newAmount(amount.value, this.maybeWrapToken(amount.token));
   }
+
+  isWrapOrUnwrap(tokenIn: Coin, tokenOut: Coin): boolean {
+    return (
+      (
+        tokenIn === this.nativeToken && (
+          tokenOut.type === 'erc20' && tokenOut.address === this.wrappedToken.address
```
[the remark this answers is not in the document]
Benjimmutable: Oh yep I misread your original comment

packages/internal/dex/sdk/src/exchange.ts
```diff
@@ -143,6 +146,109 @@ export class Exchange {
...
+    gasPrice: CoinAmount<Native> | null,
+  ) : Promise<WrapUnwrapTransactionDetails> {
+    const calldata = wimxInterface.encodeFunctionData('withdraw', [tokenAmount.value]);
+    // 35216 is the upper bound of the gas estimate for the withdraw function.
+    // The differenfce between upper and normal is less than 0.0001 IMX, so we just use the upper bound
+    // to avoid an extra remote call.
+    const gasEstimate = ethers.BigNumber.from(IMX_UNWRAP_GAS_COST);
+
+    const gasFeeEstimate = gasPrice ? toPublicAmount(calculateGasFee(gasPrice, gasEstimate)) : null;
+    // 3. Get the transaction details. This is calling `deposit` or `withdraw` on the WETH/WIMX contract.
```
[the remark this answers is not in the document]
Benjimmutable: Forgot to update some of these when I moved some stuff around.

packages/internal/dex/sdk/src/exchange.ts
```diff
@@ -174,6 +280,18 @@ export class Exchange {
 
     const amountSpecified = newAmount(amount, tokenSpecified);
 
+    if (this.nativeTokenService.isWrapOrUnwrap(tokenIn, tokenOut)) {
+      // If the user is swapping between the native token and the wrapped native token,
+      // we want to just wrap/unwrap the native token instead of swapping
+      return this.getUnsignedWrapUnwrapTx(
```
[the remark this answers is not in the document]
Benjimmutable: This will break the quote functionality. In that case, `if isWrap -> quoteToken = WIMX`. However, the case should be `if isWrap ^ tradeType = FROM_AMOUNT_OUT -> quoteToken = IMX`.

As in, if we wrap, but it's a "fromAmountOut" swap, the quote token should be the `tokenIn`. Specifying the `tokenIn` is the sufficient + necessary info for determining `isWrap`, and `otherToken` is the sufficient + necessary info for determining which token should be in the quote.

So we can still reduce by 1 parameter, and then the function looks like:
```
private async getUnsignedWrapUnwrapTx(
    tokenSpecified: Coin,
    amountIn: CoinAmount<Coin>,
    fromAddress: string,
    gasPrice: CoinAmount<Native> | null,
  ): Promise<TransactionResponse> {
    const isWrap = tokenSpecified.type === 'native';
    const quoteToken = isWrap ? this.wrappedNativeToken : this.nativeToken;
    const otherTokenCoinAmount = {
      token: quoteToken,
      value: amountIn.value,
    };
```

Thoughts?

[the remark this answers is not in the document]
keithbro-imx: Ah yeah I get you, good point

keithbro-imx (approved):

pano-skylakis (approved):
Nice one! 🚀

PR #1398, opened 2024-01-24
auth token refreshes if expired before executing zkevm transactions

scastleman-immutable:
# Summary
signTypedData and sendTransaction requests now call `authManager.getUser` in order to ensure the user is logged in, which is the same approach as is implemented in passportImxProvider. Token refreshing now checks if a token is expiring within the upcoming 10 seconds and pre-emptively renews it.

Is 10 seconds an appropriate grace period? Should we consider increasing it further?


packages/passport/sdk/src/zkEvm/zkEvmProvider.ts
```diff
@@ -113,6 +113,15 @@ export class ZkEvmProvider implements Provider {
       && this.guardianClient !== undefined;
   }
 
+  private async refreshUserLogin() {
```
haydenfowler, line 116: I think the approach we've taken here only partially solves the issue. When we fixed this in `passportImxProvider`, the approach that we took was to never store a reference to the user. Instead, we would always fetch the user from the `authManager`. When `eth_requestAccounts` is called, we pass the `user` to the `RelayerClient` constructor, so even though we're refreshing `this.user` in `zkEvmProvider`, the `user` that the `RelayerClient` uses would still be out of date.

Maybe a better approach here would be to update the `RelayerClient` & `GuardianClient` constructors to accept the `authManager`, that way they can both retrieve the `user` themselves. This would also be more consistent with the approach we've taken in `passportImxProvider`.

Thoughts?

haydenfowler: To keep the `isLoggedIn()` check nice and simple, we could just store the `ethAddress` and check that 🤔 especially since `eth_sendTransaction` & `eth_signTypedDataV4` only need a user for that address, and to pass the `accessToken` to the `RelayerClient` & `GuardianClient` (but with the above changes, they should get the `accessToken` by calling the `AuthManager`)

scastleman-immutable: Good call, will update the solution to reflect that.

packages/passport/sdk/src/zkEvm/zkEvmProvider.ts
```diff
@@ -106,73 +105,76 @@ export class ZkEvmProvider implements Provider {
...
       }
       case 'eth_accounts': {
-        return this.isLoggedIn() ? [this.user.zkEvm.ethAddress] : [];
+        const user = await this.authManager.getUser() as UserZkEvm;
+        return this.isLoggedIn() ? [user.zkEvm.ethAddress] : [];
       }
       case 'eth_signTypedData':
       case 'eth_signTypedData_v4': {
+        const user = await this.authManager.getUser() as UserZkEvm;
         if (!this.isLoggedIn()) {
```
haydenfowler, line 168: Looks like we've forgotten to pass the `user` to the `isLoggedIn` function here.
To avoid a similar issue in the future, can we store the `user.zkEvm.ethAddress` in a property & reference? That way, we don't have to call `getUser()` from this module and can just pass the `ethAddress` to subsequent modules (e.g. `sendTransaction`), and `isLoggedIn` can be simplified to:
```
  private isLoggedIn(): this is LoggedInZkEvmProvider {
    return this.magicProvider !== undefined
      && this.ethAddress !== undefined
      && this.relayerClient !== undefined
      && this.guardianClient !== undefined;
  }
```

packages/passport/sdk/src/guardian/guardian.ts
```diff
@@ -11,11 +12,11 @@ import { PassportConfiguration } from '../config';
...
   config: PassportConfiguration;
+  authManager: AuthManager;
 };
 
 export type GuardianEvaluateImxTransactionParams = {
   payloadHash: string;
-  user: UserImx;
 };
 
 type GuardianEVMValidationParams = {
```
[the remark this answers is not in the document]
scastleman-immutable: We can, the reason I passed through the user is because the only call to `evaluateEvmTransaction` is on the very next line after we retrieve the user in `validateEvmTransaction`, which is something we need to do in order to provide the address for `confirmationScreen.requestConfirmation`, so it seemed redundant. Happy to change it for consistency if you'd prefer though.

packages/passport/sdk/src/guardian/guardian.ts
```diff
@@ -155,8 +160,8 @@ export default class GuardianClient {
   private async evaluateEVMTransaction({
     chainId,
     nonce,
-    user,
     metaTransactions,
+    user,
```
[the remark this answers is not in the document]
scastleman-immutable: We can, the reason I passed through the user is because the only call to evaluateEvmTransaction is on the very next line after we retrieve the user in validateEvmTransaction, which is something we need to do in order to provide the address for confirmationScreen.requestConfirmation, so it seemed redundant. Happy to change it for consistency if you'd prefer though.

packages/passport/sdk/src/guardian/guardian.ts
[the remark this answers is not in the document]
scastleman-immutable: They were still being passed in, but I'll remove it as part of the other requested change

packages/passport/sdk/src/zkEvm/zkEvmProvider.ts
```diff
@@ -108,71 +109,72 @@ export class ZkEvmProvider implements Provider {
...
   }
 
   private async performRequest(request: RequestArguments): Promise<any> {
     switch (request.method) {
       case 'eth_requestAccounts': {
+        const existingUser = await this.authManager.getUser() as UserZkEvm;
         if (this.isLoggedIn()) {
-          return [this.user.zkEvm.ethAddress];
+          return [existingUser.zkEvm.ethAddress];
         }
```
haydenfowler, line 123: Can we move the call to `getUser` back into `loginZkEvmUser` please? We can also update this return to be `return [this.zkevmAddress];`

packages/passport/sdk/src/zkEvm/user/loginZkEvmUser.ts
```diff
@@ -25,13 +25,8 @@ export const loginZkEvmUser = async ({
...
   jsonRpcProvider,
 }: LoginZkEvmUserInput): Promise<LoginZkEvmUserOutput> => {
-  let user = null;
-  try {
-    user = await authManager.getUser();
-  } catch (err) {
-    // eslint-disable-next-line no-console
-    console.warn('eth_requestAccounts` failed to retrieve a cached user session:', err);
-  }
+  let user = await authManager.getUser();
```
carmen0208, line 28: Why remove the try-catch?

packages/passport/sdk/src/guardian/guardian.ts
```diff
@@ -227,8 +229,9 @@ export default class GuardianClient {
   }
 
   private async evaluateMessage(
-    { chainID, payload, user }:GuardianMessageValidationParams,
+    { chainID, payload }:GuardianMessageValidationParams,
   ): Promise<guardian.MessageEvaluationResponse> {
+    const user = await this.authManager.getUser() as UserZkEvm;
```
carmen0208, line 234: is the `this.authManager.getUser()` can also returns null?

packages/passport/sdk/src/zkEvm/user/loginZkEvmUser.ts
scastleman-immutable: Was leftover from previous approach before I implemented feedback. Will restore it

packages/passport/sdk/src/guardian/guardian.ts
scastleman-immutable: Only if the user is not logged in, and `eth_signTypedData` checks that the user is logged in before we get here, so this call should never return null.

carmen0208: Yes, understand that would not return null as `eth_signTypedData` checks existed.
But **validateMessage** is a public method for the **guardian** class, meaning that anyone/functions can call it. We should also check/validate getUser here to avoid potential errors.

packages/passport/sdk/src/guardian/guardian.test.ts
```diff
@@ -38,15 +39,29 @@ describe('Guardian', () => {
...
+      guardianClient = new GuardianClient({
+        confirmationScreen: mockConfirmationScreen,
+        config: new PassportConfiguration({
+          baseConfig: {} as ImmutableConfiguration,
+          clientId: 'client123',
+          logoutRedirectUri: 'http://localhost:3000/logout',
+          redirectUri: 'http://localhost:3000/redirect',
+        }),
+        authManager: { getUser: jest.fn().mockResolvedValue(mockUserImx) } as unknown as AuthManager,
+      });
```
haydenfowler, line 58: Can we use `jest.clearAllMocks` here, rather than having to reinitialise the `guardianClient` again?

carmen0208 (dismissed):

carmen0208 (approved):

haydenfowler (approved):

PR #1399, opened 2024-01-24
fix: WT-2084 Patch sendTransaction overrides configuration

andrearampin:
# Summary

Checkout was incorrectly setting 
```
maxFeePerGas
maxPriorityFeePerGas
```
for L1 transactions.

With this patch we are only setting the overrides if the transaction is on L2 by checking the provider current network.

Applying the overrides to L1 made the bridge operation fail for wallets with low gas tokens.

# Impact of Change

## Fixed

Only set the overrides for L2 transactions.


packages/checkout/sdk/src/env/constants.ts
```diff
@@ -375,8 +375,8 @@ export const ERC721ABI = [
 ];
 
 // Gas overrides -- Anti-spam mechanism for when the baseFee drops low
-// https://github.com/immutable/imx-docs/blob/main/docs/main/zkEVM/overview/gas-configuration.md
+// https://docs.immutable.com/docs/zkEVM/architecture/gas-config
 export const GAS_OVERRIDES = {
-  maxFeePerGas: BigNumber.from(102e9),
-  maxPriorityFeePerGas: BigNumber.from(101e9),
+  maxFeePerGas: BigNumber.from(15e9),
+  maxPriorityFeePerGas: BigNumber.from(10e9),
```
andrearampin, line 381: new values, confirmed that this are good to go

ZacharyCouchman: Can you extract these as constants and give them a name like ZKEVM_MAX_FEE_PER_GAS etc...

packages/checkout/sdk/src/utils/utils.ts
```diff
@@ -1,3 +1,9 @@
+import { ChainId } from '../types';
+
 export const isMatchingAddress = (addressA: string = '', addressB: string = '') => (
   addressA.toLowerCase() === addressB.toLowerCase()
 );
+
+export const isZkEvmChainId = (chainId: ChainId) => chainId === ChainId.IMTBL_ZKEVM_DEVNET
+  || chainId === ChainId.IMTBL_ZKEVM_TESTNET
+  || chainId === ChainId.IMTBL_ZKEVM_MAINNET;
```
imx-mikhala, line 9: can you add jest tests

```
  describe('isZkEvmChainId', () => {
    it('should return true if devnet zkEVM chain', () => {
      const chainId = isZkEvmChainId(ChainId.IMTBL_ZKEVM_DEVNET);
      expect(chainId).toBeTruthy();
    });

    it('should return true if testnet zkEVM chain', () => {
      const chainId = isZkEvmChainId(ChainId.IMTBL_ZKEVM_TESTNET);
      expect(chainId).toBeTruthy();
    });

    it('should return true if mainnet zkEVM chain', () => {
      const chainId = isZkEvmChainId(ChainId.IMTBL_ZKEVM_MAINNET);
      expect(chainId).toBeTruthy();
    });

    it('should return false if not zkEVM chain', () => {
      const chainId = isZkEvmChainId(ChainId.SEPOLIA);
      expect(chainId).toBeFalsy();
    });
  });
```

packages/checkout/sdk/src/env/constants.ts
ZacharyCouchman: perhaps rename GAS_OVERRIDES to IMMUTABLE_ZKVEM_GAS_OVERRIDES

ZacharyCouchman (dismissed):

packages/checkout/sdk/src/utils/utils.ts
```diff
@@ -1,3 +1,9 @@
+import { ChainId } from '../types';
+
 export const isMatchingAddress = (addressA: string = '', addressB: string = '') => (
   addressA.toLowerCase() === addressB.toLowerCase()
 );
+
+export const isZkEvmChainId = (chainId: ChainId) => chainId === ChainId.IMTBL_ZKEVM_DEVNET
```
dreamoftrees, line 7: What do you think about this living in the `/network/network.ts` file? I worry utils becomes too generic as it grows.

dreamoftrees (commented):
Nice find @andrearampin 👍

packages/checkout/sdk/src/utils/utils.ts
imx-mikhala: network is focused on switching/adding networks etc and this isn't being used for those functions. it could live clsoer to where its used - `transaction/transaction.ts`

imx-mikhala (approved):

ZacharyCouchman (approved):

proletesseract (approved):
LGTM

PR #1401, opened 2024-01-24
feat: [WT-2079] balances not updating

dreamoftrees:
# Summary
Refactored Wallet Widget to use the custom hook `useBalance`. This hook handles silent refreshing of the token balances and hooks into the window focus event to force a refresh on balances.
Blockscout TTL has been reduced to 5 seconds.

https://immutable.atlassian.net/browse/WT-2079

# Impact of Change
Wallet Widget balances should be more responsive after external wallet interactions.


packages/checkout/widgets-lib/src/lib/hooks/useBalance.ts
```diff
@@ -0,0 +1,78 @@
...
+}: UseBalanceParams) => {
+  const [balancesLoading, setBalancesLoading] = useState(true);
+
+  const refreshBalances = useCallback(async (
+    // If silent is true, the balances will not be set to loading
+    silent: boolean = false,
+  ) => {
+    if (!checkout || !provider) return;
+    try {
+      const network = await checkout.getNetworkInfo({
```
andrearampin, line 31: we should have the `chainId` as a prop and if present skip this call otherwise we would have an RPC call every time we want to refresh the balances

packages/checkout/widgets-lib/src/lib/hooks/useBalance.ts
```diff
@@ -0,0 +1,78 @@
...
+        provider,
+      });
+
+      /* If the provider's network is not supported, return out of this and let the
+      connect loader handle the switch network functionality */
+      if (!network.isSupported) {
+        return;
+      }
```
andrearampin, line 39: I reckon this logic is a out-dated and unnecessary -- @ZacharyCouchman might have more to say about this

packages/checkout/widgets-lib/src/lib/hooks/useBalance.ts
```diff
@@ -0,0 +1,78 @@
...
+  useInterval(() => refreshBalances(true), REFRESH_BALANCE_INTERVAL_MS);
+
+  // Listen for window focus event to refresh after a potential interaction with a wallet
+  useEffect(() => {
+    const handleFocus = () => {
+      refreshBalances(true);
+    };
+    window.addEventListener('focus', handleFocus);
+    return () => {
+      window.removeEventListener('focus', handleFocus);
```
andrearampin, line 70: Is it possible to pass in a ref to the widget rather than attaching the listener to the window? I am thinking, what happens if the widget is simply brought back to screen by a user click, would that trigger a refresh?

dreamoftrees: @andrearampin it depends on how the widget is hidden from view by the developer. 
- If it's using the mount/unmount via our API it will trigger a refresh. 
- If the user clicks anywhere outside of the browser window (including browser extensions) it will trigger a refresh when coming back (focus event).
- If they are hiding the UI with CSS then it wont trigger a refresh until 10sec idle
- Some other events like Smart Checkout buy will flush the balance cache as well

packages/checkout/widgets-lib/src/lib/hooks/useBalance.ts
dreamoftrees: Hmmm I don't see those RPC calls every balance refresh. I'm seeing that the `getNetworkInfo` calls checks the `network` object on the provider first before doing a call. So I think it must be using cached network data for that provider instance... if the chain changes the widgets listen into that window event and rewraps the provider anyway causing a refresh, but we are assuming that is all going to be managed upstream if I remove these checks.

packages/checkout/widgets-lib/src/lib/hooks/useBalance.ts
dreamoftrees: As above, provider window events will be hooked into and this refresh will most likely always be bailed out before having to check the supported underlying chain. It's just all upstream so we happy to make that call it will always be handled elsewhere?

packages/checkout/sdk/src/api/blockscout/blockscout.ts
```diff
@@ -18,7 +18,7 @@ type CacheData = {
   ttl: number
 };
 
-const CACHE_DATA_TTL = 60; // seconds
+const CACHE_DATA_TTL = 5; // seconds
```
jhesgodi, line 21: wonderful change!!! 🙌🏾

jhesgodi (approved):

packages/checkout/widgets-lib/src/lib/hooks/useBalance.ts
imx-mikhala: If unsupported network is true and this exit is removed it looks like it can still call the sdk but the sdk itself also checks the chain is supported before making blockscout calls. The provider events set the widget to the switch network screen but I can see there are still calls to the sdk unnecessarily when this check is removed

packages/checkout/widgets-lib/src/lib/hooks/useBalance.ts
imx-mikhala: Im not seeing any rpc calls relating to this call in particular (not sure what the provider is doing around caching or whatever because I thought the call in the sdk `await web3Provider.getNetwork();` does make an rpc call)

However Im noticing a lot of rpc spam particularly when the chain id is sepolia. Just hard coded it like this:
`const balances: GetBalanceResult[] = await getTokenBalances(checkout, provider, ChainId.SEPOLIA);`
Do not see the same thing when I set to zkevm

dreamoftrees: Oh interesting, I'm also seeing those `eth_chainId` calls but only for Sepolia.
Screenshot 2024-01-25 at 1 51 35 pm

dreamoftrees: I can see the `getAllBalances` is checking for supported chains anyway.. so happy to remove this as you suggested @andrearampin .. I think if the underlying provider changes it still has checks in place and if it still goes through and is a problem the window events will intercept it anyway.
```  if (!config.networkMap.get(chainId)) {
    throw new CheckoutError(
      `chain ID ${chainId} not supported by the environment`,
      CheckoutErrorType.CHAIN_NOT_SUPPORTED_ERROR,
    );
  }```

andrearampin: Number of requests are probably due to https://github.com/immutable/ts-immutable-sdk/blob/main/packages/checkout/widgets-lib/src/widgets/wallet/views/WalletBalances.tsx#L127 -- let remove that logic, it has really little use

andrearampin (approved):

PR #1404, opened 2024-01-24
sdk config now accepts optional headers overrides

scastleman-immutable:
# Summary
optional headers overrides can now be passed through multi-rollup config, allowing the passport sample app to provide an sdk version compatible with the passport v2 endpoint changes

# Impact of Change

## Security


packages/passport/sdk-sample-app/src/context/ImmutableProvider.tsx
```diff
@@ -41,6 +41,7 @@ const getCoreSdkConfig = (environment: EnvironmentNames) => {
 
 const getPassportConfig = (environment: EnvironmentNames): PassportModuleConfiguration => {
   const sharedConfigurationValues = {
+    headers: { 'x-sdk-version': 'ts-immutable-sdk-1.0.0' },
```
carmen0208, line 44: Hey @scastleman-immutable 
Nik also mentioned the **x-sdk-version** is not overridable.
so I think that's probably not going to work

carmen0208: Here is the test to ensure it not override
https://github.com/immutable/ts-immutable-sdk/blob/main/packages/internal/generated-clients/src/config/config.test.ts#L24

scastleman-immutable: They weren't overridable, but I made changes so this gets passed through. Everything worked in testing on my end, but have a go yourself and see if they work for you too

packages/passport/sdk/src/config/config.ts
```diff
@@ -96,6 +96,7 @@ export class PassportConfiguration {
         }),
         passport: createConfig({
           basePath: overrides.passportMrBasePath,
+          headers: overrides.headers,
```
carmen0208, line 99: This can be dangerous because it is a potential leak for our SDK and risks all the API calls.

scastleman-immutable: It's part of the overrides that already allow for custom base paths, relayer URLs, magic keys and more, all of which are accessible for anyone using the SDK. Additionally, there's nothing stopping an attacker from just forking our code and manually adding whatever headers they want anyway, then using that forked code in their application. I don't see this as introducing any risks beyond what we already have as part of our codebase or architecture.

packages/passport/sdk-sample-app/src/context/ImmutableProvider.tsx
carmen0208: @CodeSchwert, can you confirm that the header is overriderable?
I did a test myself and noticed that it is overridden. But as you said to me earlier, We didn't expect **createConfig** [function ](https://github.com/immutable/ts-immutable-sdk/blob/main/packages/internal/generated-clients/src/config/config.ts#L21) overridable, 
so my understanding is even if we change the passport config to allow the config override. it still couldn't override the header. I was confused there.

packages/passport/sdk/src/config/config.ts
yundifu, line 99: yeah, i am not convinced that we should add this to fix a sample app issue. as we are exposing headers to other clients, might have risks but meantime i don't know if the headers will be used by other clients.

PR #1406, opened 2024-01-24
[Sale Widget] Fix sale widget orchestration event type

jhesgodi:
# Summary
Requesting a bridge event should originate from Sale Widget and not from Bridge Widget


mimi-imtbl (approved):

dreamoftrees (approved):
👍

PR #1409, opened 2024-01-24
fix: WT-2085 - patched onramp

andrearampin:
# Summary

We were showing the loading screen too early in the process and therefore the 3DS check screen was getting obfuscated by a Checkout (`IN_PROGRESS`) screen preventing the user to complete the onramp.

https://github.com/immutable/ts-immutable-sdk/assets/4301501/0ba46335-bb7d-449c-bf28-f0eb74c8fd75

# Impact of Change

Enable onramp with user verifications.

## Fixed

Transak event listening


packages/checkout/widgets-lib/src/widgets/on-ramp/views/OnRampMain.tsx
```diff
@@ -121,7 +121,11 @@ export function OnRampMain({
       return;
     }
 
-    if (event.event_id === TransakEvents.TRANSAK_ORDER_CREATED) {
```
andrearampin, line 124: `TransakEvents.TRANSAK_ORDER_CREATED` still used in Segment event tracking

dreamoftrees (approved):
Nice find @andrearampin .. does that mean that we will always have a 6sec delay for the progress screen?

jhesgodi (approved):
Thanks for fixing this!

andrearampin:
> Nice find @andrearampin .. does that mean that we will always have a 6sec delay for the progress screen?

Yes, the UX looked broken without that delay and I figured this is matching what Cass proposed.

PR #1421, opened 2024-01-29
fix: bridge gas limit value

wcgcyx:
# Summary

We identified that the estimates shown to the user did not match what shown on MetaMask (3x bigger). After some investigation we identified 2 core bugs:
1. We were incorrectly calculating the `getGasPriceInWei`. We have to add a tip for the miner to the `lastBaseFeePerGas` fee not the `maxFeePerGas` fee
2. Currently Bridge SDK returns an `approvalFee` as part of the response to `getFee` and `getUnsignedBridgeTx` function calls. This value is added to the `totalFee`. The `approvalFee` is always returned even if an approval request is not needed.

# Impact of Change

## Fixed

1. How we calculate `getGasPriceInWei`
2. Standardised the way Checkout fetches the estimated fees

We are looking to:
1. Patch the Bridge SDK so to return the `approvalFee` only if needed
2. Refactor Checkout to provide a `one-function` to get the unsigned transactions with the approval fees


packages/checkout/sdk/src/gasEstimate/gasPriceInWei.test.ts
```diff
@@ -4,42 +4,42 @@ import { getGasPriceInWei } from './gasPriceInWei';
 describe('gasPriceInWei', () => {
   it('should return gas price in wei using maxFeePerGas and maxPriorityFeePerGas', () => {
```
ermyas, line 5:
```suggestion
  it('should return gas price in wei using lastBaseFeePerGas and maxPriorityFeePerGas', () => {
```

packages/checkout/widgets-lib/src/widgets/bridge/components/BridgeForm.tsx
```diff
@@ -227,24 +226,44 @@ export function BridgeForm(props: BridgeFormProps) {
...
+    const tokenAddress = tokenIsNative ? NATIVE.toUpperCase() : formToken?.token.address;
+
+    const [unsignedApproveTransaction, unsignedTransaction] = await Promise.all([
+      tokenBridge!.getUnsignedApproveBridgeTx({
+        senderAddress: from!.walletAddress!,
+        token: tokenAddress ?? NATIVE.toUpperCase(),
+        amount: amountToBridge,
+        sourceChainId: from!.network.toString(),
+        destinationChainId: to!.network.toString(),
+      }),
```
ermyas, line 243: Should we omit this if the transfer is for a native asset? (i.e. `transferIsNative` is `true`?)

packages/internal/bridge/sdk/src/lib/gasPriceInWei.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    const fee = getGasPriceInWei({
+      lastBaseFeePerGas: null,
+      maxFeePerGas: null,
+      maxPriorityFeePerGas: null,
+      gasPrice: BigNumber.from(11),
+    });
+    expect(fee).toEqual(BigNumber.from(11));
+  });
+
+  it('should return gas price in wei when maxFeePerGas missing', () => {
```
ermyas, line 25:
```suggestion
  it('should return gas price in wei when lastBaseFeePerGas missing', () => {
```

ermyas (dismissed):

packages/checkout/widgets-lib/src/widgets/bridge/components/BridgeForm.tsx
andrearampin: I guess we could, however, could we perhaps treat it as a separate PR because this will need further testing given that  I will need to make these changes across two screens. Right now I have copied the logic from a view that has already been tested.

packages/checkout/sdk/src/gasEstimate/gasPriceInWei.test.ts
```diff
@@ -2,44 +2,44 @@ import { BigNumber } from 'ethers';
...
       maxFeePerGas: null,
       maxPriorityFeePerGas: null,
-      gasPrice: BigNumber.from(1),
+      gasPrice: BigNumber.from(11),
     });
-    expect(fee).toEqual(BigNumber.from(1));
+    expect(fee).toEqual(BigNumber.from(11));
   });
 
   it('should return gas price in wei when maxFeePerGas missing', () => {
```
ermyas, line 25:
```suggestion
  it('should return gas price in wei when lastBaseFeePerGas missing', () => {
```

andrearampin (approved):

ermyas (approved):

PR #1425, opened 2024-01-30
fix: [GPR-347][Sale widget] Handle failed/success events for individual transactions

jhesgodi:
# Summary
After a transaction is executed or cancelled handle the necessary events.

## Fixed
* A successful transaction event was being sent every time regardless of transaction completion

- [x] Prefix your PR title with `breaking-change:` if you have introduced breaking changes to this SDK. Breaking changes are modification that necessitates immediate adjustments by this SDK's users to their applications, clients, or integrations to avert disruptions to existing features or functionalities. Ensure you write clear summary explain the change.

### DIFF
--- a/packages/checkout/widgets-lib/src/widgets/sale/hooks/useSignOrder.ts
+++ b/packages/checkout/widgets-lib/src/widgets/sale/hooks/useSignOrder.ts
@@ -183,7 +183,7 @@ export const useSignOrder = (input: SignOrderInput) => {
       data: string,
       gasLimit: number,
       method: string,
-    ): Promise<string | undefined> => {
+    ): Promise<[error: any, hash: string | undefined]> => {
       let transactionHash: string | undefined;
 
       try {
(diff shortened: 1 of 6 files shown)


mimi-imtbl (approved):

packages/checkout/widgets-lib/src/widgets/sale/hooks/useSignOrder.ts
```diff
@@ -183,7 +183,7 @@ export const useSignOrder = (input: SignOrderInput) => {
       data: string,
       gasLimit: number,
       method: string,
-    ): Promise<string | undefined> => {
+    ): Promise<[error: any, hash: string | undefined]> => {
```
imx-mikhala, line 186: non-blocking - normally i would expect the error to be the second option here as the hash is more important

imx-mikhala (approved):

packages/checkout/widgets-lib/src/widgets/sale/hooks/useSignOrder.ts
jhesgodi: thank you, will do the fix.

imx-mikhala (approved):

mimi-imtbl (approved):

PR #1436, opened 2024-02-01
feat: script to update sdk version, additional husky rule

scastleman-immutable:
# Summary
New script to update the SDK version 

## Added 
Script in sdk-sample-app to update __SDK_VERSION__ to latest git tag, updated tests to ensure this version does not get committed. Added husky hook to test project to prevent projects with failing tests being committed as further safeguard.

Needs pipeline change before deployed sample app will function.


.github/workflows/passport-sdk-sample-app-deployment.yaml
```diff
@@ -46,6 +46,9 @@ jobs:
       - name: Install dependencies
         run: yarn install --immutable
 
+      - name: Ppdate SDK version
+        run: yarn workspace @imtbl/passport-sdk-sample-app update-sdk-version
```
yundifu, line 50: does this work on CI now ?

yundifu (approved):

carmen0208 (approved):

scastleman-immutable:
> LGTM, but this seems quite specific for the passport package. I wonder whether other packages will run into this issue. Was this the proposed approach from the SDK team?

Yes, Carmen spoke with Nik Ho and he recommended this approach.

.husky/pre-commit
```diff
@@ -2,3 +2,4 @@
 . "$(dirname -- "$0")/_/husky.sh"
 
 yarn lint-staged
+yarn test --watchAll=false
```
cajames, line 5: Why was this added?

PR #1441, opened 2024-02-02
WT-2017 Split checkout bundle by widget

imx-mikhala:
Confluence documentation around this implementation [can be read here](https://immutable.atlassian.net/wiki/spaces/WT/pages/2409267532/Widgets+Split+Bundle)

Currently we have a singular `widgets.js` file that contains the entire bundle.

This PR adds `Suspense` components at the widget level to split out the widgets into a new `widgets-esm.js` bundle. This is an opt-in way of loading the widgets and is used when getting the widgets factory and setting the `useEsModules: true` flag. This is to ensure we continue to support the `umd` build for those that are integrating and are unable to support es modules.

The files can be seen here on jsdelivr https://cdn.jsdelivr.net/npm/@imtbl/sdk@1.6.0-alpha.1/dist/browser/checkout/
Before we used to only serve the widgets via `widgets.js`. Now we have the original umd bundle `widgets.js` plus the split bundle with the entry point being `widgets-esm.js`.

## Testing

The PR can be tested out locally on the playground using the npm version `1.6.0-alpha.1`, which is a release on npm of this branch. Make sure to force the version locally in `WidgetContext.tsx` and delete the next cache. Also may need to remove the ^ when installing the dep in the package.json or it brings down the non-alpha `"@imtbl/sdk": "1.6.0-alpha.1",`

Check in the network tab that you are pulling the `widgets-esm.js` on playground to check the right version is being tested.

```
const widgetsFactory = await checkoutInstance.widgets({
          config: { theme: checkout.WidgetTheme.DARK },
          version: {
            major: 1,
            minor: 6,
            patch: 0,
            prerelease: 'alpha',
            build: 1,
          },
        });
```

It is also possible to test the umd fallback by removing the `/* webpackIgnore: true */` in the index.js in the `node_modules/@imtbl/sdk/dist` in the widgets function, but ensure you delete the next cache and rerun `npm run dev` again


packages/checkout/widgets-lib/rollup.config.js
```diff
@@ -23,16 +23,17 @@ export default [
...
   {
     watch: false,
     input: 'src/index.ts',
     output: {
-      file: 'dist/widgets.js',
-      format: 'umd',
-      name: 'ImmutableCheckoutWidgets',
-      inlineDynamicImports: true
+      dir: 'dist',
+      format: 'es',
```
imx-mikhala, line 35: umd -> es is required for bundle splitting

sdk/scripts/copyBrowserBundles.js
```diff
@@ -53,6 +53,23 @@ const main = () => {
...
+            if (path.extname(chunkFile) === '.js') {
+              // Skip copying the original file and only copy .js chunks
+              if (chunkFile !== path.basename(sourceFile)) {
+                const chunkSrcPath = path.join(srcDirectory, chunkFile);
+                const chunkDestPath = path.join(directoryPath, chunkFile);
+                fs.copyFileSync(chunkSrcPath, chunkDestPath);
+              }
+            }
+          });
+        }
```
imx-mikhala, line 72: This file copies over the widgets build into the dist that gets put on jsdelivr. This bit of code ensures all the chunks are also copied over.

sdk/module-release.json
```diff
@@ -16,9 +16,10 @@
   },
   "fileCopy": [
     {
-      "src": "../packages/checkout/widgets-lib/dist/widgets.js",
+      "src": "../packages/checkout/widgets-lib/dist/index.js",
```
imx-mikhala, line 19: The build was outputting a file called `widgets.js` inside widgets-lib dist, but with split builds + rollup it did not seem easy/straightforward to rename the index.js file into widgets.js - so the entry file in the widgets-lib dist is just index.js but will still be renamed to widgets.js when the file ends up on jsdelivr.

packages/checkout/sdk/src/sdk.ts
```diff
@@ -119,22 +121,40 @@ export class Checkout {
...
+  }
+
+  private async loadUmdBundle(
+    config: WidgetConfiguration,
+    version?: SemanticVersion,
+  ) {
+    const checkout = this;
+
     const factory = new Promise<ImmutableCheckoutWidgets.WidgetsFactory>((resolve, reject) => {
       function checkForWidgetsBundleLoaded() {
```
dreamoftrees, line 143: One thing I think worth introducing here is `bundleLoadAttempts` as an incremental number. Each time `checkForWidgetsBundleLoaded` is called and we can increment that.. after x number of attempts we throw an error and bail out? Just so we don't get stuck forever..

dreamoftrees (approved):
Awesome work!

packages/checkout/sdk/src/sdk.ts
imx-mikhala: Yeah good idea - I'll add something in as part of this PR to break out after ~10 attempts or so

imx-mikhala: Hm looks like on slower networks this can run for a bit so its hard to actually put a number here on attempts (took like ~60 attempts on 'fast 3g').

dreamoftrees: Yep sounds good, timings like this are always hard.

dreamoftrees (approved):
👍

carleeto (approved):

packages/checkout/sdk/src/sdk.ts
ZacharyCouchman: Hey @imx-mikhala , one change I'd love to make here that we didn't do last time was adding the script onload event listener to resolve the promise that way. I did some testing on this and it worked correctly.

packages/checkout/sdk/src/sdk.ts
```diff
@@ -152,6 +172,30 @@ export class Checkout {
...
+    config: WidgetConfiguration,
+    version?: SemanticVersion,
+  ) {
+    const checkout = this;
+    try {
+      const cdnUrl = getWidgetsEsmUrl(version);
+
+      // WebpackIgnore comment required to prevent webpack modifying the import statement and
+      // breaking the dynamic import in certain applications integrating checkout
+      const checkoutWidgetsModule = await import(/* webpackIgnore: true */ cdnUrl);
```
ZacharyCouchman, line 185: This looks like a much nicer way to load and wait for the script 👍

packages/checkout/sdk/src/sdk.ts
ZacharyCouchman: The problem with this last time is that we were only creating the event listener after the script had loaded. We need to create the script element. Hook into the on load event and resolve the promise from in there. And then add script.src which will start loading it.

packages/checkout/widgets-lib/src/widgets/connect/ConnectWidgetRoot.tsx
```diff
@@ -24,15 +27,18 @@ export class Connect extends Base<WidgetType.CONNECT> {
...
 
     this.reactRoot.render(
       <React.StrictMode>
         <CustomAnalyticsProvider checkout={this.checkout}>
           <ThemeProvider id="connect-container" config={this.strongConfig()}>
-            <ConnectWidget
-              config={this.strongConfig()}
-              checkout={this.checkout}
-            />
+            <Suspense fallback={<LoadingView loadingText={t('views.LOADING_VIEW.text')} />}>
```
ZacharyCouchman, line 36: Just wondering if this adds another loading view flicker while we wait for the bundle to load. I'll pull down and test

imx-mikhala: Locally for me when I tested it looked like the spinner reset its location when it moved to the next widget's loading screen but I didn't see a flicker when changing from the connect -> the widget other than that. But yep please pull down and try it out 👍 I did this on a slower network. Before I added the fallback there was a while where the widget was removed from the screen before the actual widget loaded (like when you are going from connect -> swap for example)

dreamoftrees (approved):
Great work @imx-mikhala 👍

carleeto (approved):

PR #1442, opened 2024-02-04
Small bridge txn fixes

ZacharyCouchman:
## Bridge fix

When clicking back from the Transactions screen, reset the to wallet in state so that wallets are not pre-filled in the wallet network selector.

Also, re-fetch transactions on wallet change in the transactions screen.


sharonsheah (approved):
Looks good!! 🚀

deepti-imx (approved):

PR #1443, opened 2024-02-04
refactor: [WT-2104] Swap and Move fees breakdown

dreamoftrees:
Swap and Move Widget fee component refactor. https://immutable.atlassian.net/browse/WT-2104
The Widgets gas estimate fetching has been updated to use CancellablePromises to ensure only the latest quotes are used within the widgets.

# Summary
Swap and Move Widget fees include an updated fee breakdown when required (service, approval, and gas fees)
Swap and Move Widget inputs are non-blocking while they fetch quotes.

Screenshot 2024-02-05 at 10 14 49 am

Screenshot 2024-02-05 at 10 15 07 am


packages/checkout/widgets-lib/src/lib/hooks/useDebounce.ts
```diff
@@ -0,0 +1,20 @@
+import { useState, useEffect } from 'react';
+
+const useDebounce = (value, delay = 500) => {
```
ZacharyCouchman, line 3: Any reason you've gone with making this a hook with useEffect rather than just a function like debounce(fn, delay)?

dreamoftrees: My thinking is that a custom hook encapsulates the functionality well and cleans up after itself hooking into Reacts lifecycle, the other benefit is hooking into Reacts state/effects will then easily react to changes in `value` and `delay` without having to manage any of that in every instance it's used.. flushing and syncing the debounce.

dreamoftrees: Just been thinking of ways we can leverage React more with things like custom hooks to reduce boilerplate and shared functionality across our Widgets.. this felt like a nice baby step towards that :)

packages/checkout/widgets-lib/src/components/Fees/FeeStyles.ts
```diff
@@ -10,3 +10,12 @@ export const feeBoxStyles = {
   display: 'flex',
   flexDirection: 'column',
 };
+
+export const gasAmountAccordionStyles = {
+  backgroundColor: 'base.color.translucent.emphasis.100',
+};
```
imx-mikhala, line 16: test

packages/checkout/widgets-lib/src/lib/hooks/useDebounce.ts
ZacharyCouchman: My thinking was around the fact that we will be calling the function which sets the value to be used in debounce over and over and only the debouncedValue will update after the delay, versus, debouncing the actual function which will only be called once (after the delay) to set the value.

With how we are using this in the swap widget, we will call setAmount() on every input keystroke and then only the debouncedValue updates after the delay and I was expecting that we would debounce the inputChange so that setAmount() is only called once after the delay.

The difference will be in the amount of re-renders happening as we are still setting the 'amount' piece of state over and over.

packages/checkout/widgets-lib/src/components/Fees/Fees.tsx
```diff
@@ -1,84 +1,101 @@
...
 
 export function Fees({
-  title, fiatPricePrefix, gasFeeValue, gasFeeToken, gasFeeFiatValue,
+  gasFeeValue, gasFeeToken, gasFeeFiatValue, fees, onFeesClick, loading, sx,
 }: FeesProps) {
+  const [showFeeBreakdown, setShowFeeBreakdown] = useState(false);
   const { t } = useTranslation();
-  const { track } = useAnalytics();
+  // eslint-disable-next-line no-param-reassign
+  if (!gasFeeValue && !loading) return <Box sx={...sx} />;
```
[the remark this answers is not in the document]
dreamoftrees: Hmm strange. I've simplified that code a little and removed the obsolete eslint rule.. I'm hoping cleans that up.

packages/checkout/widgets-lib/src/widgets/swap/components/SwapForm.tsx
```diff
@@ -256,16 +280,17 @@ export function SwapForm({ data }: SwapFromProps) {
         toToken,
       );
 
-      const resolved = await Promise.all([quoteResultPromise, conversionResultPromise]);
+      const currentQuoteRequest = CancellablePromise.all<any>([
```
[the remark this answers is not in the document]
dreamoftrees: Nice catch, cleaned up these types.

packages/checkout/widgets-lib/src/components/Fees/Fees.tsx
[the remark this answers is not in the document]
imx-mikhala: think thats fixed it 👍

packages/checkout/widgets-lib/src/lib/hooks/useDebounce.ts
dreamoftrees: It's a good callout.. from my experience delaying the data change has other impacts to the UX that can be harder to map to, but the fetch is really the key of what we need to debounce.

imx-mikhala: just from testing out locally seems like the re-rendering isnt too bad

ZacharyCouchman (approved):
👏

imx-mikhala (approved):

ZacharyCouchman (approved):

PR #1444, opened 2024-02-05
fix: NOJIRA passport dev fix

dom-murray:

# Summary
The current overrides prevents us from using the dev environment with Passport outside of the SDK because `ImxApiClients` is an internal package. This fixes that by using the existing `imxPublicApiDomain` override to create the client. 

## Security


scastleman-immutable (dismissed):

keithbro-imx (approved):
Ideally would have a test that demonstrates the override not needing ImxApiClients but won't block you as I know this can be tricky sometimes.

keithbro-imx (approved):

PR #1445, opened 2024-02-05
Add a test to check basePath on imxApiClients

keithbro-imx:


packages/passport/sdk/src/Passport.ts
```diff
@@ -45,35 +98,16 @@ export class Passport {
...
-    this.passportImxProviderFactory = new PassportImxProviderFactory({
-      authManager: this.authManager,
-      config: this.config,
-      confirmationScreen: this.confirmationScreen,
-      immutableXClient: this.immutableXClient,
-      magicAdapter: this.magicAdapter,
-      passportEventEmitter: this.passportEventEmitter,
-      imxApiClients,
-    });
+    const privateVars = buildPrivateVars(passportModuleConfiguration);
```
keithbro-imx, line 101: Extract it out to make it testable

packages/passport/sdk/src/starkEx/passportImxProviderFactory.ts
```diff
@@ -33,7 +33,7 @@ export class PassportImxProviderFactory {
 
   private readonly passportEventEmitter: TypedEventEmitter<PassportEventMap>;
 
-  private readonly imxApiClients: ImxApiClients;
+  public readonly imxApiClients: ImxApiClients;
```
keithbro-imx, line 36: Correct me if I'm wrong but this class is not exposed to users, therefore no harm in making this public (note still readonly). Allows for testability.

packages/passport/sdk/src/types.ts
```diff
@@ -58,12 +58,12 @@ export interface PassportOverrides {
   passportDomain: string;
   imxPublicApiDomain: string;
   immutableXClient: IMXClient;
-  imxApiClients: ImxApiClients;
   zkEvmRpcUrl: string;
   relayerUrl: string;
   indexerMrBasePath: string;
   orderBookMrBasePath: string;
   passportMrBasePath: string;
+  imxApiClients?: ImxApiClients; // needs to be optional because ImxApiClients is not exposed publicly
```
keithbro-imx, line 66: only packages within the SDK can provide this so need to be optional

scastleman-immutable (approved):

PR #1454, opened 2024-02-07
fix: [NO-JIRA] Support for Checkout Widget cypress test translations

dreamoftrees:
Fixes to Checkout Widget Cypress tests to support non blocking gas estimates and translations.


packages/checkout/widgets-lib/cypress/support/commands.ts
```diff
@@ -35,3 +35,19 @@
 //     }
 //   }
 // }
```
imx-mikhala, line 37: if we are gonna be using this file now lets remove some of this comment clutter

imx-mikhala (dismissed):

sharonsheah (approved):

PR #1456, opened 2024-02-08
fix:[TP-1924] gas includes secondary fee

scastleman-immutable:
# Summary

We would like to include an estimate for the extra gas cost of secondary fees. This doesn't have to be overly-complicated. Using a few swaps I got the approx difference in gas.

Added extra gas cost if a swap is using secondary fees.


Benjimmutable (dismissed):

packages/internal/dex/sdk/src/lib/transactionUtils/swap.ts
```diff
@@ -474,7 +474,9 @@ export function getSwap(
   );
 
   // TODO: Add additional gas fee estimates for secondary fees
```
pano-skylakis, line 476: We can remove the TODO now 🎉

Benjimmutable: very true. deleted

pano-skylakis (approved):

Benjimmutable (approved):

packages/internal/dex/sdk/src/lib/transactionUtils/approval.ts
```diff
@@ -157,7 +157,7 @@ export const getApproval = async (
     preparedApproval.amount.token.address,
   );
 
-  const gasFeeEstimate = gasPrice ? calculateGasFee(gasPrice, gasEstimate) : null;
+  const gasFeeEstimate = gasPrice ? calculateGasFee(false, gasPrice, gasEstimate) : null;
```
keithbro-imx, line 160: Boolean arguments are not considered very ergonomic. Looking at this line in isolation, without context I would have no idea what `false` does or means. Some alternatives:

* Use a separate function e.g. have both `calculateGasFee` and `calculateGasFeeViaSwapProxy()`
* Use an object for options e.g. `calculateGasFee(gasPrice, gasEstimate, { usesSwapProxy: true })`

Both of these help the reader understand what's happening.

PR #1459, opened 2024-02-08
feat: [WT-2101] Map token icons in Checkout Widgets

dreamoftrees:
Updates to show token icons in Wallet, Bridge, and Swap Widgets.
https://immutable.atlassian.net/browse/WT-2101

# Summary
Checkout Widgets now displays token icons.

## Demo
Screenshot 2024-02-09 at 12 02 26 pm

### DIFF
--- a/packages/checkout/widgets-lib/src/lib/balance.ts
+++ b/packages/checkout/widgets-lib/src/lib/balance.ts
@@ -61,7 +61,29 @@ export const getAllowedBalances = async ({
     type: allowTokenListType,
   });
 
-  // Map token icon assets
+  const tokensAddresses = new Map();
+  allowList.tokens.forEach((token) => tokensAddresses.set(token.address?.toLowerCase() || NATIVE, true));
+
+  const allowedBalances = tokenBalances.balances
+    .filter((balance) => {
+      // Balance is <= 0 and it is not allow to have zeros
+      if (balance.balance.lte(0) && !allowZero) return false;
+      return tokensAddresses.get(balance.token.address?.toLowerCase() || NATIVE);
+    })
+    .map((balanceResult) => ({
+      ...balanceResult,
+      token: {
+        ...balanceResult.token,
+        icon: getTokenImageByAddress(
+          checkout.config.environment as Environment,
+          isNativeToken(balanceResult.token.address)
+            ? balanceResult.token.symbol.toLowerCase()
+            : balanceResult.token.address ?? '',
+        ),
+      },
+    })) ?? [];
+
+  // Map token icon assets to allowlist
   allowList.tokens = allowList.tokens.map((token) => ({
     ...token,
     icon: getTokenImageByAddress(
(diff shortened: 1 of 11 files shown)


packages/checkout/widgets-lib/src/lib/balance.ts
```diff
@@ -59,6 +61,15 @@ export const getAllowedBalances = async ({
     type: allowTokenListType,
   });
 
+  // Map token icon assets
+  allowList.tokens = allowList.tokens.map((token) => ({
+    ...token,
+    icon: getTokenImageByAddress(
+      checkout.config.environment as Environment,
+      isNativeToken(token.address) ? token.symbol.toLowerCase() : token.address ?? '',
```
ZacharyCouchman, line 69: nit: the toLowerCase() here is unneccessary as it will happen in the function itself

packages/checkout/widgets-lib/src/widgets/wallet/components/BalanceItem/BalanceItem.tsx
```diff
@@ -84,7 +84,10 @@ export function BalanceItem({
 
   return (
     <MenuItem testId={`balance-item-${balanceInfo.symbol}`} emphasized>
-      <MenuItem.FramedIcon icon="Coins" circularFrame />
+      <MenuItem.FramedImage
+        imageUrl={balanceInfo.iconLogo}
```
ZacharyCouchman, line 88: @dreamoftrees  can we double check if this is meant to be icon or iconLogo?

packages/checkout/widgets-lib/src/lib/balance.ts
dreamoftrees: In this case I'm just making sure because its mapping to the svg URL.. which I guess shouldn't be case sensative.. but the imx.svg and eth.svg are lowercase and prefer to map that exactly.

packages/checkout/widgets-lib/src/widgets/wallet/components/BalanceItem/BalanceItem.tsx
ZacharyCouchman: ah we do the mapping in tokenBalances function

ZacharyCouchman (approved):

packages/checkout/widgets-lib/src/widgets/wallet/components/BalanceItem/BalanceItem.tsx
dreamoftrees: The underlying TokenInfo is `icon` but in the `BalanceInfo` class we are mapping it to `iconLogo` .. ideally its consistent using `icon` but didn't want to refactor that right now so left it mapping across.

ZacharyCouchman:
@dreamoftrees, can you add the icons to TransactionItem and TransactionItemWithdrawalPending components?

dreamoftrees:
> @dreamoftrees, can you add the icons to TransactionItem and TransactionItemWithdrawalPending components?

Yep good call have updated to support these as well.

packages/checkout/widgets-lib/src/widgets/wallet/components/BalanceItem/BalanceItem.tsx
dreamoftrees: Ok refactored this to use `icon` consistently.. feels a lot cleaner.

packages/checkout/widgets-lib/src/lib/balance.ts
dreamoftrees: Ahh cool .. you are correct! sorry missed that deeper in.. taken this out. Nice catch!

ZacharyCouchman (approved):

imx-mikhala (approved):

PR #1465, opened 2024-02-12
fix: Move Widget gas estimate flow optimisation

dreamoftrees:
# Summary
Removed the initial gas estimates in the Move Widget amount view. 
Added `enter` key down event to speed up form submission.

## Removed
The Move Widget amount input view no longer includes a gas fee estimate.

### DIFF
--- a/packages/checkout/widgets-lib/src/widgets/bridge/components/BridgeForm.tsx
+++ b/packages/checkout/widgets-lib/src/widgets/bridge/components/BridgeForm.tsx
@@ -5,17 +5,13 @@ import {
   OptionKey,
 } from '@biom3/react';
 import {
-  GasEstimateBridgeToL2Result,
-  GasEstimateType,
   GetBalanceResult,
 } from '@imtbl/checkout-sdk';
 import {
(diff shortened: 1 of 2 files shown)


sharonsheah (approved):
LGTM - tested in sample app and working well

packages/checkout/widgets-lib/src/widgets/bridge/components/BridgeForm.tsx
```diff
@@ -207,125 +189,11 @@ export function BridgeForm(props: BridgeFormProps) {
...
-    }
-    setLoading(false);
-
-    setGasFee('');
-    setGasFeeFiatValue('');
-  };
-
   const insufficientFundsForGas = useMemo(() => {
     const nativeTokenBalance = tokenBalances
       .find((balance) => isNativeToken(balance.token.address));
```
ZacharyCouchman, line 194: Do we need this here anymore if we can't compare against the gas?

ZacharyCouchman (approved):

imx-mikhala (approved):

PR #1472, opened 2024-02-14
refactor: [GPR-399][Sale Widget] Update sale widget to use currency from currency API

mimi-imtbl:
# Summary

Sale widget users will not have to pass the currency token address in the URL query params. This is now handled inside the widget where it's getting the token address from the BE.


packages/checkout/widgets-lib/src/widgets/sale/hooks/useCurrency.ts
```diff
@@ -0,0 +1,58 @@
...
+import { useState, useCallback } from 'react';
+
+type CurrencyResponse = {
+  decimals: number;
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  erc20_address: string;
+  name: string;
+};
+
+const PRIMARY_SALES_API_BASE_URL = {
```
jwhardwick, line 11: This is already defined in `useSignOrder`

packages/checkout/widgets-lib/src/widgets/sale/hooks/useCurrency.ts
```diff
@@ -0,0 +1,58 @@
...
+        throw new Error('Failed to fetch currency');
+      }
+      setCurrencyResponse(data.currencies);
+
+      return data.currencies;
+    } catch (error) {
+      setCurrencyError(error as Error);
+    }
+
+    return undefined;
```
jwhardwick, line 54: I'm a bit confused with the pattern we're using here - why is this in a `useCallback` and in a hook? And then why are we using `currencyResponse/currencyError ` - I don't see that being consumed anywhere in the code.

The two places that use it are calling `fetchCurrency` and then processing the response and handling errors - if we're going to use a hook the hook should handle all this.

If I understand correctly the currency value only depends on the `env` and `environmentId`, and we only want to call it once per load. What do you think about calling the currency API once on app load inside the `SaleContextProvider` (do any data processing + error handling), and then exposing that value in the provider? The components that consume it can add that as a dependancy.

mimi-imtbl: great points, re: useCallback in a hook - the main reason to extract the currency call into a custom hook and using `useCallback` is for code organization and reusability. 

But having `currencyResponse/currencyError` can help with a component that instantiates the hook to reuse the value instead of refetching it. 

I might have overengineered it here, as the currency endpoint is currently being called twice - one for `/sign` and one for smart checkout. Once per load sounds reasonable.

@jhesgodi what are your thoughts on this? having a hook or no hook?

packages/checkout/widgets-lib/src/widgets/sale/hooks/useCurrency.ts
jhesgodi: @mimi-imtbl lets move this up to utils/config

packages/checkout/widgets-lib/src/widgets/sale/hooks/useCurrency.ts
jhesgodi: Thanks guys, overall having a hook should the right thing, However, I must agree that the behaviour is not the ideal given that is exposing the fetch to the consumer.

From my POV this could be a better way to setup this hook:
- ** Hook goal:** to fetch and store available currencies from API, and expose selected one
- ** Hook name:** base on the goal the name should be `useSelectCurrency` and the returned value `{ tokenAddress }`
- ** Hook behaviour:**  fetch all currencies, and returns the selected one by token name
```
const { currency } = useSelectCurrency({ name: 'USDC' });
const { decimals, erc20Address: fromTokenAddress } = currency;

// Notes:
// Loading should be deferred until currency was set.
// If currency was not found by name, the it should return the first one found (in our current state, this will be trivial as USDC will always be first but also be available to select, however adding the select by name ensure we switch to default a different currency in the future without refactors)

```

jhesgodi (changes requested):

packages/checkout/widgets-lib/src/widgets/sale/hooks/useCurrency.ts
jhesgodi: **Regarding error handling**
I must also agree, this hook should not need to expose the errors to the consumer.
Instead, handle errors internally to default to an available currency.

If no currencies were found (backend fails or returns empty array), the upper View component should handle this by redirecting to SaleErrorView with type: `SaleErrorTypes.INVALID_PARAMETERS`. As this error view will block the flow preventing user from progressing with a misconfigured params.

packages/checkout/sdk/src/widgets/definitions/parameters/sale.ts
```diff
@@ -16,8 +15,6 @@ export type SaleWidgetParams = {
   amount?: string;
   /** Environment id from Immutable Hub */
   environmentId?: string;
-  /** Contract address of the token accepted for payments */
-  fromTokenAddress?: string;
```
jhesgodi, line 20: @mimi-imtbl just a note to remember to update the docs.
https://github.com/immutable/docs/pull/805

packages/checkout/widgets-lib/src/widgets/sale/SaleWidget.tsx
```diff
@@ -54,14 +62,33 @@ export function SaleWidget(props: SaleWidgetProps) {
...
+    currencyName: CURRENCY_NAME,
+  });
+
+  const [fromTokenAddress, setFromTokenAddress] = useState('');
 
-  const loadingText = viewState.view.data?.loadingText
-    || t('views.LOADING_VIEW.text');
+  const loadingText = viewState.view.data?.loadingText || t('views.LOADING_VIEW.text');
+
+  useEffect(() => {
```
jhesgodi, line 85: @mimi-imtbl setting a state again is not really necessary. as you can destructure currency values here
```  const { erc20Address: fromTokenAddress } = useCurrency({
    env,
    environmentId,
    currencyName: CURRENCY_NAME,
  });```

**Review by jhesgodi — CHANGES_REQUESTED** [review]

File: packages/checkout/widgets-lib/src/widgets/sale/SaleWidget.tsx [hunk 5]
```diff
@@ -31,20 +39,19 @@ import { TopUpView } from '../../views/top-up/TopUpView';
 import { UserJourney } from '../../context/analytics-provider/SegmentAnalyticsProvider';
 import { sendSaleWidgetCloseEvent } from './SaleWidgetEvents';
 import { EventTargetContext } from '../../context/event-target-context/EventTargetContext';
+import { useCurrency } from './hooks/useCurrency';
 
-export interface SaleWidgetProps extends Required<Omit<SaleWidgetParams, 'walletProviderName'>> {
+const CURRENCY_NAME = 'USDC';
```
jhesgodi, line 44: please move this one to the config file

packages/checkout/widgets-lib/src/widgets/sale/SaleWidget.tsx
```diff
@@ -54,14 +61,25 @@ export default function SaleWidget(props: SaleWidgetProps) {
     eventTargetState: { eventTarget },
   } = useContext(EventTargetContext);
 
+  const env = checkout!.config.environment ?? Environment.SANDBOX;
```
jhesgodi, line 64: this is already at `config.environment`

packages/checkout/widgets-lib/src/widgets/sale/hooks/useCurrency.ts
```diff
@@ -0,0 +1,58 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import { useState, useEffect } from 'react';
+import { PRIMARY_SALES_API_BASE_URL } from '../utils/config';
+
+type CurrencyResponse = {
+  decimals: number;
+  erc20_address: string;
```
jhesgodi, line 7: can you please re-map the return object to be camelCase?

packages/checkout/widgets-lib/src/widgets/sale/utils/config.ts
```diff
@@ -0,0 +1,6 @@
+import { Environment } from '@imtbl/config';
```
jhesgodi, line 1: please also move the default currency to this file

packages/checkout/widgets-lib/src/widgets/sale/hooks/useCurrency.ts
```diff
@@ -0,0 +1,58 @@
...
+        const response = await fetch(baseUrl, {
+          method: 'GET',
+          headers: {
+            'Content-Type': 'application/json',
+          },
+        });
+
+        const data = await response.json();
+
+        let selectedCurrency;
```
jhesgodi, line 39: this block can be simplified as the checks are not necessary for find to work
```
const selectedCurrency = data.currencies.find((c: CurrencyResponse) => c.name === currencyName) || data.currencies[0];

setCurrencyResponse(selectedCurrency);
```

packages/checkout/widgets-lib/src/widgets/sale/hooks/useCurrency.ts
```diff
@@ -0,0 +1,58 @@
...
+  env,
+  environmentId,
+  currencyName,
+}: UseCurrencyParams) => {
+  const [currencyResponse, setCurrencyResponse] = useState<
+  CurrencyResponse | undefined
+  >(undefined);
+
+  useEffect(() => {
+    const fetchCurrency = async () => {
```
jhesgodi, line 27: before this line you should consider checking if `environmentId` or `env` are undefined. if so fail early

packages/checkout/sdk/src/widgets/definitions/parameters/sale.ts
mimi-imtbl: done

sharonsheah (approved):

sharonsheah (approved):

jhesgodi (approved):

PR #1473, opened 2024-02-14
fix: [WT-2150] Checkout Widget amount inputs correctly handle decimal inputs

dreamoftrees:
# Summary
Checkout Widgets (Swap and Move) input validator fix.
https://immutable.atlassian.net/browse/WT-2150

## Fixed
Fix for an issue where Checkout Widgets input validator would not support number inputs starting with decimals.


sharonsheah (approved):
Nice work!! 🚀

ZacharyCouchman (approved):

PR #1475, opened 2024-02-14
feat: EXP add cross product identity matching

shineli1984:

# Summary
Add ability for passport SDK consumer to pass in an identity string to connect with immutable's authentication system.

## Added 

Allow an anonymous id to be passed to passport login flows so that the user identity tracking can be aligned on the caller and immutable's data set.


oliyoung (approved):

packages/internal/metrics/src/index.ts
```diff
@@ -5,3 +5,5 @@ export {
   setPassportClientId,
   setPublishableApiKey,
 } from './details';
+export { getDetail } from './utils/state';
+export { Detail } from './utils/constants';
```
cajames, line 9: This won't work as expected due to possible multiple instantiations of the SDK. You might get a different "details" store.

cajames (approved):

oliyoung (approved):

PR #1487, opened 2024-02-15
feat: EXP cross product identity matching - bug fix

shineli1984:

# Summary

Add ability for passport SDK consumer to pass in an identity string to connect with immutable's authentication system.

## Added 

Allow an anonymous id to be passed to passport login flows so that the user identity tracking can be aligned on the caller and immutable's data set.


packages/passport/sdk/src/authManager.ts
```diff
@@ -153,10 +154,20 @@ export default class AuthManager {
...
+   * login
+   * @param anonymousId Caller can pass an anonymousId if they want to associate their user's identity with immutable's internal instrumentation.
+   */
+  public async login(anonymousId?: string): Promise<User> {
     return withPassportError<User>(async () => {
+      const rid = getDetail(Detail.RUNTIME_ID);
       const popupWindowFeatures = { width: 410, height: 450 };
       const oidcUser = await this.userManager.signinPopup({
+        extraQueryParams: {
+          ...this.userManager.settings.extraQueryParams,
```
shineli1984, line 167: oidc-client-ts uses [default value](https://github.com/authts/oidc-client-ts/blob/74b6b9624562796d704e51edc950f3eea0fc566d/src/OidcClient.ts#L114) for extraQueryParams from settings when there is no extraQueryParams passed in. This approach leads to any additional extraQueryParams needs to be passed in here with the `this.userManager.settings.extraQueryParams`.

shineli1984: Tested with sample app and the access token now is valid.

haydenfowler (approved):

carleeto (approved):

PR #1494, opened 2024-02-16
chore: ID-1423 Remove redundant userInfo auth request

Jon-Alonso:
# Summary
Removed a redundant [userInfo](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo) request that was being triggered on every token generation request. Instead, it reads the values from the ID token.

This will result in faster (~700ms reduction): registrations, logins and token rotations (refresh).


packages/passport/sdk/src/authManager.test.ts
```diff
@@ -170,13 +169,13 @@ describe('AuthManager', () => {
...
-              ...imxProfileData,
-            },
+        mockSigninPopup.mockResolvedValue(mockOidcUser);
+        // eslint-disable-next-line global-require
+        (require('jwt-decode') as jest.Mock).mockReturnValue({
+          passport: {
+            imx_eth_address: mockUserImx.imx.ethAddress,
+            imx_stark_address: mockUserImx.imx.starkAddress,
+            imx_user_admin_address: mockUserImx.imx.userAdminAddress,
           },
```
haydenfowler, line 179: To avoid having the `require('jwt-decode')` & disabling the linter on the next line, can we instead follow the same pattern that's being used for the `UserManager`? E.g.:

```
import jwt_decode from 'jwt-decode';
...
jest.mock('jwt-decode');
...
(jwt_decode as jest.Mock).mockImplementation(() => ({
  passport: {
    imx_eth_address: mockUserImx.imx.ethAddress,
    imx_stark_address: mockUserImx.imx.starkAddress,
    imx_user_admin_address: mockUserImx.imx.userAdminAddress,
  },
}));
```

haydenfowler (dismissed):

haydenfowler (approved):

yundifu (approved):

PR #1501, opened 2024-02-20
fix: WT-2132 Use default token images for tokens without icons

imx-mikhala:
# Summary

Adds the default token image for tokens that do not have a token image on cdn so the FramedImage component falls back to the default token image instead of biome default image

### Demo - using the new icon for default tokens

https://github.com/immutable/ts-immutable-sdk/assets/122326421/1f56b854-35f9-49f5-bc3f-535d738e933a

### Bg added to Option.FramedImage

Screenshot 2024-02-21 at 1 30 02 pm
Screenshot 2024-02-21 at 1 29 58 pm


dreamoftrees (dismissed):
👍 Nice work @imx-mikhala

dreamoftrees (approved):
👍 Looking great!

deepti-imx (approved):
Nice work! 🚀

PR #1507, opened 2024-02-20
feat: internal script for issuing bulk USDC on Sepolia

proletesseract:
# Summary

The USDC contract we deployed on Sepolia can only issue 10k USDC at a time. Here we've created a script/tool in the bridge sample app that will run the issuance multiple times as configured in the env file.

You will require to have the private key which is the owner of the USDC contract to run this script. The number you enter as `AMOUNT_10000` in the env file is the number of issuances of 10,000 tokens you want to do. 

So `AMOUNT_10000=100` will loop the script 100 times until 1,000,000 tokens are issued.

Tokens are issued to the owner address and then need to be transferred to whoever needs them.

Remember to keep this address topped up with Sepolia ETH if you use it a lot.


packages/internal/bridge/bridge-sample-app/.example.env
```diff
@@ -59,4 +59,9 @@ FLOW_RATE_TOKEN_DECIMALS=18
 # Values are in whole tokens and will be converted to wei using the decimals above
 FLOW_RATE_CAPACITY=15516
 FLOW_RATE_REFILL_RATE=4.31
-FLOW_RATE_LARGE_TRANSFER_THRESHOLD=7758
\ No newline at end of file
+FLOW_RATE_LARGE_TRANSFER_THRESHOLD=7758
```
ermyas, line 62: The large transfer threshold currently configured for USDC on Testnet is `10008`. Do you want to use that instead?

ermyas (approved):

packages/internal/bridge/bridge-sample-app/.example.env
proletesseract: can if we want but these are just examples and the settings should be figured out per token you're wanting to map though.

wcgcyx (approved):

PR #1518, opened 2024-02-22
chore: Regenerate IMX API clients

CodeSchwert:

# Summary

Regenerate the internal use IMX API clients.

This change should not impact external users as it updates an internal module.

This should update the IMX API clients to match the generated API clients in Core SDK now!


packages/internal/generated-clients/src/imx/api.ts
```diff
@@ -2,9 +2,9 @@
 /* eslint-disable */
 /**
  * Immutable X API
- * No description provided (generated by Openapi Generator https://github.com/openapitools/openapi-generator)
+ * Immutable X API
  *
- * The version of the OpenAPI document: 3.0
+ * The version of the OpenAPI document: 3.0.0
```
wdinau, line 7: what consideration made this happen?

packages/internal/generated-clients/src/imx/base.ts
```diff
@@ -18,7 +18,7 @@ import { Configuration } from "./configuration";
 // @ts-ignore
 import globalAxios, { AxiosPromise, AxiosInstance, AxiosRequestConfig } from 'axios';
 
-export const BASE_PATH = "https://api.sandbox.x.immutable.com".replace(/\/+$/, "");
+export const BASE_PATH = "https://api.sandbox.immutable.com".replace(/\/+$/, "");
```
wdinau, line 21: why no `x` anymore? 
all our docs has the `x` in the example queries.

packages/internal/generated-clients/src/imx-openapi.json
```diff
@@ -4,14 +4,16 @@
   ],
   "swagger": "2.0",
   "info": {
+    "description": "Immutable X API",
     "title": "Immutable X API",
```
carleeto, line 8: Looks like the same title and description resulted in the duplicate `Immutable X API` comments in the generated files. Might be something worth fixing in the source when we get a chance.

carleeto (approved):

packages/internal/generated-clients/src/imx/base.ts
CodeSchwert: I tried hunting down the source of this change, it's a bit of a mystery why the `x` is removed since the `host` in the spec file contains the `x`. API requests to the endpoints without the `x` do work though, I've tested several endpoints with and without the `x` and they both perform the same.

timm088: This _should_ work, but i cannot guarantee it @CodeSchwert @wdinau . We put both hosts in for the eventual migration to this hostname for "multil rollup".
In saying this, it is 'cleaner' for now (maybe?) for customers to understand .x relates to different API functionality?

CodeSchwert: Opted to go with the "cleaner" solution. Reverted the URL to include the `x`.

We should probably leave the URL change to a seperate epic / PR.

packages/internal/generated-clients/src/imx/api.ts
CodeSchwert: The original looks like an error on our part? Versions should have been using semver format from the start? (We do that for all our SDKs)

wdinau (approved):

wdinau (approved):

carleeto (approved):

PR #1519, opened 2024-02-22
refactor: WT-1875 Move fetching swap availability logic within render

imx-mikhala:
# Summary

Moves the call to isSwapAvailable inside the render to show a loader instead of nothing while waiting for the call to resolve.


sharonsheah (approved):
🚀

deepti-imx (approved):

PR #1520, opened 2024-02-22
feat: [WT-2170] Added EIP-6963 support to Passport SDK

dreamoftrees:
# Summary
Passport SDK zkEVM provider supports EIP-6963 (Multi Injection Provider Discovery)
https://immutable.atlassian.net/browse/WT-2170

## Added 
Passport connectEvm() will now announce the zkEVM provider via EIP-6963 events.


packages/passport/sdk/src/config/config.ts
```diff
@@ -62,6 +65,7 @@ export class PassportConfiguration {
     this.oidcConfiguration = oidcConfiguration;
     this.baseConfig = baseConfig;
     this.crossSdkBridgeEnabled = crossSdkBridgeEnabled || false;
+    this.announceProvider = announceProvider ?? true;
```
haydenfowler, line 68: Hey Charlie, is there any situation in which a consumer _wouldn't_ wan't Passport to announce itself? What's the use case here?

packages/passport/sdk/src/Passport.ts
```diff
@@ -138,14 +139,29 @@ export class Passport {
...
       authManager: this.authManager,
       magicAdapter: this.magicAdapter,
       config: this.config,
       multiRollupApiClients: this.multiRollupApiClients,
       guardianClient: this.guardianClient,
     });
+
+    // EIP-6963 Multi Injected Provider Discovery
+    if (this.config.announceProvider) {
+      announceProvider({
```
haydenfowler, line 153: The `Reference Implementation` in the [EIP-6963 docs](https://eips.ethereum.org/EIPS/eip-6963#:~:text=window.ethereum%20namespace.-,Reference%20Implementation,-Wallet%20Provider) suggest that the provider should be announced on page load, but this implementation only announces the provider after `connectEvm` has been called. 

This seems to go against the intended use case of the spec, and if we advertise Passport as EIP-6963 compliant, then I worry that consumers may not know that `connectEvm` has to be called for the provider to be announced

packages/passport/sdk/package.json
```diff
@@ -22,6 +22,7 @@
     "ethers": "^5.7.2",
     "jwt-decode": "^3.1.2",
     "magic-sdk": "^21.2.0",
+    "mipd": "^0.0.5",
```
haydenfowler, line 25: This will also need to be included as a dependency in `packages/passport/sdk/package.json`

haydenfowler, line 25: This will also need to be included as a dependency in `/sdk/package.json`

packages/passport/sdk/src/config/config.ts
dreamoftrees: Hey Hayden, it's a good question. The only use case I thought here is that some DApps may be listening for announced providers and showing them in a list. I imagine they would allowlist wallets they support but this option gives a little more granular control about how you are using the SDK and how you might want to connect to it.

packages/passport/sdk/src/Passport.ts
dreamoftrees: `announceProvider` includes a reference to the `provider` itself so we can't announce before instantiating the provider itself. Some customers may not want zkEVM but instead want an IMXProvider so I think we don't want to announce and create an instance unnecessarily.
  
My reading of EIP-6963 was that it solves the problem of race conditions so the providers can be announced at any stage as they respond to a requestProviders call. I agree we ideally want to fire this asap, but didn't want to impact the flow of calling `connectEVM`.

dreamoftrees: I agree with the potential disconnect on when the `announceProvider` fires based on the config property.. something we could document well? Not sure of a better way here work with the provider instance output from `connectEVM`?

packages/passport/sdk/src/Passport.ts
```diff
@@ -138,14 +139,29 @@ export class Passport {
...
+        info: {
```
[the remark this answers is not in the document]
dreamoftrees: Sure can @Jon-Alonso , where would you recommend? `src / zkEvm / walletHelpers.ts` or create something seperate `src / zkEvm / eip6963.ts`?

packages/passport/sdk/src/Passport.ts
dreamoftrees: @haydenfowler Something that is relatively unique here (maybe not, but haven't seen other solutions like this in the wild) is that our solution is SDK based. There is just more room for variation in how the SDK is bundled/loaded, instantiated and the provider created. Majority of injected providers will be coming from browser extensions that can hook into that page load event.. I feel like we could get away with making sure it's documented, or another option could be putting the `announceProvider` flag inside the `connectEvm` function. Making the flag more contextual to the creation of the provider?

PR #1524, opened 2024-02-23
fix: [GPR-443][Sale Widget] Remove block confirmations wait while signing sale order

jhesgodi:
# Summary
Overall a non breaking change that aims to improve order confirmations speed. 
[](url)
Before orders were confirmed in the UI after `approve` and `execute` where included in 1 block in the blockchain.
With this change, an order will be successful after `approve` and `execute` where submitted to the block chain.


dreamoftrees (dismissed):
Nice 👍  Just wanted to share.. one way we handled this for the other Checkout SDK endpoints was to build in some overrides like `waitFulfillmentSettlements` that basically either returns the results of send or does the wait() and awaits for txn settlement. We weren't sure which UX customers would need.. ```
const result = await checkout.cancel({
        provider,
        orderIds: [orderId],
        overrides: {
          useGaslessCancel,
          waitFulfillmentSettlements,
        }
      });

jhesgodi:
> Nice 👍 Just wanted to share.. one way we handled this for the other Checkout SDK endpoints was to build in some overrides like `waitFulfillmentSettlements` that basically either returns the results of send or does the wait() and awaits for txn settlement. We weren't sure which UX customers would need.. ``` const result = await checkout.cancel({ provider, orderIds: [orderId], overrides: { useGaslessCancel, waitFulfillmentSettlements, } });

thank you Charlie, we had just reviewed this one again, and decided it was not necessary to `wait(1)` but at least do `wait()` to be sure our approval transaction was mined before we send the execute

alejoloaiza (approved):

dreamoftrees (approved):
👍

PR #1528, opened 2024-02-26
fix: WalletList to select correct providerName

dreamoftrees:
# Summary
Fix to WalletList selected provider name. Issue was provider name being set to MetaMask when it could have also been a different selected provider.
https://immutable.atlassian.net/browse/WT-2180

## Fixed
WalletList to correctly select provider name.


packages/checkout/widgets-lib/src/widgets/connect/components/WalletList.tsx
```diff
@@ -62,7 +62,7 @@ export function WalletList(props: WalletListProps) {
   const connectCallback = async (ethereumProvider) => {
     if (ethereumProvider.connected && ethereumProvider.session) {
       const web3Provider = new Web3Provider(ethereumProvider as any);
-      selectWeb3Provider(web3Provider);
+      selectWeb3Provider(web3Provider, 'walletconnect');
```
keithbro-imx, line 65: Would be better to put the type assertion here, as this is the exception

dreamoftrees: Hey @keithbro-imx this is a different pathway that should execute as a result of selecting WalletConnect. We haven't turned this on yet so the users shouldn't see WalletConnect as an option.
I think elsewhere as you pointed out we had overwritten the providerName to force it as MetaMask, I've adjusted that so it should correctly use the selected providerName.

dreamoftrees: I'm about to refactor this component to support EIP-6963 and plan to adjust this implementation how we are using the providerName here.

sharonsheah (approved):

ZacharyCouchman (approved):

PR #1529, opened 2024-02-26
chore: Use the same version of node for CI and local dev

carleeto:
# Summary

Use the same version of node for CI and local dev

No code changes - only Github CI workflows


.editorconfig
wdinau, line 1: does this file only work for VS or it's something generic that intellij ppl (other IDE) also can benefit from? how does this conflict (or not) with linting?

carleeto: Its a standard that is supported by [most modern editors](https://editorconfig.org/#download)

wdinau (approved):

CodeSchwert (approved):

PR #1531, opened 2024-02-26
feat: WT-2158 handle wallet disconnect

ZacharyCouchman:
# Summary

Adding functionality into the Disconnect Wallet button in the Wallet Widget. If the current provider is a WalletConnect provider, we make sure to disconnect all existing pairings and disconnect the provider. This enables reconnection in the connect widget with a different mobile wallet.

Wallet Widget Configurations

When creating the Wallet Widget through the factory, you can now pass in flags to show/hide

* Disconnect Button
* Network Menu

Both of these configurations default to true.

### DIFF
--- a/packages/checkout/widgets-lib/src/widgets/wallet/views/Settings.tsx
+++ b/packages/checkout/widgets-lib/src/widgets/wallet/views/Settings.tsx
@@ -13,7 +13,13 @@ import { ConnectLoaderContext } from '../../../context/connect-loader-context/Co
 import { EventTargetContext } from '../../../context/event-target-context/EventTargetContext';
 import { UserJourney, useAnalytics } from '../../../context/analytics-provider/SegmentAnalyticsProvider';
 
-export function Settings() {
+export interface SettingsProps {
+  showDisconnectButton: boolean;
+}
+
+export function Settings({
+  showDisconnectButton,
+}: SettingsProps) {
   const { t } = useTranslation();
   const { connectLoaderState } = useContext(ConnectLoaderContext);
   const { checkout, provider } = connectLoaderState;
@@ -78,6 +84,7 @@ export function Settings() {
         sx={settingsBoxStyle}
       >
         <WalletAddress provider={provider} />
+        {showDisconnectButton && (
         <Button
           testId="disconnect-button"
           variant="secondary"
@@ -89,6 +96,7 @@ export function Settings() {
         >
           {t('views.SETTINGS.disconnectButton.label')}
         </Button>
+        )}
       </Box>
     </SimpleLayout>
   );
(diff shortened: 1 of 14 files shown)


packages/checkout/widgets-lib/src/widgets/wallet/views/Settings.tsx
```diff
@@ -26,6 +30,36 @@ export function Settings() {
...
+      console.log('disconnecting passport provider by calling logout on Passport');
+      await checkout?.passport?.logout();
+      console.log('disconnected');
+      return;
+    }
+
+    if (isMetaMaskProvider(provider)) {
+      console.log('disconnecting MetaMask provider by revoking permissions');
+      // eslint-disable-next-line @typescript-eslint/naming-convention
+      await (provider?.provider as any).request({ method: 'wallet_revokePermissions', params: [{ eth_accounts: {} }] });
```
dreamoftrees, line 44: I feel like this might be too much. Revoking permissions would include clearing token approvals and allowances, that might have a direct impact on the hosting website if they are using things beyond our widgets.
Something we haven't quite included is the `activeProvider` or `selectedProvider` something we can persist across sessions that I think makes sense to clear here. So auto-wallet selection even after a refresh is cleared.

dreamoftrees: We could also potentially include an information link about how users can disconnect inside their wallet: https://support.metamask.io/hc/en-us/articles/360059535551-Disconnect-wallet-from-a-dapp

dreamoftrees: Another option is including a `Revoke Permissions` button under the disconnect with a warning and more information about what it means?

dreamoftrees (commented):
Nice work @ZacharyCouchman had some comments about how I think revokePermissions might be too much in this instance.

packages/checkout/widgets-lib/src/widgets/wallet/views/Settings.tsx
ZacharyCouchman: Yeah that's a good point, I didn't think it cleared token approvals and allowances becuase they should be stored on the respective smart contracts. 

I think what it mainly does is revokes the connecting in MetaMask to the dapp so that they would need to call eth_requestAccounts / wallet_requestPermissions again and connect in the MM pop up.

What we could do in the meantime is just listen for the disconnect event in the WidgetRoot and clear the web3provider meaning that the widgets just go back to the Connect Wallet state.

What do you reckon?

ZacharyCouchman: @dreamoftrees I actually think we should do any passport logout or wallet revoke permissions for now. The partner can listen for this event and do a logout / revoke permission if they wish.

The only provider we do need to do this for is WalletConnect. We need to disconnect the pairings so that upon reconnect with WalletConnect the user sees the QR Modal again and can choose a different wallet if need be.

sharonsheah (approved):

dreamoftrees (approved):
👍

PR #1538, opened 2024-02-27
feat: WT-2156 Bridge Widget Wallet Images for WC

sharonsheah:

# Summary

**JIRA ticket:** https://immutable.atlassian.net/browse/WT-2156

**Demo + screenshots:** https://imtbl.slack.com/archives/C0414KQ5162/p1709005294289049

## Added 

- pulled wallet ids from WC explorer into constants
- in WalletConnectManager, fetch wallet image url from WC Explorer API based on provider name
- extend useWalletConnect hook to getWalletLogoUrl from WalletConnectManager to be accessed in relevant components
- updates WC modal background overlay colour (https://imtbl.slack.com/archives/C0414KQ5162/p1708489972283279)
- implement small WC logo in top right corner in `WalletNetworkButton`, `BridgeReviewSummary`, `Transactions`


packages/checkout/widgets-lib/src/lib/walletConnect.ts
```diff
@@ -149,4 +159,31 @@ export class WalletConnectManager {
...
+    try {
+      const response = await fetch(walletListingsApi);
+      const data = await response.json();
+      return data;
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error('Error fetching wallet listings', error);
+    }
+    return undefined;
+  }
```
ZacharyCouchman, line 175: This is great 👏 . It might be worth saving/caching this response so we don't have to call it everytime we want call getWalletLogoUrl.

packages/checkout/widgets-lib/src/widgets/bridge/components/BridgeReviewSummary.tsx
```diff
@@ -269,7 +294,17 @@ export function BridgeReviewSummary() {
         emphasized
         sx={bottomMenuItemStyles}
       >
-        {fromWalletProviderName && (
+        {fromWalletProviderName && fromWalletIsWalletConnect && fromWalletLogoUrl && (
```
ZacharyCouchman, line 297: Do we need `fromWalletProviderName` here anymore?

packages/checkout/widgets-lib/src/components/Transactions/ChangeWallet.tsx
```diff
@@ -39,18 +49,38 @@ export function ChangeWallet({ onChangeWalletClick }: ChangeWalletProps) {
     onChangeWalletClick();
   };
 
+  useEffect(() => {
+    if (isWalletConnectEnabled) {
+      setIsWalletConnect((from?.web3Provider.provider as any)?.isWalletConnect);
```
ZacharyCouchman, line 54: Just FYI we do have util functions for checking these, isMetaMaskProvider(provider), isPassportProvider(provider) and isWalletConnectProvider(provider)

ZacharyCouchman:
Looks great @sharonsheah , nice work. I'd just check if we are making the call to get listings everytime we need to get a wallet logo url. I'd say that the wallet listing metadata wouldn't change too often and it is probably safe for us to cache it

packages/checkout/widgets-lib/src/components/Transactions/ChangeWallet.tsx
dreamoftrees: Oh nice, I didn't realise! Might make sense to add an isWalletConnectProvider(provider) util as well?

packages/checkout/widgets-lib/src/lib/walletConnect.ts
```diff
@@ -149,4 +159,31 @@ export class WalletConnectManager {
...
+
+  public async getWalletLogoUrl(): Promise<string | undefined> {
+    const walletListings = await this.loadWalletListings() as any;
+    const walletName = this.ethereumProvider?.session?.peer.metadata.name;
+    if (!walletListings || !walletName) {
+      return undefined;
```
dreamoftrees, line 181: Would do you think about returning a default / fallback URL here? something that we can display if we can't resolve the wallet icon.

packages/checkout/widgets-lib/src/widgets/bridge/components/BridgeReviewSummary.tsx
```diff
@@ -306,7 +341,17 @@ export function BridgeReviewSummary() {
         emphasized
         sx={topMenuItemStyles}
       >
-        {toWalletProviderName && (
+        {toWalletProviderName && toWalletIsWalletConnect && toWalletLogoUrl && (
+          <>
+            <MenuItem.FramedImage
+              imageUrl={toWalletLogoUrl}
+              alt="wallet connect"
```
dreamoftrees, line 348: small nit: `walletconnect` one word.

dreamoftrees (commented):
Great work Sharon, just a couple of comments. Thanks.

packages/checkout/widgets-lib/src/lib/walletConnect.ts
ZacharyCouchman: Good call. I think we should just fallback to the WalletConnect logo in this case

sharonsheah: @dreamoftrees It currently does fallback to the `WalletConnectSymbol` in the `<Logo>` component if the `getWalletLogoUrl` returns undefined.

dreamoftrees: @sharonsheah perfect sorry missed that!

dreamoftrees (approved):
👍 Noice work!

ZacharyCouchman (approved):

PR #1540, opened 2024-02-28
[ID-1455] Create wagmi connector for Passport

alex-connolly:
Creates a https://wagmi.sh/ connector for Passport in a new package '@imtbl/passport-wagmi-connector'.

### DIFF
--- a/packages/passport/wagmi-connector/src/index.ts
+++ b/packages/passport/wagmi-connector/src/index.ts
@@ -1,0 +1,71 @@
+import { passport } from '@imtbl/sdk';
+import { SwitchChainError, getAddress } from 'viem';
+import {
+  ChainNotConfiguredError, createConnector, normalizeChainId, CreateConnectorFn,
+} from '@wagmi/core';
+
+export function PassportConnector(pp: passport.Passport): CreateConnectorFn {
+  type Properties = {};
+
+  return createConnector<passport.Provider, Properties>((config) => ({
+    id: 'immutable-passport',
+    name: 'Immutable Passport',
+    type: 'Immutable Passport',
+
+    async setup() {
+      const provider = await this.getProvider();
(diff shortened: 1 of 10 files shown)


packages/passport/wagmi-connector/src/index.ts
```diff
@@ -0,0 +1,71 @@
...
+    async getAccounts() {
+      const provider = await this.getProvider();
+      const accounts = await provider.request({ method: 'eth_requestAccounts' });
+      return (accounts as Array<string>).map((x) => getAddress(x));
+    },
+    async getProvider() {
+      return pp.connectEvm();
+    },
+    async isAuthorized() {
+      // TODO: does this work? Won't this just pop up the login window (may not be intended)
```
haydenfowler, line 39: Correct, this will trigger the login flow if the user is not authenticated. It looks like we should be calling [eth_accounts](https://docs.immutable.com/docs/zkevm/products/passport/wallet/rpc-methods/eth_accounts/) which will return an array containing the users address if the user is auth'd, or an empty array if they are not

packages/passport/wagmi-connector/src/index.ts
```diff
@@ -0,0 +1,71 @@
...
+      provider.on('disconnect', this.onDisconnect);
+    },
+    async connect() {
+      return {
+        accounts: await this.getAccounts(),
+        chainId: await this.getChainId(),
+      };
+    },
+    async disconnect() {
+      // TODO: is this implemented
```
haydenfowler, line 28: The `logout` function can be called on the root Passport module: https://github.com/immutable/ts-immutable-sdk/blob/ed9a7450c71f82e0f7e899201e5bf96749b501e2/packages/passport/sdk/src/Passport.ts#L225

PR #1542, opened 2024-02-29
feat: [WT-2170] Injected providers support via EIP-6963

dreamoftrees:
# Summary
Checkout Widgets support for injected providers via EIP-6963.
https://immutable.atlassian.net/browse/WT-2170

Screenshot 2024-03-01 at 6 43 35 am

## Added 
Support for allowlisted EIP-6963 injected wallet providers.

## Fixed
Detection of MetaMask browser wallet via EIP-6963, removing any conflicts from existing/additional wallets.


packages/checkout/widgets-lib/src/lib/walletConnect.ts
```diff
@@ -2,6 +2,7 @@ import { WalletConnectModal } from '@walletconnect/modal';
 import EthereumProvider from '@walletconnect/ethereum-provider';
 import { ChainId, WidgetTheme } from '@imtbl/checkout-sdk';
 import { Environment } from '@imtbl/config';
+import { ConnectConfig } from '@imtbl/checkout-sdk/dist/types';
```
ZacharyCouchman, line 5: This hasn't been exported properly from the sdk. Either we export the type from there or create an equivalent type in widgets lib

dreamoftrees: Oh good catch! I've just used a local type for now, the only property I'm interested on that is `walletconnect` anyway so should be pretty safe if they go out of sync.

packages/checkout/widgets-lib/src/widgets/bridge/components/WalletAndNetworkSelector.tsx
```diff
@@ -90,28 +86,15 @@ export function WalletAndNetworkSelector() {
...
-    if (isWalletConnectEnabled) {
-      options.push('walletconnect');
-    }
-    return options;
-  }, [checkout, fromNetwork, fromWalletProviderName, isWalletConnectEnabled]);
+  const toWalletSelectorOptions = useMemo(() => (
+    providers
+      .filter((providerDetail) => (
+        providerDetail.info.rdns !== WalletProviderRdns.PASSPORT
+        || (providerDetail.info.rdns === WalletProviderRdns.PASSPORT && fromNetwork === l1NetworkChainId)
```
ZacharyCouchman, line 95: Are we assuming here that we're not using passport injected into the checkout instance anymore?

packages/checkout/widgets-lib/src/lib/hooks/useInjectedProviders.ts
```diff
@@ -0,0 +1,87 @@
...
+    filteredProviders.unshift(getPassportProviderDetail(passportWeb3Provider.provider as EIP1193Provider));
+  }
+
+  return filteredProviders;
+};
+
+/**
+ * Hook that supplies a filters and sorted list of EIP-6963 injected providers.
+ */
+export const useInjectedProviders = ({ checkout }: UseInjectedProvidersParams) => {
```
ZacharyCouchman, line 59: So just making sure I understand what we're doing here, every time we use this hook we get an instance of the InjectedProvidersManager, that fires the event to call for all of the providers to reveal themselves. Then we add the Passport provider into that list if it hasn't been revealed by the 6963 event. Then we store the state for the passport provider at the top of this file and only create it once.

packages/checkout/widgets-lib/src/lib/hooks/useWalletConnect.ts
```diff
@@ -90,7 +83,7 @@ export const useWalletConnect = ({ checkout }: UseWalletConnectParams) => {
         const pairingTopicFromUrl = data.split('@')[0].replace('wc:', '');
         // eslint-disable-next-line no-console
         console.log('useWalletConnect::pairingTopic', pairingTopicFromUrl);
-
+        console.log('walletConnectModal', walletConnectModal, data);
```
ZacharyCouchman, line 86: we should probably clean up a lot of these console logs now

packages/checkout/widgets-lib/src/lib/providerUtils.ts
```diff
@@ -13,30 +14,32 @@ export function isWalletConnectProvider(provider?: Web3Provider | null) {
...
+  if (isPassportProvider(web3Provider)) return WalletProviderName.PASSPORT.toString();
+  if (isWalletConnectProvider(web3Provider)) return 'walletconnect';
 
-  return WalletProviderName.METAMASK;
+  if (providers && web3Provider) {
+    // Find the matching provider in the providerDetail
+    const matchedProviderDetail = providers.find((providerDetail) => providerDetail.provider === web3Provider.provider);
+    if (matchedProviderDetail) {
+      return matchedProviderDetail.info.name;
+    }
```
ZacharyCouchman, line 30: 👏

packages/checkout/widgets-lib/src/widgets/bridge/components/WalletAndNetworkSelector.tsx
dreamoftrees: I've integrated `passport` into the useInjectedProviders hook, so it will use the injected passport provider via EIP6963 .. if it can't resolve that then it attempts a fallback to the injected passport instance provider. And then just simulates the EIP6963 data so it should be future proof.

packages/checkout/widgets-lib/src/lib/hooks/useInjectedProviders.ts
dreamoftrees: Yep exactly.. the idea is that we support a fallback mechanism and then I'll work with passport SDK to add support for EIP6963 and the instance will be managed elsewhere then.

packages/checkout/widgets-lib/src/lib/walletConnect.ts
```diff
@@ -46,18 +50,20 @@ const lightThemeVariables = {
...
 const frontierId = '85db431492aa2e8672e93f4ea7acf10c88b97b867b0d373107af63dc4880f041';
 const ledgerLiveId = '19177a98252e07ddfc9af2083ba8e07ef627cb6103467ffebb3f8f4205fd7927';
-// const coinbaseId = 'fd20dc426fb37566d803205b19bbc1d4096b248ac04548e3cfb6b3a38bd033aa';
+const coinbaseId = 'fd20dc426fb37566d803205b19bbc1d4096b248ac04548e3cfb6b3a38bd033aa';
 // const phantomId = 'a797aa35c0fadbfc1a53e7f675162ed5226968b44a19ee3d24385c64d1d3c393';
 // const rainbowId = '1ae92b26df02f0abca6304df07debccd18262fdf5fe82daa81593582dac9a369';
 
 const productionWalletWhitelist = [metamaskId, frontierId, ledgerLiveId];
-const sandboxWalletWhitelist = [metamaskId];
+const sandboxWalletWhitelist = [metamaskId, frontierId, coinbaseId];
```
sharonsheah, line 58: Not sure what wallets we want in this array

ZacharyCouchman: frontier doesn't work in Sandbox as it doesn't have Sepolia so it shouldn't be there. I haven't tested with coinbase mobile wallet either. Can we check this?

sharonsheah (dismissed):

packages/checkout/widgets-lib/src/lib/walletConnect.ts
dreamoftrees: Good catch @sharonsheah .. I've removed these both from sandbox!

packages/checkout/widgets-lib/src/components/Transactions/Transactions.tsx
```diff
@@ -301,6 +237,49 @@ export function Transactions({
...
+        control: 'WalletProvider',
+        controlType: 'Select',
+        extras: {
+          walletProviderName: event.providerDetail.info.name,
+        },
+      });
+
+      try {
+        const web3Provider = new Web3Provider(event.provider as any);
+        const connectedProvider = await connectToProvider(checkout, web3Provider, true);
```
ZacharyCouchman, line 254: should this changeAccount always hardcode to true here? I think this should just be for MetaMask right?

ZacharyCouchman: Yeah, just tested, this is breaking when trying to use WalletConnect in the Transactions screen

packages/checkout/widgets-lib/src/components/Transactions/Transactions.tsx
```diff
@@ -144,66 +140,6 @@ export function Transactions({
     }
   }, [checkout, from]);
```
ZacharyCouchman, line 142: I'm not sure why all of this is being deleted. This is handling WalletConnect differently to Passport and Metamask

packages/checkout/widgets-lib/src/components/Transactions/Transactions.tsx
dreamoftrees: yep good catch fixed.

packages/checkout/widgets-lib/src/lib/provider/injected/register.ts
```diff
@@ -0,0 +1,25 @@
+import { EIP1193Provider } from '../types/eip1193';
+
+export interface Register {}
+
+export type DefaultRegister = {
+  /** The EIP-1193 Provider. */
+  provider: EIP1193Provider;
+  /** Reverse Domain Name Notation (rDNS) of the Wallet Provider. */
+  rdns: 'com.coinbase' | 'com.enkrypt' | 'io.metamask' | 'io.zerion';
```
ZacharyCouchman, line 9: out of interest, where do we get these from?

packages/checkout/widgets-lib/src/components/Transactions/Transactions.tsx
dreamoftrees: Talked it through

ZacharyCouchman (approved):

sharonsheah (approved):

jhesgodi (approved):

packages/checkout/widgets-lib/src/components/RawImage/RawImage.tsx
```diff
@@ -0,0 +1,35 @@
+import { Box, useConvertSxToEmotionStyles } from '@biom3/react';
+import { imageStyles, rawImageStyles } from './RawImageStyles';
+
+export interface RawImageProps {
```
jhesgodi, line 4: not a blocker, to keep the type proper, you could import `BoxProps` and  do `sx: BoxProps['sx']`

packages/checkout/widgets-lib/src/factory.ts
```diff
@@ -15,6 +15,7 @@ import { Sale } from 'widgets/sale/SaleWidgetRoot';
 import { Web3Provider } from '@ethersproject/providers';
 import { Bridge } from 'widgets/bridge/BridgeWidgetRoot';
 import { WalletConnectManager } from 'lib/walletConnect';
+import { ConnectConfig } from '@imtbl/checkout-sdk/dist/types';
```
jhesgodi, line 18: is it safe to get t his type from `dist`?

packages/checkout/widgets-lib/src/lib/hooks/useAnimation.ts
```diff
@@ -0,0 +1,24 @@
+export const useAnimation = () => {
```
jhesgodi, line 1: I think the you don't require a hook for keeping this static values, it will more efficient to directly import them. Thoughts?

packages/checkout/widgets-lib/src/lib/hooks/useInjectedProviders.ts
```diff
@@ -0,0 +1,90 @@
...
+} from '../provider';
+
+export interface UseInjectedProvidersParams {
+  checkout: Checkout | null;
+}
+
+let passportWeb3Provider: Web3Provider;
+const processProviders = (checkout: Checkout | null, injectedProviders: EIP6963ProviderDetail<EIP1193Provider>[]) => {
+  // Apply wallet providers allowlist
+  const allowedWalletRdns = [
```
jhesgodi, line 18: Can this Rdns be set from config? give that some are disable now, I'll think this list changes overtime?

carleeto (approved):

packages/checkout/widgets-lib/src/lib/hooks/useAnimation.ts
dreamoftrees: Yep totally agree! will move it out

packages/checkout/widgets-lib/src/lib/hooks/useInjectedProviders.ts
dreamoftrees: Yep good call, that was the plan. Just not in this PR.

packages/checkout/widgets-lib/src/lib/provider/injected/register.ts
dreamoftrees: I just saw them on the EIP-6963 window post message, but they are also similar to the domain listed on walletconnect explorer https://explorer.walletconnect.com/coinbase-wallet

packages/checkout/widgets-lib/src/components/RawImage/RawImage.tsx
dreamoftrees: Nice input, patched this.

packages/checkout/widgets-lib/src/factory.ts
dreamoftrees: Fixed.

packages/checkout/widgets-lib/src/lib/hooks/useAnimation.ts
dreamoftrees: Fixed.

PR #1544, opened 2024-02-29
feat: WT-2157 Handle add new network and switch network functionality

sharonsheah:

# Summary

https://immutable.atlassian.net/browse/WT-2157

## Added 

- add separate `addNetwork` function to public sdk
- execute addNetwork if provider is WC + metamask
- if zkevm network doesn't exist in MM app, it prompts to add new network and switch network
- if zkevm network exists in MM app and user is on other network, it prompts to switch network
- if user currently on zkevm, it should complete connection flow to success screen
- display generic copy for WC provider on switch network screen
- adds production and sandbox whitelisted wallets; some refactoring around wallet ids
- removes coinbase from whitelisted wallets


packages/checkout/widgets-lib/src/widgets/connect/views/SwitchNetworkZkEVM.tsx
```diff
@@ -38,7 +39,63 @@ export function SwitchNetworkZkEVM() {
...
+          view: {
+            type: ConnectWidgetViews.SUCCESS,
+          },
+        },
+      });
+
+      return;
+    }
+
+    const walletName = (provider.provider as any)?.session?.peer?.metadata?.name.toLowerCase();
```
dreamoftrees, line 68: Does it make sense to execute this if isWalletConnectProvider(provider) is false?

packages/checkout/widgets-lib/src/widgets/connect/views/SwitchNetworkZkEVM.tsx
```diff
@@ -38,7 +39,63 @@ export function SwitchNetworkZkEVM() {
...
+
     try {
+      if (isWalletConnectProvider(provider) && walletName === 'metamask') {
+        try {
```
dreamoftrees, line 72: The 2 layers of try catch feels a little redundant.. might be worth flattening them out to different code blocks?

dreamoftrees (commented):
Nice work @sharonsheah just a couple of comments. thanks!

packages/checkout/sdk/src/types/network.ts
```diff
@@ -3,6 +3,16 @@ import { ChainId } from './chains';
...
 
+/**
+ * Interface representing the parameters for {@link Checkout.addNetwork}.
+ * @property {Web3Provider} provider - The provider to connect to the network.
+ * @property {ChainId} chainId - The ID of the network to add.
+ */
+export interface AddNetworkParams {
+  provider: Web3Provider;
+  chainId: ChainId;
+}
```
ZacharyCouchman, line 14: When we add docs for this method it would be good to note that we only support adding Immutable zkEVM and/or Immutable zkEVM Testnet for now

packages/checkout/widgets-lib/src/widgets/connect/views/SwitchNetworkZkEVM.tsx
ZacharyCouchman: Just want to make sure that the flow is still correct for WalletConnect if flattening this out. For example we're catching the first error and skipping then going to switch

sharonsheah: Yeah, not sure if it'll work as we want if we flatten this out. Happy to pair to see how we could flatten if there's a way.

dreamoftrees: All good!

dreamoftrees (approved):
Noice.

dreamoftrees (approved):
👍 sweet

ZacharyCouchman (approved):

PR #1554, opened 2024-03-01
fix: [GPR-454][Sale Widget] Update fractional balance checks

jhesgodi:
# Summary
Fractional balance checks are done before invoking smart checkouts (this will possibly change later).

Fractional balance checks needed to skip gas requirements if the wallet is passport, given that gas is sponsored with this wallet.

## Added 
* A check to skip IMX balance if wallet is passport

## Changed
* Removes filtering swap funding routes from smart checkout results, because on-ramps are now fully integrated and swaps now have a liquidity pool.

## Deprecated
* Delete and replace obsolete tests

## Fixed
* Fixes bug where purchases initiated with a passport wallet with zero IMX are rejected, even if it has enough ERC20 to pay for the total amounts.


alejoloaiza (approved):

packages/checkout/widgets-lib/src/widgets/sale/functions/smartCheckoutUtils.ts
```diff
@@ -120,32 +135,60 @@ export const smartCheckoutTokensList = (
...
+
+  if (isPassport && !smartCheckoutResult.sufficient && onlyNativeBalanceIsInsufficient) {
+    return {
+      sufficient: true,
+      transactionRequirements: smartCheckoutResult.transactionRequirements,
+    } as SmartCheckoutSufficient;
+  }
+
+  // otherwise, filter out disabled steps if required
+  const stepTypesToFiler = [
```
jhesgodi, line 163: at the moment of this PR, it has not been decided to remove BRIDGE as a secondary flow for a primary sale.

We are deferring deprecating this filter until its UX is clarified. Also to keep in mind, smart checkouts will at some point include a way to skip gas check or disable funding routes from configurations. When that happens this filtering will not longer required.

dreamoftrees (approved):
👍

packages/checkout/widgets-lib/src/widgets/sale/functions/smartCheckoutUtils.ts
jhesgodi: cc @dreamoftrees

alejoloaiza (approved):

PR #1555, opened 2024-03-01
feat: SMR-2490 Tenderly Gas

proletesseract:
# Summary
Adds L1 Gas Estimation using Tenderly Simulations as per issue[ SMR-2490](https://immutable.atlassian.net/browse/SMR-2490).

Some SDK interfaces have been updated. Since the Bridge SDK is internal this only effects Checkout.

@immutable/wallets team will need to pair to update widget to pass the new params to the bridge SDK

Some more unit tests will be added on monday.


packages/internal/bridge/sdk/src/constants/bridges.ts
```diff
@@ -64,9 +64,18 @@ export const ETH_MAINNET_TO_ZKEVM_MAINNET: BridgeInstance = {
  * @typedef {Object} axelarAPIEndpoints - API endpoints for the testnet & mainnet Axelar environment configurations
  */
 export const axelarAPIEndpoints:Record<string, string> = {
-  testnet: 'https://testnet.api.gmp.axelarscan.io',
   mainnet: 'https://api.gmp.axelarscan.io',
-  devnet: '',
+  testnet: 'https://testnet.api.gmp.axelarscan.io',
+  devnet: 'https://testnet.api.gmp.axelarscan.io',
```
ermyas, line 69: I think the devnet endpoint is: https://stagenet.axelarscan.io/resources/chains

proletesseract: updated.

ermyas (approved):

PR #1557, opened 2024-03-03
feat: Connect Widget allow for targeting L1 chains

ZacharyCouchman:
Allowing for the Connect Widget to target Ethereum (L1) chains. This is to support using the Connect Widget and not wanting to auto switch to Immutable zkEVM.

By default if the targetChain param is not specified, it will fallback to targeting the zkEVM chain for that environment.

Refactored the ConnectLoader to use targetChainId instead of target layer.

# Summary
Allowing the Connect Widget to target L1 chains when connecting. Pass targetChainId when mounting the connect widget.


packages/checkout/widgets-lib/src/widgets/connect/views/ReadyToConnect.tsx
```diff
@@ -86,21 +86,27 @@ export function ReadyToConnect({ targetChainId, allowedChains }: ReadyToConnectP
...
+      if (isPassport) {
         viewDispatch({
           payload: {
             type: ViewActions.UPDATE_VIEW,
-            view: { type: ConnectWidgetViews.SWITCH_NETWORK },
+            view: { type: ConnectWidgetViews.SUCCESS },
           },
         });
         return;
       }
```
ZacharyCouchman, line 102: @dreamoftrees I was thinking we handle Passport by just removing it from the Wallet list when targetChainId is an L1 chain. This code will currently let it pass through and connect correctly. But we know Passport is only L2 for now.

packages/checkout/widgets-lib/src/widgets/connect/ConnectWidgetRoot.tsx
```diff
@@ -22,6 +25,33 @@ export class Connect extends Base<WidgetType.CONNECT> {
   }
 
   protected getValidatedParameters(params: ConnectWidgetParams): ConnectWidgetParams {
+    const validatedParams = params;
+
+    // validating targetChainId per environment
+    if (!params.targetChainId
+      && this.checkout.config.environment === Environment.PRODUCTION) {
```
sharonsheah, line 32: was thinking if we have a isProduction/isSandbox util function?

sharonsheah (dismissed):
LGTM! It looks cleaner with stating chain id rather than 'layer'. Nice work! 💯

sharonsheah (approved):

packages/checkout/widgets-lib/src/widgets/connect/ConnectWidgetRoot.tsx
```diff
@@ -37,6 +67,7 @@ export class Connect extends Base<WidgetType.CONNECT> {
               <ConnectWidget
                 config={this.strongConfig()}
                 checkout={this.checkout}
+                targetChainId={this.parameters.targetChainId}
```
dreamoftrees, line 70: Nice one.. just double checking does this use the validated Params?

dreamoftrees (approved):
Awesome work @ZacharyCouchman looks nice and clean!

packages/checkout/widgets-lib/src/widgets/connect/ConnectWidgetRoot.tsx
ZacharyCouchman: Yes it does. The validated params stuff get calls in the constructor of BaseWidgetRoot. So this.parameters should be validated by there

PR #1558, opened 2024-03-04
fix: IMXClient not sending API keys

CodeSchwert:

# Summary

The underlying API clients in the `IMXClient` class were not correctly setting the API keys from Hub in request headers.

`IMXClient` methods should now send the API keys.

## Fixed
- IMXClient methods were not sending Hub API keys with requests

### DIFF
--- a/packages/x-client/src/config/index.ts
+++ b/packages/x-client/src/config/index.ts
@@ -4,6 +4,7 @@ import {
   Environment,
   ImmutableConfiguration,
   ModuleConfiguration,
+  addKeysToHeadersOverride,
 } from '@imtbl/config';
 
 export { Environment, ImmutableConfiguration } from '@imtbl/config';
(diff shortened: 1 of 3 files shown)


carleeto (approved):

packages/x-client/src/config/index.ts
```diff
@@ -174,9 +185,9 @@ export const imxClientConfig = (configOptions: ConfigOptions): ImxModuleConfigur
     throw new Error(`Invalid environment: ${configOptions.environment}`);
   }
 
-  const clientConfig = new ImxConfiguration({
+  const clientConfig = {
     baseConfig: new ImmutableConfiguration(configOptions),
-  });
+  } as ImxModuleConfiguration;
```
wdinau, line 190: just curious, why is this "as" matters since you are returning clientConfig anyway?

wdinau (approved):

packages/x-client/src/config/index.ts
CodeSchwert: Betting typing for the user. `IMXClient` expects a `ImxModuleConfiguration`, the function returns an object with the required properties, but when inspected it just says it's an object with properties, instead of a `ImxModuleConfiguration`.

PR #1562, opened 2024-03-05
Chore: Migrate to seaport-js v3

Sam-Jeston:
# Summary

seaport-js V2 has bugs associated with partial fills. This necessitates a move to V3 to allow ERC1155 support, however that requires interoperability with ethers V6. This has been achieved in a non-breaking way and keeps the orderbook module interfaces Ethers V5 compatible.

## Added 

- Babel to the SDK build pipeline to transpile Ethers V6 and seaport-js V3 as these have ES7 syntax which are not natively compatible with rollup at this point.

## Changed

- Upgraded from seaport-js V2 to V3
- The orderbook overrides, which serve internal use-cases only, to take a JsonRpcUrl instread of a JsonRpcProvider directly

Babel was configured to purposefully configured to only transpile modules that relate to the orderbook to reduce the impact of the bundle as much as possible.


Sam-Jeston:
This is not ready to merge yet. I've found fulfillments don't work from the browser, for either Metamask or Passport, whereas functional tests are working

Sam-Jeston:
I have confirmed that this version of SDK for orderbook features works in checkout playground. Will test further next week

Sam-Jeston:
This branch has now been tested significantly in the checkout playground by both myself and @frankisawesome. We tested at 1.24.2-alpha.1 from commit [988574b](https://github.com/immutable/ts-immutable-sdk/pull/1562/commits/988574b249dd29740b731413795f71a0c6b66897)

lfportal (approved):

packages/orderbook/src/config/config.ts
```diff
@@ -17,7 +17,7 @@ export interface OrderbookModuleConfiguration {
   zoneContractAddress: string;
   apiEndpoint: string;
   chainName: string;
-  provider: providers.JsonRpcProvider | providers.Web3Provider;
```
carleeto, line 20: Same here - wouldn't it break backwards compatibility if someone was using a Web3Provider here?

carleeto (commented):
A couple of questions.

packages/orderbook/src/config/config.ts
```diff
@@ -5,7 +5,7 @@ export const TESTNET_CHAIN_NAME = 'imtbl-zkevm-testnet';
 export const MAINNET_CHAIN_NAME = 'imtbl-zkevm-mainnet';
 
 export interface OrderbookOverrides {
-  provider?: providers.JsonRpcProvider | providers.Web3Provider;
```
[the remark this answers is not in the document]
Sam-Jeston: Strictly speaking yes, but overrides were only ever added to ease our development experience, they were not made for public use (although strictly they can be used).  I have confirmed the following about our two marketplace partners:
 - Token Trove use checkouts. This change is not breaking for checkouts as they do not use any overrides when initialising the orderbook from their components
 - Merit Circle do not use our SDK

So the impact here would be limited to people using the orderbook SDK for some other use case, and considering the SDK never signs or sends transactions (it can only be used to build transactions), even in the case where a user was passing in a provider, it would in no way break their experience to pass in the rpc url instead.

I have also confirmed none of our internal utils libraries use the override, except one tool that traders own for testing.

packages/orderbook/src/config/config.ts
Sam-Jeston: This module config type is not used when initiating the module. Only the base config and overrides can be provided

carleeto (approved):

PR #1571, opened 2024-03-06
fix: [GPR-468][Sale Widget] Improve flow speed by removing unnecessary loaders

jhesgodi:
# Summary
Remove loader that waits for client config to be initialised.
Config endpoint avg. response time is less than 300ms, hence widget will optimistically let user move to select payment method, once user selects an option the config should be already available. if not, and an error will be thrown before signing.

## Added 
* Validate currency address was selected before proceeding to sign order

## Changed
* Removed currency validation on widget initialisation

## Fixed
* Remove array from translation files as its not supported


alejoloaiza (approved):

dreamoftrees (approved):
👍

packages/checkout/widgets-lib/src/widgets/sale/views/PayWithCoins.tsx
```diff
@@ -13,14 +13,25 @@ export function PayWithCoins() {
...
+    sendSuccessEvent,
   } = useSaleEvent();
-  const { execute, signResponse, executeResponse } = useSaleContext();
+  const {
+    execute, signResponse, executeResponse, signTokenIds,
+  } = useSaleContext();
   const executedTxns = executeResponse?.transactions.length || 0;
 
-  let loadingText = t('views.PAYMENT_METHODS.loading.ready');
+  let loadingText = [
```
dreamoftrees, line 23: Hey @jhesgodi just wanted to show this option for getting back an array of keys..
t('views.PAYMENT_METHODS.loading.ready', { returnObjects: true })

PR #1580, opened 2024-03-08
feat: Hook up error drawers to connect widget

sharonsheah:

# Summary

## **User rejected/ ChangedYourMindDrawer flow**

**_MetaMask_**

https://github.com/immutable/ts-immutable-sdk/assets/37326128/03e9d475-1eff-4aeb-b334-80d712bfcb1a

**_Passport_**

https://github.com/immutable/ts-immutable-sdk/assets/37326128/df9cf426-5d66-4897-a1aa-f33cc7aaa12e

## **Other errors / UnableToConnectDrawer flow**

https://github.com/immutable/ts-immutable-sdk/assets/37326128/f541c35b-2c90-4d66-a0d5-3a68d18ba298

## Changed

- showing ChangedYourMind drawer when user rejects connection, all other errors will show UnableToConnect drawer

## Removed

- replaced generic error view with unable to connect drawer

## Fixed

- add `isWalletConnectEnabled` as dependency to avoid multiple errors for WC initialisation

### DIFF
--- a/packages/checkout/widgets-lib/src/widgets/connect/components/WalletList.tsx
+++ b/packages/checkout/widgets-lib/src/widgets/connect/components/WalletList.tsx
@@ -1,6 +1,7 @@
 import { Box } from '@biom3/react';
 import {
   ChainId,
+  CheckoutErrorType,
   WalletProviderName, WalletProviderRdns,
 } from '@imtbl/checkout-sdk';
 import {
@@ -9,11 +10,12 @@ import {
 import { motion } from 'framer-motion';
 import { useTranslation } from 'react-i18next';
 import { Web3Provider } from '@ethersproject/providers';
+import { UnableToConnectDrawer } from 'components/UnableToConnectDrawer/UnableToConnectDrawer';
+import { ChangedYourMindDrawer } from 'components/ChangedYourMindDrawer/ChangedYourMindDrawer';
 import { ConnectWidgetViews } from '../../../context/view-context/ConnectViewContextTypes';
 import { ConnectActions, ConnectContext } from '../context/ConnectContext';
 import { WalletItem } from './WalletItem';
 import {
-  SharedViews,
   ViewActions,
   ViewContext,
 } from '../../../context/view-context/ViewContext';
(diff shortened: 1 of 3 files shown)


packages/checkout/widgets-lib/src/widgets/connect/components/WalletList.tsx
```diff
@@ -316,6 +333,20 @@ export function WalletList(props: WalletListProps) {
...
+        checkout={checkout!}
+        onCloseDrawer={() => setShowChangedYourMindDrawer(false)}
+        onTryAgain={onChosenProviderDetailChange}
+      />
+
+      <UnableToConnectDrawer
+        visible={showUnableToConnectDrawer}
+        checkout={checkout!}
+        onCloseDrawer={() => setShowUnableToConnectDrawer(false)}
+        onTryAgain={() => setShowUnableToConnectDrawer(false)}
```
sharonsheah, line 348: Would like your opinion around this - do you think when there is an unknown/other error than user rejected error, do we want try again button to just close the drawer? 

I'm just thinking because we might not know what the error is, maybe we want to restart the whole wallet selection flow rather than keep trying to call the same code that retries a specific wallet connection flow.

If this is the behaviour we want (close drawer), I can make changes to the drawer component itself so that we don't have to pass in the same thing for `onCloseDrawer` and `onTryAgain`.

packages/checkout/widgets-lib/src/widgets/connect/components/WalletList.tsx
```diff
@@ -150,22 +156,24 @@ export function WalletList(props: WalletListProps) {
...
       // eslint-disable-next-line no-console
       console.error('Connect unknown error', err);
 
-      viewDispatch({
-        payload: {
-          type: ViewActions.UPDATE_VIEW,
-          view: { type: SharedViews.ERROR_VIEW, error: err },
-        },
-      });
+      setShowUnableToConnectDrawer(true);
```
sharonsheah, line 176: Note: replacing generic error view with the new unable to connect drawer here

packages/checkout/widgets-lib/src/widgets/connect/components/WalletList.tsx
sharonsheah: But also might need to consider the copy of the Unable to connect drawer - currently it says: `Check your wallet and try again, or close and try a different option`.

ZacharyCouchman (approved):

packages/checkout/widgets-lib/src/widgets/connect/components/WalletList.tsx
dreamoftrees: yeah I was thinking if there was an unknown error we would bomb out to the whole error view screen. But might be hard to tell the variety of errors we get here..

dreamoftrees (approved):
Nice work @sharonsheah !

PR #1585, opened 2024-03-09
feat: cache eth_chainId

andrearampin:

- [x] `breaking-change:` if you have introduced modification that necessitates immediate adjustments by this SDK's users to their applications, clients, or integrations to avert disruptions to existing features or functionalities.

# Summary

https://immutable.atlassian.net/wiki/spaces/PA/pages/2451111964/Passport+eth+chainId

## Added 

- Cache layer for `eth_chainId`

## Changed

- `ZkEvmProvider` instance of `StaticJsonRpcProvider` 

### DIFF
--- a/packages/passport/sdk/src/zkEvm/zkEvmProvider.ts
+++ b/packages/passport/sdk/src/zkEvm/zkEvmProvider.ts
@@ -225,7 +225,6 @@ export class ZkEvmProvider implements Provider {
           guardianClient: this.#guardianClient,
         });
       }
-      // Pass through methods
       case 'eth_chainId': {
         // Call detect network to fetch the chainId so to take advantage of
         // the caching layer provided by StaticJsonRpcProvider.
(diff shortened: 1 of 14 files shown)


packages/passport/sdk/src/zkEvm/zkEvmProvider.ts
```diff
@@ -219,11 +220,29 @@ export class ZkEvmProvider implements Provider {
...
+        // the provider to re-fetch the chainId from remote.
+        return new Promise<string>((resolve, reject) => {
+          (async () => {
+            try {
+              const { chainId } = await this.#staticJsonRpcProvider.detectNetwork();
+              resolve(utils.hexlify(chainId));
+            } catch (err) {
+              reject(err);
+            }
+          })();
```
haydenfowler: What's the reason for wrapping this in a Promise & a try/catch? This can be simplified to the following without any functional difference, can't it?

```suggestion
        const { chainId } = await this.#staticJsonRpcProvider.detectNetwork();
        return utils.hexlify(chainId);
```

haydenfowler (approved):
LGTM 🙌 tested locally and everything appears to work as expected.

PR #1592, opened 2024-03-15
fix: [Sale Widget] Unable to purchase without gas using a passport wallet

jhesgodi:
# Summary
Fix failure to allow payments without gas when using a passport wallet


dreamoftrees (approved):
👍

mimi-imtbl (approved):

PR #1593, opened 2024-03-15
[GASBT-10][Sale Widget] Delay calls to smart checkout until token address is set

jhesgodi:
# Summary
Delay calls to smart checkout until token address was set

# Fixes bug
Sale widget was throwing error `Can't check wallet balances` on first time, given that calls to get `client-config` were running slow in sandbox.


dreamoftrees (approved):
👍

mimi-imtbl (approved):

PR #1601, opened 2024-03-18
fix: [GPR-484][Sale Widget] Remove Insufficient coins banner

jhesgodi:
# Summary
- Remove insufficient coins banner, and replace it with a redirect to Top Up View with custom copy
 
# Changes
## Before
Screenshot 2024-03-18 at 5 56 57 pm

## After 
image


packages/checkout/widgets-lib/src/views/top-up/TopUpView.tsx
```diff
@@ -27,25 +31,30 @@ import {
...
   analytics: {
-    userJourney: UserJourney
-  },
-  onCloseButtonClick: () => void,
-  onBackButtonClick?: () => void,
+    userJourney: UserJourney;
+  };
+  onCloseButtonClick: () => void;
+  onBackButtonClick?: () => void;
+  heading?: [key: string, options?: $Dictionary];
```
jhesgodi, line 56: @dreamoftrees 
after testing this, I noticed that if we pass an already translated string it wont re-render if you set a different language

So I opted to pass an array and use give the params to the translation function. Not adding the `defaultValue` as for this use case this component has the default copy.

packages/checkout/widgets-lib/src/locales/en.json
```diff
@@ -59,11 +59,7 @@
     },
     "ERROR_VIEW": {
       "heading": "Something's gone wrong",
-      "body": [
-        "You can try again or contagact",
-        "support",
-        "for help."
-      ],
+      "body": ["You can try again or contagact", "support", "for help."],
```
sharonsheah, line 62:
```suggestion
      "body": ["You can try again or contact", "support", "for help."],
```

sharonsheah (approved):

mimi-imtbl (approved):

PR #1615, opened 2024-03-22
Fix:gpr-494 sale widget failure event bug

mimi-imtbl:

# Summary
**Currently:** we're skipping the wait for txn settlement if the wallet is MM to speed up the user signing process. We emit `success` event even when the transaction has failed.

**Bug:**  we don't wait for the transactions to be mined hence we don't obtain the transaction receipt, which contains the status of the transaction (0 for failure, 1 for success). So since we omit the wait for MM, `execute` function proceeds without knowing the outcome of the transaction's execution on the blockchain. We would only know that the transaction was successfully sent and accepted for mining. 

**Solution:** remove the `waitForTrnsactionSettlement`, wait for each transaction to be mined (downside is that it will take a bit more time)


mimi-imtbl:
decided to provide an option where you can pass in `waitFulfillmentSettlements`

PR #1620, opened 2024-03-25
fix: [NO-JIRA] Onramp bug setting cryptoCurrencyCode instead of defaultCryptoCurrency

mimi-imtbl:
# Summary
Fix bug with onramp setting the tokenSymbol to `cryptoCurrencyCode` instead of `defaultCryptoCurrency`


sharonsheah (approved):

jhesgodi (changes requested):
Sorry @mimi-imtbl passing the `cryptoCurrencyCode` is required for prefilling crypto amounts.

the solution should be around, not having both `defaultCryptoCurrency` and `cryptoCurrencyCode` at the same time.

PR #1626, opened 2024-03-26
chore: upgrade to seaport-js v4

Sam-Jeston:
# Summary
Upgrade from seaport-js v3 to v4. There are no interface differences between these versions, however v4 fixes partial fill bugs and adds support for Seaport 1.6, while retaining support for Seaport 1.5

### DIFF
--- a/packages/orderbook/src/seaport/seaport-lib-factory.ts
+++ b/packages/orderbook/src/seaport/seaport-lib-factory.ts
@@ -48,11 +48,9 @@ export class SeaportLibFactory {
   ) { }
 
   create(orderSeaportVersion?: SeaportVersion, orderSeaportAddress?: string): SeaportLib {
-    const seaportVersion = orderSeaportVersion ?? SEAPORT_CONTRACT_VERSION_V1_5;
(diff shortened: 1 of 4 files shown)


lfportal (dismissed):

lfportal (approved):

packages/orderbook/src/seaport/seaport-lib-factory.ts
```diff
@@ -48,11 +48,9 @@ export class SeaportLibFactory {
   ) { }
 
   create(orderSeaportVersion?: SeaportVersion, orderSeaportAddress?: string): SeaportLib {
-    const seaportVersion = orderSeaportVersion ?? SEAPORT_CONTRACT_VERSION_V1_5;
     const seaportContractAddress = orderSeaportAddress ?? this.defaultSeaportContractAddress;
 
     return new SeaportLib(convertToV6Provider(this.provider), {
-      seaportVersion,
```
Sam-Jeston, line 55: seaport-js v4 only supports 1.5 and 1.6, which have identical ABIs, so the version is no longer passed in when initializing the lib

carleeto (approved):

lfportal (approved):

PR #1627, opened 2024-03-26
feat: [Sale Widget] Skip sending approve txn to wallet when allowance balance is sufficient

jhesgodi:
# Summary
In some instances the wallet already has an allowance balance sufficient to cover the spent request (order amounts).
This enhancement seeks to skip sending an approve request to the wallet, hence reducing the number of steps needed by the user to sign and complete the order.

Instances where allowance becomes sufficient, may include:
- User has allowed a big amount thru a prev transaction (possible with metamask)
- User was already approved the allowance, but cancel during execute and wants to retry, hence they should not approve again

## Added 
- Add a filter function that removes the approve transaction from the transactions array if sufficient allowance is found on the ERC20 contract


packages/checkout/widgets-lib/src/widgets/sale/functions/signUtils.ts
```diff
@@ -0,0 +1,49 @@
...
+        signerAddress,
+        approveTxn.params.spender,
+      ]),
+    });
+
+    const currentAmount = BigNumber.from(allowance);
+    const desiredAmount = approveTxn.params.amount ? BigNumber.from(approveTxn.params.amount) : BigNumber.from(0);
+
+    const isAllowed = currentAmount.gte(BigNumber.from('0')) && currentAmount.gte(desiredAmount);
+    console.log('🚀 ~ isAllowed:', isAllowed, currentAmount.toString(), desiredAmount.toString()); // eslint-disable-line
```
mimi-imtbl, line 37: should we remove logs?

mimi-imtbl (approved):
🔥

luads (approved):

mimi-imtbl (approved):

luads (approved):

PR #1630, opened 2024-03-27
feat: [WT-2184] Passport SDK zkEVM provider support for EIP-6963

dreamoftrees:
# Summary
Passport SDK zkEVM provider now supports announcing the zkEVM provider via EIP-6963.
EIP-6963 is a multi injection provider discovery compliant way to inject the Passport zkEVM EIP-1193 provider. Connecting the zkEVM provider to Checkout Widgets, dApps, and other connectors.
[WT-2184](https://immutable.atlassian.net/browse/WT-2184)

## Added 
EIP-6963 announce support in the Passport SDK zkEVM connectEvm function.


carleeto (approved):

carleeto (dismissed):

jhesgodi (dismissed):

packages/passport/sdk/src/zkEvm/types.ts
```diff
@@ -111,3 +111,68 @@ export type AccountsChangedEvent = Array<string>;
...
+  data: unknown
+};
+
+export type EIP1193EventMap = {
+  accountsChanged(accounts: string[]): void
+  chainChanged(chainId: string): void
+  connect(connectInfo: ProviderConnectInfo): void
+  disconnect(error: ProviderRpcError): void
+  message(message: ProviderMessage): void
+};
```
haydenfowler, line 153: These types don't seem to be used anywhere

packages/passport/sdk/src/Passport.ts
```diff
@@ -137,15 +138,28 @@ export class Passport {
     return this.passportImxProviderFactory.getProvider();
   }
 
-  public connectEvm(): Provider {
-    return new ZkEvmProvider({
+  public connectEvm(options: {
+    announceProvider: boolean
+  } = {
+    announceProvider: true,
+  }): Provider {
```
haydenfowler, line 145: Can we add a type for this please? Otherwise consumers may have to add their own

dreamoftrees: @haydenfowler I can, although I was attempting to match the current approach in that class with  the same inline typing as `login` and `loginWithDeviceFlow` method. Do you want me to keep that consistent or add a new type for this method?

packages/passport/sdk/src/zkEvm/types.ts
dreamoftrees: These types are used inside the `announceProvider` function found in `zkEVM/provider/eip6963.ts`

haydenfowler: These are very similar to other type definitions that we have within the same file. Is it not possible to update `EIP6963ProviderDetail` to use the existing `Provider` type? That way, we only need to add interfaces for `EIP6963ProviderDetail`, `EIP6963ProviderInfo` & `EIP6963AnnounceProviderEvent`.

packages/passport/sdk/src/Passport.ts
```diff
@@ -137,15 +138,28 @@ export class Passport {
     return this.passportImxProviderFactory.getProvider();
   }
 
-  public connectEvm(): Provider {
-    return new ZkEvmProvider({
+  public connectEvm(options: {
+    announceProvider: boolean
```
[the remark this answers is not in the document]
dreamoftrees: It gives more control to the developer around how they want to consume the provider.. they might have already integrated with Passport and didn't want to announce the provider again. Announcing the provider means it would potentially show up a list of injected providers within other widgets, wagmi injected connectors etc.

packages/passport/sdk/src/zkEvm/provider/eip6963.ts
```diff
@@ -0,0 +1,26 @@
...
+  const event: CustomEvent<EIP6963ProviderDetail> = new CustomEvent(
+    'eip6963:announceProvider',
+    { detail: Object.freeze(detail) },
+  ) as EIP6963AnnounceProviderEvent;
+
+  window.dispatchEvent(event);
+
+  const handler = () => window.dispatchEvent(event);
+  window.addEventListener('eip6963:requestProvider', handler);
+  return () => window.removeEventListener('eip6963:requestProvider', handler);
```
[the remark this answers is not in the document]
dreamoftrees: @Jon-Alonso This was the original implementation carried over from `mipd` library, it's a cleanup function to remove the listener. Within this context I don't see us needed it.. I've removed it.
But I guess we make the assumption that people aren't going to be calling connectEvm() more than once? otherwise we potentially add multiple listeners without ever cleaning it up.. and sending different instances of the provider.

packages/passport/sdk/src/zkEvm/types.ts
```diff
@@ -111,3 +111,68 @@ export type AccountsChangedEvent = Array<string>;
 export interface ProviderEventMap extends Record<string, any> {
   [ProviderEvent.ACCOUNTS_CHANGED]: [AccountsChangedEvent];
 }
+
+export interface EIP1193Provider {
```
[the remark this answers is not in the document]
dreamoftrees: @Jon-Alonso I was looking around but we only have the `Provider` type and it has a number of variations that don't fit the EIP-1193 spec:

- `isPassport` is passport provider specific.
- `request()` arguments params value is defined as ` params?: Array<any>;` but should be defined as
``` readonly params?: readonly unknown[] | object;``` as per EIP-1193
- `send` and `sendAsync` are depreciated methods that are not part of the EIP-1193 interface.

As I'm adding support for the EIP-6963 spec and wanted to use a correct EIP-1193 interface.
One option here could be to rename `Provider` to `PassportProvider` and tweak the `ZkEvmProvider` to use this instead?

```
export class ZkEvmProvider implements EIP1193Provider, PassportProvider {
```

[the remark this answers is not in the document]
dreamoftrees: @Jon-Alonso I outlined above why I didn't use the `Provider` type here because I was being explicit about exactly what EIP-1193 interface is as [defined here](https://eips.ethereum.org/EIPS/eip-1193).. I suggested a compromise to refactor the Provider to PassportProvider and include the interface. 
I understand the added stuff in Provider is harmless but it's not an EIP-1193 interface only which was my issue using that as EIP-6963 interface calls for that as well. I agree duplication is not desired hence my suggestion to refactor the type and be specific about both so it's clear.

But.. I've removed the interfaces I've added as you requested.

To be honest this simple PR has taken way to much time and energy +3 weeks. If you are happy with my changes I'll merge this stuff in, otherwise I'm going to leave this with you guys. Trying to merge this stuff is like pulling teeth.

jhesgodi (approved):

haydenfowler (approved):

carleeto (approved):

PR #1636, opened 2024-03-28
fix: Tune gas limit for popular tokens

wcgcyx:
# Summary

Tune bridge operation gas limit for ETH, USDC, GOG and IMX.

## Added 

Add gas limit of bridge deposit for ETH, USDC, GOG and IMX.
Add gas limit of ERC20 approval for USDC, GOG and IMX.

## Changed

Use configured gas limit if ETH, USDC, GOG or IMX is passed.

This is an interim fix while we improve the performance of #1555


packages/internal/bridge/sdk/src/types/index.ts
```diff
@@ -110,6 +110,15 @@ export enum BridgeMethodsGasLimit { // @TODO test methods on chain and put corre
   MAP_TOKEN_DESTINATION = 200000,
   FINALISE_WITHDRAWAL = 200000,
   APPROVE_TOKEN = 55000,
+
+  DEPOSIT_SOURCE_ETH = 125000,
+  DEPOSIT_SOURCE_USDC = 153000,
```
ermyas, line 115: Maybe we should the USDC to 158K? I see [this](https://etherscan.io/tx/0x4ffca70b0f96c9b99d63bec790212bb2b94f2821297d7235c3996d77e95b880a) deposit transaction for 157.3K

packages/internal/bridge/sdk/src/types/index.ts
```diff
@@ -110,6 +110,15 @@ export enum BridgeMethodsGasLimit { // @TODO test methods on chain and put corre
...
+  DEPOSIT_SOURCE_GOG = 137000,
```
ermyas, line 116: Maybe we should revise this to 142K gas for GOG, given [this transaction](https://etherscan.io/tx/0x657676ef1c2f53a05e5dd8705ba0835379615fcf1da6abdc87faae98ee67a7fd)

dreamoftrees (approved):
Nice work @wcgcyx 👍

ermyas (approved):

PR #1638, opened 2024-04-03
feat: sample app confirmation example for invoking smart contract

scastleman-immutable:
[Jira ticket](https://immutable.atlassian.net/browse/ID-1520)

# Summary
Additional sample app example for calling smart contracts


haydenfowler:
You can use the existing `eth_sendTransaction` screen to call a smart contract, the `transaction` just needs to be a valid JSON transaction (see attached recording).

https://github.com/immutable/ts-immutable-sdk/assets/9940216/4f28d3e1-64c3-4419-8a1a-0689b8a4bb97

carmen0208:
@haydenfowler Yes, understood. 
This feature will help with the testing of the generic screen. 
If we use `eth_sendTransaction`, we have to prepare the data structure and also have to remember to add the 'data' with the same values(instead not undefined or 0x). also, `JSON.stringfy` the data, which can be annoying.

packages/passport/sdk-sample-app/src/components/zkevm/EthSendTransactionExamples/CallSmartContract.tsx
```diff
@@ -0,0 +1,161 @@
...
+import { ethers } from 'ethers';
+import * as encUtils from 'enc-utils';
+
+function CallSmartContract({ disabled, handleExampleSubmitted }: RequestExampleProps) {
+  const { orderbookClient } = useImmutableProvider();
+  const defaultAddress = orderbookClient.config().seaportContractAddress;
+  const [fromAddress, setFromAddress] = useState<string>('');
+  const [toAddress, setToAddress] = useState<string>(defaultAddress);
+  const [toAddressError, setToAddressError] = useState<string>('');
+  const [data, setData] = useState<string>('123');
```
carmen0208, line 19:
```suggestion
  const [data, setData] = useState<string>('1234563343333');
```
the data have to be longer than 10 length. Otherwise, it will return the method_id, which is wrong from the passport MR. 
so for the default data, please put it larger than 10 in length.

You also need to check if the user provides more than 10-length data. 


But to sample the whole process, you can just hardcode the data and alway use the hardcoded data instead of allow user to do the input.

packages/passport/sdk-sample-app/src/components/zkevm/EthSendTransactionExamples/CallSmartContract.tsx
```diff
@@ -0,0 +1,161 @@
...
+import { Accordion, Form } from 'react-bootstrap';
+import InputGroup from 'react-bootstrap/InputGroup';
+import { usePassportProvider } from '@/context/PassportProvider';
+import WorkflowButton from '@/components/WorkflowButton';
+import { RequestExampleProps } from '@/types';
+import { useImmutableProvider } from '@/context/ImmutableProvider';
+import { ethers } from 'ethers';
+import * as encUtils from 'enc-utils';
+
+function CallSmartContract({ disabled, handleExampleSubmitted }: RequestExampleProps) {
```
carmen0208, line 13: Change it to Show Generic Screen.

carmen0208 (dismissed):

packages/passport/sdk-sample-app/src/components/zkevm/EthSendTransactionExamples/ShowGenericConfirmationScreen.tsx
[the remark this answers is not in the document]
scastleman-immutable: If we're going by the types we log for analytics purposes, we don't log `unknown`. We do log `default` though, so we could name this `DefaultTransaction` if we want to follow that schema

carmen0208 (approved):

andrearampin (approved):

PR #1646, opened 2024-04-05
fix: ID-1538 Resolved issue with redirect logout not logging out of Magic

haydenfowler:
# Summary
This PR resolves an issue where users were not being correctly logged out of their Magic wallet.

## Changed
- Passport: Fixed an issue where users were not correctly logged out of their Magic wallet when using the `redirect` logout mode.

### DIFF
--- a/packages/passport/sdk/src/Passport.ts
+++ b/packages/passport/sdk/src/Passport.ts
@@ -228,10 +228,18 @@ export class Passport {
     } catch (err) {
       logger.warn('Failed to logout from confirmation screen', err);
     }
-    await Promise.allSettled([
-      this.authManager.logout(),
-      this.magicAdapter.logout(),
-    ]);
+
+    if (this.config.oidcConfiguration.logoutMode === 'silent') {
+      await Promise.allSettled([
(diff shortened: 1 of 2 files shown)


packages/passport/sdk/src/Passport.ts
```diff
@@ -228,10 +228,18 @@ export class Passport {
...
     }
-    await Promise.allSettled([
-      this.authManager.logout(),
-      this.magicAdapter.logout(),
-    ]);
+
+    if (this.config.oidcConfiguration.logoutMode === 'silent') {
+      await Promise.allSettled([
+        this.authManager.logout(),
+        this.magicAdapter.logout(),
```
pano-skylakis, line 235: QQ related to this https://imtbl.slack.com/archives/C06SX4AUBNE/p1712280521367519 - Does this also need to be in the order of

1. `magicAdapter.logout()`
2. `authManger.logout()`

Also, do we need to prevent calling `authManager.logout` if the call to `magicAdapter.logout` failed? Currently, one of these could reject and the other could resolve - we do not check the results

PR #1648, opened 2024-04-05
feat: gpr 545 smart checkout skip gas checks

mimi-imtbl:

# Summary

Added a `skipGasCalculation` flag for smart checkout to skip gas calculations (needed when Passport sponsors gas).


dreamoftrees (changes requested):
Nice work @mimi-imtbl. I think it would be good to add an additional test here within `smartCheckout.test.ts` that tests the result of `skipGasCalculation = true` and makes sure that the gasCalculator is not called.

packages/checkout/sdk/src/sdk.ts
```diff
@@ -546,6 +546,7 @@ export class Checkout {
       web3Provider,
       itemRequirements,
       params.transactionOrGasAmount,
+      params.skipGasCalculation,
```
jhesgodi, line 549: Given this will be a new SDK property, can we discuss the most suitable name?
`skipGasCalculation` –> look fine at first, but the actual outcome is not around the calculation, but on not requirig gas checks

I think since gas is a requirement we may call it `skipGasRequirement` or `withoutGasFees` or something of the sort.

However, @dreamoftrees I just noticed, if we don't require to check for gas, then `transactionOrGasAmount` could be optional?

If thats the case, removing `transactionOrGasAmount` should make the gas fee optional

packages/checkout/sdk/src/smartCheckout/smartCheckout.ts
```diff
@@ -27,30 +24,46 @@ export const smartCheckout = async (
...
+  const aggregatedAllowances = allowanceAggregator(
+    resolvedAllowances[0],
+    resolvedAllowances[1],
   );
-  if (gasItem !== null) {
-    aggregatedItems.push(gasItem);
-    aggregatedItems = itemAggregator(aggregatedItems);
+
+  let gasItem = null;
+  if (!skipGasCalculation) {
```
jhesgodi, line 57: I can see that the gasCalculator depends on `transactionOrGasAmount`, then making this optional will be enough as a way to flag gas is not required.

I like it simplifies the api, but also wonder if there is something I missing @mimi-imtbl @dreamoftrees ?

packages/checkout/sdk/src/sdk.ts
mimi-imtbl: Hmm, true. There is no point in passing `transactionOrGasAmount` if the gas check will be skipped anyway. Or passing GasAmount as 0, for example. @dreamoftrees to confirm

packages/checkout/sdk/src/smartCheckout/smartCheckout.ts
dreamoftrees: Yep, I think in this context that makes sense and smartCheckout should skips the gasCalculator call.

mimi-imtbl:
Replaced with this PR https://github.com/immutable/ts-immutable-sdk/pull/1650

PR #1653, opened 2024-04-08
feat: [GPR-536][Sale Widget] Get multiple currencies

mimi-imtbl:
# Summary

Add functionality to the sale widget to get multiple currencies. 

## Added 
- Add logic to  `useClientConfig` hook to manage and aggregate currencies 
- Getting lists of currencies from sale BE & checkout config 
- Mark `settlement` vs `swappable` currencies
- Add function to remove duplicated currencies in the aggregated currency list & sort it based on priority: `base` currency -> `settlement` currencies -> `swappable` currencies
- Set `base` currency as `selectedCurrency`
- Add error handling 
- Add tests to utils functions


packages/checkout/widgets-lib/src/widgets/sale/hooks/useCurrency.ts
```diff
@@ -0,0 +1,170 @@
...
+  ClientConfigCurrency,
+  ClientConfigCurrencyConversion,
+  SaleWidgetCurrency,
+  SaleWidgetCurrencyType,
+  SignPaymentTypes,
+} from '../types';
+import { useCheckoutClient } from './useCheckoutClient';
+import { sortCurrencies } from '../utils/sortCurrencies';
+
+type ClientConfigResponse = {
```
jhesgodi, line 21: @mimi-imtbl consider applying the renaming consistently across. although i think this hook better keep the prev name as is not specific to currency at alone

packages/checkout/widgets-lib/src/widgets/sale/hooks/useCurrency.ts
```diff
@@ -0,0 +1,170 @@
...
+  passport,
+  defaultCurrency,
+}: UseClientConfigParams) => {
+  const [selectedCurrency, setSelectedCurrency] = useState<
+  ClientConfigCurrency | undefined
+  >();
+  const [clientConfig, setClientConfig] = useState<ClientConfig>(defaultClientConfig);
+  const [allCurrencies, setAllCurrencies] = useState<SaleWidgetCurrency[]>([]);
+
+  const checkoutClient = useCheckoutClient({ passport, environment });
```
jhesgodi, line 84: We can optimise things here, the checkout instance was already created when the widget instance is created.
Have a look [here](https://github.com/immutable/ts-immutable-sdk/blob/ffd504b90afb04feeb8b08de60ebde8a4dbcbba6/packages/checkout/widgets-lib/src/widgets/BaseWidgetRoot.ts#L22-L22), notice the `BaseWidgetRoot` class provides the checkout instance. so we can pass it down to SaleWidget and then hold it on the Context object.

jhesgodi (changes requested):

packages/checkout/widgets-lib/src/widgets/sale/hooks/useCurrency.ts
```diff
@@ -0,0 +1,170 @@
...
+    if (!environment || !environmentId) return;
+
+    const fetchSwappableCurrencies = async () => {
+      if (!checkoutClient || !passport) {
+        return [];
+      }
+      try {
+        const passportProvider = passport.connectEvm();
+        const zkEvmProvider = new providers.Web3Provider(passportProvider);
+        const checkoutNetworkInfo = await checkoutClient.getNetworkInfo({
```
jhesgodi, line 96: ideally we should use the already existing provider.

packages/checkout/widgets-lib/src/widgets/sale/hooks/useCurrency.ts
```diff
@@ -0,0 +1,170 @@
...
+  currency_conversion: {
+    [key: string]: {
+      amount: number;
+      name: string;
+      type: string;
+    };
+  };
+};
+
+const toClientConfig = (response: ClientConfigResponse): ClientConfig => ({
```
jhesgodi, line 39: since this logic became complex, I'll suggest we move into using pattern of creating utility functions within the `/functions` directory. And start adding some test to each utility.

example
https://github.com/immutable/ts-immutable-sdk/blob/853aaee1eb1b00f61d95735576e8b1c8006f949b/packages/checkout/widgets-lib/src/widgets/swap/functions/FetchQuote.ts

packages/checkout/widgets-lib/src/widgets/sale/hooks/useCurrency.ts
```diff
@@ -0,0 +1,170 @@
...
+      }
+    };
+
+    (async () => {
+      const [swappableCurrencies, settlementCurrencies] = await Promise.all([
+        fetchSwappableCurrencies(),
+        fetchSettlementCurrencies(),
+      ]);
+
+      const combinedCurrencies: SaleWidgetCurrency[] = [
```
jhesgodi, line 146: need to be careful here, we require the final list doesn't have duplicates but also to respect the priority order.
For example, ETH could be returned from the allowlist and `/client-config` in such case it should be taken as swappable

packages/checkout/widgets-lib/src/widgets/sale/utils/sortCurrencies.ts
```diff
@@ -0,0 +1,24 @@
+import { SaleWidgetCurrency } from '../types';
```
jhesgodi, line 1: nice one 🔥

packages/checkout/widgets-lib/src/widgets/sale/hooks/useCurrency.ts
```diff
@@ -0,0 +1,170 @@
...
+        const config = toClientConfig(data);
+        setClientConfig(config);
+
+        return config.currencies.map((currency) => ({
+          ...currency,
+          currencyType: SaleWidgetCurrencyType.SETTLEMENT,
+        }));
+      } catch (error) {
+        // eslint-disable-next-line no-console
+        console.error('Error fetching settlement currencies:', error);
```
jhesgodi, line 135: this may cause the widget to break.
Can you check if throwing here is captured by an error screen?

packages/checkout/widgets-lib/src/widgets/sale/hooks/useCurrency.ts
mimi-imtbl: ah yes! good one!

packages/checkout/widgets-lib/src/widgets/sale/hooks/useCurrency.ts
mimi-imtbl: This is how the `useClientConfig` is currently handling error - a `console.warn` and passing the empty default `clientConfig`. I'll add error handling for it to go straight to the error screen if the error occurs here as transak/coins flow both wouldn't work without `contractId`

packages/checkout/widgets-lib/src/widgets/sale/functions/sortAndDeduplicateCurrencies.test.ts
```diff
@@ -0,0 +1,197 @@
+import { SaleWidgetCurrency, SaleWidgetCurrencyType } from '../types';
+import { sortAndDeduplicateCurrencies } from './sortAndDeduplicateCurrencies';
+
+describe('sortAndDeduplicateCurrencies', () => {
+  it('sortAndDeduplicateCurrencies', () => {
+    const allCurrencies: SaleWidgetCurrency[] = [
+      {
```
jhesgodi, line 7: @mimi-imtbl not a blocker, the input array is already sorted, can you scramble the order to make sure the sort function is actually doing its job?

packages/checkout/widgets-lib/src/widgets/sale/functions/sortAndDeduplicateCurrencies.test.ts
```diff
@@ -0,0 +1,197 @@
...
+      },
+      {
+        name: 'zkCORE',
+        symbol: 'zkCORE',
+        address: '0x4B96E7b7eA673A996F140d5De411a97b7eab934E',
+        decimals: 18,
+        currencyType: SaleWidgetCurrencyType.SWAPPABLE,
+      },
+      {
+        name: 'USDC',
```
jhesgodi, line 178: shouldn't it be deduplicated?

packages/checkout/widgets-lib/src/widgets/sale/functions/sortAndDeduplicateCurrencies.test.ts
```diff
@@ -0,0 +1,197 @@
...
+      {
+        base: true,
+        decimals: 6,
+        name: 'USDC',
+        address: '0x3b2d8a1931736fc321c24864bceee981b11c3c57',
+        exchangeId: 'usd-coin',
+        currencyType: SaleWidgetCurrencyType.SETTLEMENT,
+      },
+      {
+        base: false,
```
jhesgodi, line 16: @mimi-imtbl worth making sure that if 2 bases are returned, it will only work with 1, first one found mostlikely

packages/checkout/widgets-lib/src/widgets/sale/functions/transformToClientConfig.test.ts
```diff
@@ -0,0 +1,112 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import { ClientConfig, SignPaymentTypes } from '../types';
+import {
+  ClientConfigResponse,
+  transformToClientConfig,
+} from './transformToClientConfig';
+
+describe('transformToClientConfig', () => {
+  it('transformToClientConfig', () => {
```
jhesgodi, line 9: good to describe what being tested 
`it('should return input object with camel case keys')`

packages/checkout/widgets-lib/src/widgets/sale/functions/sortAndDeduplicateCurrencies.test.ts
```diff
@@ -0,0 +1,197 @@
+import { SaleWidgetCurrency, SaleWidgetCurrencyType } from '../types';
+import { sortAndDeduplicateCurrencies } from './sortAndDeduplicateCurrencies';
+
+describe('sortAndDeduplicateCurrencies', () => {
+  it('sortAndDeduplicateCurrencies', () => {
```
jhesgodi, line 5: best to describe what being tested
`it('should remove duplicates and sort currencies by: base -> settlement -> swappable')`

jhesgodi (approved):
🟢 LGTM, approve with minor comments @mimi-imtbl

packages/checkout/widgets-lib/src/widgets/sale/functions/sortAndDeduplicateCurrencies.test.ts
mimi-imtbl: yep, fixed this edge case

mimi-imtbl: Based on the current implementation, it will pick the later `base: true`. This shouldn't be an issue, I'd assume. I'd prefer to not make this logic more complicated, will leave this for now

dreamoftrees (approved):
Great work @mimi-imtbl

jhesgodi (approved):

PR #1660, opened 2024-04-11
fix: elliptic import

CodeSchwert:

# Summary

[DX-2760](https://immutable.atlassian.net/browse/DX-2760)

Proper fix for the `elliptic` package when using the SDK in ES module files.

This change should not affect customers. It simply changes where x-client stark methods are being exported from internally. The external interface has not changed.

## Removed
- Core SDK as a dependency from x-client

## Fixed
- x-client dependency elliptic should work with ES module files

The `@types/elliptic` package was breaking the build.

[DX-2760]: https://immutable.atlassian.net/browse/DX-2760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


carleeto (approved):

packages/x-client/tsconfig.json
```diff
@@ -18,5 +18,5 @@
     "skipLibCheck": true
   },
   "include": ["src"],
-  "exclude": ["node_modules", "dist"]
+  "exclude": ["node_modules", "dist", "node_modules/elliptic"]
```
cajames, line 21: Why is this required?

cajames (approved):

packages/x-client/tsconfig.json
CodeSchwert: Complains about missing types otherwise

PR #1662, opened 2024-04-13
GPR-549 - set the fromToken in the swap widget

andrearampin:

# Summary

Patch a bug that prevents to preselect the from and to tokens + allow native to be selected independently by the symbol (there should not be any need to construct the option using the symbol (identified the bug because sandbox IMX symbol is `tIMX` and not `IMX`)

BEFORE

https://github.com/immutable/ts-immutable-sdk/assets/4301501/bd080588-e6c2-4442-8b5c-6876a8c0fdf6

AFTER

https://github.com/immutable/ts-immutable-sdk/assets/4301501/017c4f72-085e-461e-aba9-14e893abfc0a

## Fixed

- Patched assign `tokenFrom` in the Swap widget logic
- `formatTokenOptionsId` only use `NATIVE` for native token


jhesgodi (approved):
🔥 thanks @andrearampin

dreamoftrees (approved):
👍 Thanks @andrearampin

PR #1663, opened 2024-04-15
feat: SMR-2559 dynamic withdrawal estimation

wcgcyx:
# Summary

Enable dynamic estimation for bridge deposit and withdrawal. 

1. This PR added a new interface to the Bridge SDK: `getUnsignedBridgeBundledTx`, which is used to obtain both the optional approval transaction and an unsigned deposit or withdrawal transaction. This combines the original `getUnsignedApproveBridgeTx` and `getUnsignedBridgeTx`.

2. General optimization of the SDK performance is also done as part of this PR.

3. Dynamic gas estimation is now the default way of estimating bridge fees for both deposit and withdrawal.

This is a rework of PR https://github.com/immutable/ts-immutable-sdk/pull/1555.


packages/internal/bridge/sdk/src/constants/bridges.ts
```diff
@@ -2,6 +2,11 @@
 
 import { AxelarChainDetails, BridgeInstance } from '../types';
 
+/**
+ * @constant {string} NATIVE - The native token representation.
+ */
+export const NATIVE = 'NATIVE';
```
ermyas, line 8: the use of this string constant feels like it will be error-prone. It is declared elsewhere as well (in lowercase). have we considered abstracting native tokens into a type instead?

packages/internal/bridge/sdk/src/tokenBridge.ts
```diff
@@ -547,143 +477,640 @@ export class TokenBridge {
...
+    if (this.isWrappedIMX(token, this.config.bridgeInstance.childChainID) && allowance.lt(amount)) {
+      contractToApprove = token;
+      const erc20Contract: ethers.Contract = await withBridgeError<ethers.Contract>(
+        async () => new ethers.Contract(token, ERC20, this.config.childProvider),
+        BridgeErrorType.PROVIDER_ERROR,
+      );
+      const data: string = await withBridgeError<string>(async () => erc20Contract.interface
+        .encodeFunctionData('approve', [
+          this.config.bridgeContracts.childERC20Bridge,
+          amount.sub(allowance),
```
ermyas, line 728: I don't believe this is correct. Approve would replace the [old allowance value not increment it](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol#L287)... so instead of the delta, I think we should be using the amount here?

packages/internal/bridge/sdk/src/tokenBridge.ts
```diff
@@ -302,82 +296,18 @@ export class TokenBridge {
...
-    // Create the unsigned transaction for the approval
-    const unsignedTx: ethers.providers.TransactionRequest = {
-      data,
-      to: req.token,
-      value: 0,
-      from: req.senderAddress,
-      chainId: parseInt(req.sourceChainId, 10),
-    };
-
+    const res = await this.getUnsignedBridgeBundledTx({
```
ermyas, line 299: is my understanding correct that calling `getUnsignedBridgeBundledTx()` here would perform a number of subsequent calls (through `getUnsignedBridgeBundledTxPrivate()`) that are not related to getting the the approval transaction? I imagine these calls are probably expensive?

packages/internal/bridge/sdk/src/tokenBridge.ts
wcgcyx: Yep,  this is a previous bug, just fixed.

packages/internal/bridge/sdk/src/constants/bridges.ts
Benjimmutable: Good question. If we look at how the DEX SDK does it, [they have a specific type for the native token](https://github.com/immutable/ts-immutable-sdk/blob/main/packages/internal/dex/sdk/src/types/index.ts#L126-L137), which is then used in `Coin`, which is of type `ERC20 | Native`.

They still have to, at some point, [deal with some sort of stringy way of representing](https://github.com/immutable/ts-immutable-sdk/blob/main/packages/internal/dex/sdk/src/exchange.ts#L145-L156) that we want to use the native asset, though. Not sure if it's avoidable entirely.

packages/internal/bridge/sdk/src/tokenBridge.ts
```diff
@@ -547,143 +477,640 @@ export class TokenBridge {
...
 
-    const data = await this.getTxData(
+  private async getDynamicDepositGas(
+    sourceChainId: string,
+    sender: string,
+    recipient: string,
+    token: FungibleToken,
+    amount: ethers.BigNumber,
+  ): Promise<DynamicGasEstimatesResponse> {
+    const simulations: Array<any> = [];
```
Benjimmutable, line 859: Should we consider using stricter types here?

packages/internal/bridge/sdk/src/tokenBridge.ts
```diff
@@ -547,143 +477,640 @@ export class TokenBridge {
...
+
+    if (axiosResponse.data.simulation_results.length !== simulations.length) {
+      throw new BridgeError(
+        'Estimating gas failed with mismatched responses',
+        BridgeErrorType.TENDERLY_GAS_ESTIMATE_FAILED,
+      );
+    }
+
+    const tenderlyGasEstimatesRes = {} as DynamicGasEstimatesResponse;
+    const simResults = axiosResponse.data.simulation_results;
```
Benjimmutable, line 941: This could be moved out into a separate function e.g. `submitTenderlySimulations`.

packages/internal/bridge/sdk/src/tokenBridge.ts
```diff
@@ -547,143 +477,640 @@ export class TokenBridge {
...
+        tenderlyGasEstimatesRes.approvalGas = simResults[0].simulation.gas_used;
+        tenderlyGasEstimatesRes.sourceChainGas = simResults[1].simulation.gas_used;
+      }
+    } else {
+      throw new BridgeError(
+        `Estimating gas failed with unexpected number responses ${axiosResponse.data.simulation_results.length}`,
+        BridgeErrorType.TENDERLY_GAS_ESTIMATE_FAILED,
+      );
+    }
+    return tenderlyGasEstimatesRes;
```
Benjimmutable, line 974: This result error processing could be moved into a `processTenderlySimulations` function

packages/internal/bridge/sdk/src/tokenBridge.ts
```diff
@@ -547,143 +477,640 @@ export class TokenBridge {
...
+    );
+    const executeData = await withBridgeError<string>(
+      async () => axelarAdapterContract.interface
+        .encodeFunctionData('execute', [commandId, sourceChain, sourceAddress, payload]),
+      BridgeErrorType.INTERNAL_ERROR,
+    );
+
+    // Build simulation
+    const axelarGateway = this.getAxelarGateway(destinationChainId);
+    const simulations = [{
```
Benjimmutable, line 1030: Not sure if any functions could be written to reduce the code duplication with here and the simulation stuff in getDynamicDepositGas

packages/internal/bridge/sdk/src/tokenBridge.ts
```diff
@@ -751,6 +1178,116 @@ export class TokenBridge {
     }
   }
 
+  private getWrappedIMX(source: string) {
```
Benjimmutable, line 1181: All of these functions could probably move into a utils file.

packages/internal/bridge/sdk/src/constants/bridges.ts
wcgcyx: Yes, the intention is to avoid any breaking change. Checkout [uses the string of 'native'](https://github.com/immutable/ts-immutable-sdk/blob/e573fcc6c926e85f29f8160be3273fff6fd58580/packages/checkout/widgets-lib/src/lib/constants.ts#L7)  across different components so this I think ideally this should be done in a separate PR to avoid any potential issues of this change.

packages/internal/bridge/sdk/src/tokenBridge.ts
wcgcyx: Yes, that's right. I was thinking about deprecating these two function calls `getUnsignedApproveBridgeTx` and `getUnsignedBridgeTx` or maybe there is a use case for each call individually? 🤔  WDYT?

packages/internal/bridge/sdk/src/constants/bridges.ts
```diff
@@ -33,6 +38,84 @@ export const IMTBL_ZKEVM_NATIVE_TOKEN_ADDRESS = '0x00000000000000000000000000000
  */
 export const ETHEREUM_NATIVE_TOKEN_ADDRESS = '0x0000000000000000000000000000000000000eee';
 
+/**
+ * @constant {string} WITHDRAW_SIG - Signature of the withdraw command.
+ */
+export const WITHDRAW_SIG = '0x7a8dc26796a1e50e6e190b70259f58f6a4edd5b22280ceecc82b687b8e982869';
```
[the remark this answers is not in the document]
wcgcyx: This is the withdrawal defined here https://github.com/immutable/zkevm-bridge-contracts/blob/c1928b9e520493d985bb8a66a9c5356774966fe0/src/child/ChildERC20Bridge.sol#L64

packages/internal/bridge/sdk/src/constants/bridges.ts
```diff
@@ -33,6 +38,84 @@ export const IMTBL_ZKEVM_NATIVE_TOKEN_ADDRESS = '0x00000000000000000000000000000
...
+
+/**
+ * @constant {string} SLOT_PREFIX_CONTRACT_CALL_APPROVED - The prefix of the storage slot to store contract call approved mapping.
+ */
+export const SLOT_PREFIX_CONTRACT_CALL_APPROVED = '0x07b0d4304f82012bd3b70b1d531c160e326067c90829e2a3d386722ad10b89c3';
```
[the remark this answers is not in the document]
wcgcyx: This is `keccak256(toUtf8Bytes('contract-call-approved'))` the prefix of the storage slot https://github.com/axelarnetwork/axelar-cgp-solidity/blob/8b0ea6693178457b7c155b1efb3ecba6236857cc/contracts/AxelarGateway.sol#L67C31-L67C60 which we are using in the state override in tenderly simulation to mark the simulated withdrawal transaction as approved.

packages/internal/bridge/sdk/src/tokenBridge.ts
```diff
@@ -110,12 +148,16 @@ export class TokenBridge {
...
    */
   public async getFee(req: BridgeFeeRequest): Promise<BridgeFeeResponse> {
-    await this.validateChainConfiguration();
-
-    if (req.action !== BridgeFeeActions.FINALISE_WITHDRAWAL) {
-      await this.validateChainIds(req.sourceChainId, req.destinationChainId);
-    }
-    return await this.getFeePrivate(req);
+    const [, , res] = await Promise.all([
+      this.initialise(),
```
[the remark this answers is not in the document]
wcgcyx: Yes, the intention is to avoid any breaking change and how this SDK is being used. Calling `initialise` in each of the calls makes sure this instance is initialised.

packages/internal/bridge/sdk/src/tokenBridge.ts
```diff
@@ -547,143 +477,640 @@ export class TokenBridge {
...
+    let contractToApprove: string | null;
+    let unsignedApprovalTx: ethers.providers.TransactionRequest | null;
+    let sourceChainFee: ethers.BigNumber = ethers.BigNumber.from(0);
+    let approvalFee: ethers.BigNumber = ethers.BigNumber.from(0);
+    const bridgeFee: ethers.BigNumber = axelarFee;
+    const imtblFee: ethers.BigNumber = ethers.BigNumber.from(0);
+
+    // Approval required only for WIMX tokens with insufficient allowance.
+    if (this.isWrappedIMX(token, this.config.bridgeInstance.childChainID) && allowance.lt(amount)) {
+      contractToApprove = token;
```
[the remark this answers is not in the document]
wcgcyx: Approval is not required on L2 since the bridge has mint/burn roles of all ERC20s except for WIMX which is deployed separately.

packages/internal/bridge/sdk/src/constants/bridges.ts
Benjimmutable: Yeah we can still accept the string, but then convert it into an appropriate type early on in our SDK's flow

packages/internal/bridge/sdk/src/constants/bridges.ts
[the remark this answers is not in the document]
Benjimmutable: Can we add a little comment saying that it is the keccak256 hash of "WITHDRAW"?

packages/internal/bridge/sdk/src/constants/bridges.ts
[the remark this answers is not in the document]
Benjimmutable: Oh yep sweet

packages/internal/bridge/sdk/src/tokenBridge.ts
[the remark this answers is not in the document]
Benjimmutable: Yep that makes sense.

packages/internal/bridge/sdk/src/tokenBridge.ts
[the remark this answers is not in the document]
Benjimmutable: Oh right, and because we don't currently support L2-native tokens that's a safe assumption. Makes sense.

Benjimmutable (dismissed):

Benjimmutable (approved):

packages/internal/bridge/sdk/src/tokenBridge.ts
dreamoftrees: Would be nice to define the simulation type..
```
{
        network_id,
        estimate_gas,
        simulation_type,
        from,
        to,
        input,
      }

dreamoftrees (approved):
Great work! Just a minor comment from me.

packages/internal/bridge/sdk/src/tokenBridge.ts
Benjimmutable: @dreamoftrees In the backlog https://immutable.atlassian.net/browse/SMR-2719

ermyas (approved):

PR #1664, opened 2024-04-15
feat: GPR-581 - "Advanced Options" in Balance Widget top up screen

luads:
# Summary

Introducing a link to Layerswap in the Toolkit, opening up in a new tab.

image

## Added 

New Top Up feature: advances options for fast bridge and CEX deposits (Layerswap)

### DIFF
--- a/packages/checkout/widgets-lib/src/views/top-up/TopUpView.tsx
+++ b/packages/checkout/widgets-lib/src/views/top-up/TopUpView.tsx
@@ -13,6 +13,7 @@ import {
   GasEstimateType,
   IMTBLWidgetEvents,
 } from '@imtbl/checkout-sdk';
+import { Environment } from '@imtbl/config';
 import { DEFAULT_TOKEN_SYMBOLS } from 'context/crypto-fiat-context/CryptoFiatProvider';
 import { BridgeWidgetViews } from 'context/view-context/BridgeViewContextTypes';
 import { Web3Provider } from '@ethersproject/providers';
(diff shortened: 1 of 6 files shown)


packages/checkout/widgets-lib/src/views/top-up/TopUpView.tsx
```diff
@@ -57,6 +58,11 @@ interface TopUpViewProps {
   subheading?: [key: string, options?: $Dictionary];
 }
 
+export const TOOLKIT_BASE_URL = {
+  [Environment.SANDBOX]: 'https://checkout-playground.sandbox.immutable.com',
+  [Environment.PRODUCTION]: 'https://toolkit.immutable.com',
+};
+
```
mimi-imtbl, line 65: we have a config file under `ts-immutable-sdk/packages/checkout/widgets-lib/src/widgets/sale/utils/config.ts` that we can move this to 👀

mimi-imtbl (commented):
just one smol comment - thanks for picking this up 🔥

packages/checkout/widgets-lib/src/views/top-up/TopUpView.tsx
luads: Is there a similar place (like `packages/checkout/widgets-lib/src/lib/utils.ts`) outside of the sale scope that would be suitable? I was conflicted with this one since it's not necessarily sale specific

mimi-imtbl: ah true, `TopUpView` is outside of the `/widgets` folder - happy for it to stay here for now!

mimi-imtbl (approved):

dreamoftrees (approved):
👍 nice work @luads

PR #1676, opened 2024-04-17
fix: ID-1584 Force confirmation popup to be opened while provider is instantiated

Jon-Alonso:
# Summary
Updated the EVM provider `eth_sendTransaction` and `eth_signTypedData_v4` methods to open the confirmation popup before the ethSigner is initialised. This should prevent browsers from blocking the popup, since the popup will no longer be opened after an async call.

## Changed
On the zkEVM provider:
- eth_sendTransaction
- eth_signTypedData_v4
- eth_signTypedData

Before:

https://github.com/immutable/ts-immutable-sdk/assets/24399271/91840cc7-436a-4e75-b8d6-3a60d299691d

After:

https://github.com/immutable/ts-immutable-sdk/assets/24399271/ff2954be-1b90-4341-af92-5144391d7cb2

### DIFF
--- a/packages/passport/sdk/src/zkEvm/sendTransaction.test.ts
+++ b/packages/passport/sdk/src/zkEvm/sendTransaction.test.ts
@@ -11,7 +11,6 @@ import GuardianClient from '../guardian';
 
 jest.mock('./walletHelpers');
 jest.mock('../network/retry');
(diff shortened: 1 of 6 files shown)


haydenfowler (approved):
LGTM 👍

andrearampin (approved):

PR #1678, opened 2024-04-18
feat: [Sale Widget] Add drafted order summary component

jhesgodi:
# Summary
- Add a draft for order summary component


mimi-imtbl (approved):

packages/checkout/widgets-lib/src/widgets/sale/views/OrderSummary.tsx
```diff
@@ -0,0 +1,87 @@
...
+        type: CryptoFiatActions.SET_TOKEN_SYMBOLS,
+        tokenSymbols,
+      },
+    });
+  }, [cryptoFiatDispatch, allCurrencies]);
+
+  return (
+    <Box>
+      {subView === OrderSummarySubViews.INIT && (
+        <LoadingView
```
dreamoftrees, line 72: Did you consider making this loading just an overlay rather than introducing the sub views? or that wouldn't work?

dreamoftrees (commented):
Just a query about the loading state..

PR #1681, opened 2024-04-18
[feat][sale widget]gpr-593 smart checkout funding routes sufficient

mimi-imtbl:

# Summary

Update smartCheckout to accept `includeFundingRoutesOnSufficient` flag.

## Changed

- `includeFundingRoutesOnSufficient` is by default `false` - not breaking current usage of the SDK
- if `includeFundingRoutesOnSufficient` is `true` - return a `Promise<SmartCheckoutRouter>` to ensure the method remains non-blocking and responsive
- if `includeFundingRoutesOnSufficient` is `false`, behaves as it would originally - not returning any routes.


packages/checkout/sdk/src/smartCheckout/smartCheckout.ts
```diff
@@ -27,21 +24,35 @@ export const smartCheckout = async (
   provider: Web3Provider,
   itemRequirements: ItemRequirement[],
   transactionOrGasAmount?: FulfillmentTransaction | GasAmount,
+  includeFundingRoutesOnSufficient: boolean = false,
```
dreamoftrees, line 27: What do you think about shortening this a little? maybe `allowRoutesOnSufficient` or `sufficientFundingRoutes`?

dreamoftrees: Also just checking this would be a non-breaking change right? because we default to false here and we never returned the routes before if sufficient?

jwhardwick (commented):
Looks good @mimi-imtbl .

I wonder if we can do without the `includeFundingRoutesOnSufficient` flag and just return the `routerPromise` every time it's `sufficient`?

It wouldn't add any execution time and reduces the complexity of the method by removing a flag.

dreamoftrees (dismissed):
Great work @mimi-imtbl I think the work I've done here shouldn't cause too much hassle merging but happy to sync when you need.

PR #1684, opened 2024-04-18
fix: [GPR-594] Updates to cloud image component image resizer url based on environment

dreamoftrees:
# Summary
CloudImage component within the Checkout widgets now correctly switches its CDN image resizer url based on environment.
[GPR-594](https://immutable.atlassian.net/browse/GPR-594)

## Added 
CloudImage component imageResizerUrl attributes.

[GPR-594]: https://immutable.atlassian.net/browse/GPR-594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/checkout/widgets-lib/src/components/ChangedYourMindDrawer/ChangedYourMindDrawer.tsx
+++ b/packages/checkout/widgets-lib/src/components/ChangedYourMindDrawer/ChangedYourMindDrawer.tsx
@@ -11,6 +11,7 @@ import { Checkout } from '@imtbl/checkout-sdk';
 import { Environment } from '@imtbl/config';
 import { getRemoteImage } from 'lib/utils';
 import { useTranslation } from 'react-i18next';
+import { IMAGE_RESIZER_URL } from '../../lib';
 
 export interface ChangedYourMindDrawerProps {
   visible: boolean;
@@ -57,6 +58,7 @@ export function ChangedYourMindDrawer({
         >
           <CloudImage
             imageUrl={walletErrorRedUrl}
+            imageResizeServiceUrl={IMAGE_RESIZER_URL[checkout.config.environment]}
             sx={{ paddingTop: 'base.spacing.x4', paddingBottom: 'base.spacing.x9' }}
           />
           <ButtCon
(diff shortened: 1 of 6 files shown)


packages/checkout/widgets-lib/src/components/ChangedYourMindDrawer/ChangedYourMindDrawer.tsx
```diff
@@ -57,6 +58,7 @@ export function ChangedYourMindDrawer({
         >
           <CloudImage
             imageUrl={walletErrorRedUrl}
+            imageResizeServiceUrl={IMAGE_RESIZER_URL[checkout.config.environment]}
```
keithbro-imx, line 61: Should we make this property mandatory so that people remember to fill it in? Is there any scenario where we shouldn't be providing this? @glomotion

keithbro-imx: Also minor, but would be nice if you could import the Url already determined instead of having to pick the right one out of the IMAGE_RESIZER_URL object in every component that wants to use it.

keithbro-imx (approved):

packages/checkout/widgets-lib/src/components/ChangedYourMindDrawer/ChangedYourMindDrawer.tsx
dreamoftrees: Ya I was being consistent with our use of the `constants.ts` file..  one of the differences here is that the environment is runtime config during the SDK setup as opposed to ENV variables being set at build time.

dreamoftrees: Could potentially have a bag of this variables in the checkout config.. will give it some thought.

keithbro-imx: Cool neither of these are blockers, can be pondered async from the PR

jwhardwick (approved):

PR #1689, opened 2024-04-22
Feat/gpr 593 smart checkout router callbacks

mimi-imtbl:

# Summary

## Security


packages/checkout/sdk/src/smartCheckout/routing/routingCalculator.ts
```diff
@@ -188,11 +195,22 @@ export const getOnRampFundingStep = async (
   return onRampFundingStep;
 };
 
+const handleRoutePromise = (
```
dreamoftrees, line 198: Only thing Id recommend is maybe tweaking the name to be more contextual to the next/progress callback.. like `handleRouteProgress` or `routeProgress` etc

dreamoftrees (approved):
Great work @mimi-imtbl really elegant way to wrap those promises.

PR #1702, opened 2024-04-26
[feat] gpr 536 smart checkout override balance requirement on sufficient

mimi-imtbl:

# Summary

Add `overrideSufficientBalanceCheckResult` to get funding routes when `sufficient: true` and `onComplete` callback is defined.

### DIFF
--- a/packages/checkout/sdk/src/smartCheckout/smartCheckout.test.ts
+++ b/packages/checkout/sdk/src/smartCheckout/smartCheckout.test.ts
@@ -1,6 +1,9 @@
 import { BigNumber } from 'ethers';
 import { Web3Provider } from '@ethersproject/providers';
-import { smartCheckout } from './smartCheckout';
+import {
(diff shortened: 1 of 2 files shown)


jhesgodi (changes requested):
~Thanks @mimi-imtbl  upon checking I have a concern about the current approach.~

~The override strategy intends to force an "insufficient" result, to trick the routing calculator logic to generate funding routes.~

~This will mislead users relying on the actual sufficient SMC result.~

~Instead, we should decouple funding routes from the balance check, allowing them to be independent results irrespective of the initial balance status.~

~This shift ensures SMC can now provide accurate balance check, and all alternative possible paths.~

jhesgodi (dismissed):
👍🏾

jhesgodi (approved):

dreamoftrees (approved):
👍 Nice work @mimi-imtbl

PR #1706, opened 2024-04-29
feat: [GPR-544][Sale Widget] Remove ability to override base currency

jhesgodi:
# Summary
Remove ability to override base currency. Base currency should only be determined by the config response, as price discrepancies can occurred when missmatched.


mimi-imtbl (approved):

dreamoftrees (approved):
👍

PR #1707, opened 2024-04-29
feat: [GPR-544][Sale Widget]  Add optional token address on sign method

jhesgodi:
# Summary
Can provide token address to sign when invoking `sign` method


mimi-imtbl (approved):

dreamoftrees (approved):
👍

PR #1708, opened 2024-04-29
feat: [GPR-544][Sale Widget] Utility to get top up view data from requirements

jhesgodi:
# Summary
Add utility to get top up view attributes for an insufficient transaction requirement

### DIFF
--- a/packages/checkout/widgets-lib/src/widgets/sale/functions/getTopUpViewData.ts
+++ b/packages/checkout/widgets-lib/src/widgets/sale/functions/getTopUpViewData.ts
@@ -1,0 +1,55 @@
+import { ItemType, TransactionRequirement } from '@imtbl/checkout-sdk';
+
+const tokenInfo = (req?: TransactionRequirement) => {
+  if (req && req.current.type !== ItemType.ERC721) {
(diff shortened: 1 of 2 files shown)


mimi-imtbl (approved):

packages/checkout/widgets-lib/src/widgets/sale/functions/getTopUpViewData.ts
```diff
@@ -0,0 +1,55 @@
...
+    ({ type }) => type === ItemType.ERC20,
+  );
+
+  const balances = {
+    erc20: {
+      value: erc20?.delta.formattedBalance,
+      symbol: tokenInfo(erc20)?.symbol,
+    },
+    native: {
+      value: native?.delta.formattedBalance,
```
dreamoftrees, line 26: Is it worth defaulting this value to zero if native not found in transactionRequirements? .. or doesn't matter so much as symbol won't be defined anyway?
```suggestion
      value: native?.delta.formattedBalance ?? 0,
```

dreamoftrees (approved):
Just small query.. no blocker. 👍

PR #1719, opened 2024-05-01
[feat][sale widget] Move CC warning drawer to the root

mimi-imtbl:

## Changed

Move the CC warning component to the root so it shows up globally.

### DIFF
--- a/packages/checkout/widgets-lib/src/widgets/sale/SaleWidget.tsx
+++ b/packages/checkout/widgets-lib/src/widgets/sale/SaleWidget.tsx
@@ -5,6 +5,7 @@ import {
   useMemo,
   useReducer,
   useRef,
+  useState,
 } from 'react';
 
 import {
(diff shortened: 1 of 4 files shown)


packages/checkout/widgets-lib/src/widgets/sale/SaleWidget.tsx
```diff
@@ -183,6 +187,10 @@ export default function SaleWidget(props: SaleWidgetProps) {
               subheading={viewState.view.data?.subheading}
             />
           )}
+          <CreditCardWarningDrawer
+            visible={showCreditCardWarning}
```
dreamoftrees, line 191: I was wondering how does this get triggered now? After the move I can't see how it links up from the payment methods 🤔

dreamoftrees (commented):
@mimi-imtbl just a little unsure after this change under which conditions the credit card warning is shown?

packages/checkout/widgets-lib/src/widgets/sale/SaleWidget.tsx
mimi-imtbl: accidentally removed in the latest `Clean up` commit, reverted it back

packages/checkout/widgets-lib/src/widgets/sale/components/CreditCardWarningDrawer.tsx
```diff
@@ -4,19 +4,24 @@ import {
...
 export function CreditCardWarningDrawer({
   visible,
   setShowCreditCardWarning,
-  setPaymentMethod,
 }: CreditCardWarningDrawerProps) {
   const { t } = useTranslation();
+  const { setPaymentMethod } = useSaleContext();
+
+  const handleCtaButtonClick = () => {
+    setShowCreditCardWarning(false);
```
jhesgodi, line 22: shouldn't be `true`?

jhesgodi (approved):

dreamoftrees (approved):
👍 Nice one @mimi-imtbl

PR #1722, opened 2024-05-01
feat: ID-1590 Updates AuthManager Logout to use OIDC methods

oliyoung:
Updates logout functions in AuthManager to use `/oidc/` endpoint (documented https://auth0.com/docs/authenticate/login/logout/log-users-out-of-auth0) to work around looming changes to third-party cookie restrictions

Changes logout method from retrieving user session via cookie to supplying the user session to the logout endpoint

## Changed
* New method to build logout arguments, and supplying those to the logout functions

## Fixed
* Rewrites URL construction to use URI objects rather than compounding strings.

### DIFF
--- a/packages/passport/sdk/src/authManager.ts
+++ b/packages/passport/sdk/src/authManager.ts
@@ -38,7 +38,7 @@ const formUrlEncodedHeader = {
   },
 };
 
-const logoutEndpoint = '/oidc';
+const logoutEndpoint = '/oidc/logout';
 const authorizeEndpoint = '/?authorize';
 
 const getAuthConfiguration = (config: PassportConfiguration): UserManagerSettings => {
@@ -360,11 +360,8 @@ export default class AuthManager {
 
   public async getLogoutArgs(): Promise<SignoutRedirectArgs> {
     const user = await this.getUser();
-    const { oidcConfiguration } = this.config;
-
     return {
       id_token_hint: user?.idToken,
-      post_logout_redirect_uri: oidcConfiguration.logoutRedirectUri,
     };
   }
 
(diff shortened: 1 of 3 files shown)


packages/passport/sdk/src/authManager.ts
```diff
@@ -37,15 +38,19 @@ const formUrlEncodedHeader = {
   },
 };
 
+const logoutEndpoint = '/oidc/logout';
+const authorizeEndpoint = '/?authorize';
```
haydenfowler, line 42:
```suggestion
const authorizeEndpoint = '/authorize';
```

packages/passport/sdk/src/authManager.ts
```diff
@@ -37,15 +38,19 @@ const formUrlEncodedHeader = {
...
+
 const getAuthConfiguration = (config: PassportConfiguration): UserManagerSettings => {
   const { authenticationDomain, oidcConfiguration } = config;
 
   const store = typeof window !== 'undefined' ? window.localStorage : new InMemoryWebStorage();
   const userStore = new WebStorageStateStore({ store });
 
-  let endSessionEndpoint = `${authenticationDomain}/v2/logout?client_id=${oidcConfiguration.clientId}`;
+  const endSessionEndpoint = new URL(logoutEndpoint, authenticationDomain);
+  endSessionEndpoint.searchParams.append('client_id', oidcConfiguration.clientId);
```
haydenfowler, line 51: This is a nitpick, but we can use `set` instead of `append`? If a particular query string parameter has already been set and `append` is used, we'll end up with two values, e.g:
```
const myUrl = new URL('/hello/world', 'https://example.com');
myUrl.searchParams.append('queryOne', 'someValue');
myUrl.searchParams.append('queryOne', 'someSecondValue');
myUrl.toString();
// 'https://example.com/hello/world?queryOne=someValue&queryOne=someSecondValue'
```

haydenfowler:
Tested locally with redirect & silent based logouts, everything seems to work as expected 👍

carmen0208 (approved):

haydenfowler (approved):

PR #1735, opened 2024-05-06
refactor:[SMR-2722] calculate gas fee moved

Benjimmutable:
# Summary
Small refactor to move the `calculateGasFee` functions and create a `gas.ts` file.

## Changed
- Renamed `gasPriceInWei.ts` to `gas.ts`, to encompass other gas-related utils.
- Moved `calculateGasFee` into `gas.ts`.
- Added extra value check to `doesChainSupportEIP1559` since all three EIP1559-related fields should be non-null.
- This function was always mocked out in the pre-existing tests, so I added 3 more tests for it.


packages/internal/bridge/sdk/src/lib/gas.ts
```diff
@@ -1,7 +1,8 @@
 import { FeeData } from '@ethersproject/providers';
 import { BigNumber } from 'ethers';
 
-const doesChainSupportEIP1559 = (feeData: FeeData) => !!feeData.maxFeePerGas && !!feeData.maxPriorityFeePerGas;
+const doesChainSupportEIP1559 = (feeData: FeeData) => !!feeData.maxFeePerGas
+    && !!feeData.maxPriorityFeePerGas && !!feeData.lastBaseFeePerGas;
```
Benjimmutable, line 5: A sanity check on this would be great.

wcgcyx (approved):

ermyas (approved):

PR #1740, opened 2024-05-07
feat: Add ERC1155 support to orderbook and checkout

Sam-Jeston:
# Summary
Add ERC1155 support to the orderbook and checkout buy and sell flows.

## Added 
 - Support for ERC1155 in orderbook
 - Support for ERC1155 items in Buy and Sell from checkout


packages/orderbook/src/orderbook.ts
```diff
@@ -246,7 +249,7 @@ export class Orderbook {
       );
     }
 
-    return this.seaport.fulfillOrder(orderResult, takerAddress, extraData);
+    return this.seaport.fulfillOrder(orderResult, takerAddress, extraData, amountToFill);
   }
 
   async fulfillBulkOrders(
```
naveen-imtb, line 255: @Sam-Jeston Shouldn't we update the `fulfillBulkOrders` function also to support partial fulfilment ?

packages/orderbook/src/orderbook.ts
```diff
@@ -209,12 +209,15 @@ export class Orderbook {
    * @param {string} listingId - The listingId to fulfil.
    * @param {string} takerAddress - The address of the account fulfilling the order.
    * @param {FeeValue[]} takerFees - Taker ecosystem fees to be paid.
+   * @param {string} amountToFill - Amount of the order to fill, defaults to sell item amount.
```
naveen-imtb, line 212: @Sam-Jeston Isn't saying `defaults to sell item amount` problematic? Because we don't have the original sell amount here. Should we instead say - `if this parameter is not provided, the entire order will be filled by default` ?

naveen-imtb (changes requested):

packages/orderbook/src/orderbook.ts
Sam-Jeston: indeed we should!

naveen-imtb:
We need to update the sell token validation in [api-client.ts](https://github.com/immutable/ts-immutable-sdk/blob/main/packages/orderbook/src/api-client/api-client.ts#L101)  as it currently throws `Only ERC721 tokens can be listed` for ERC1155 orders.

PR #1747, opened 2024-05-08
fix:[CM-527][sales widget] Top up view with slippage

jiyounglee:
# Summary
Apply slippage on Top up view


jhesgodi (approved):

jhesgodi (changes requested):
@jiyounglee lets put this one on hold, it is possible the swap widget will be already adding slippage already, this changes may cause it to 2x

PR #1749, opened 2024-05-08
feat: [CM-607][Sale Widget] Get order config from product skus + qty

jhesgodi:
# Summary
Improve how primary sales widget gets the quote and configurations for an order. Instead of the widget knowing which `amount` needs to be paid, it can now instead pass the items `sku` & `qty` to the `/client-config` service, and it will be able to return a response with by item amounts and total amounts including conversions.

**BREAKING CHANGE:**
Requires customers to implement new [Quotes endpoint](https://docs.immutable.com/docs/zkEVM/products/checkout/widgets/primary-sales/backend/byo#quote-endpoint)

# Screenshots
**Before:**
Screenshot 2024-05-09 at 11 18 29 AM

**After: Multi item with correct pricing conversion**
Screenshot 2024-05-08 at 1 39 15 PM

## Added 
- Able to list multiple products multiple currency
- Able to quote a price by having the items sku and qty

## Changed
- No longer required to be initialised with the `amount` param

## Removed
- `amount` param

## Fixed
- Item pricing is display correct in the settlement currency selected by customer

### DIFF
--- a/packages/checkout/widgets-lib/src/widgets/sale/hooks/useSmartCheckout.ts
+++ b/packages/checkout/widgets-lib/src/widgets/sale/hooks/useSmartCheckout.ts
@@ -15,7 +15,6 @@ type UseSmartCheckoutInput = {
   checkout: Checkout | undefined;
   provider: Web3Provider | undefined;
   items: SaleItem[];
-  amount: string;
   tokenAddress: string;
 };
 
@@ -23,9 +22,10 @@ export const useSmartCheckout = ({
   checkout,
   provider,
   items,
-  amount,
+  
   tokenAddress,
 }: UseSmartCheckoutInput) => {
+  const amount = '0';
   const [smartCheckoutResult, setSmartCheckoutResult] = useState<
   SmartCheckoutResult | undefined
   >(undefined);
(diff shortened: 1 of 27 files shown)


packages/checkout/widgets-lib/src/widgets/sale/hooks/useSmartCheckout.ts
```diff
@@ -15,17 +15,17 @@ type UseSmartCheckoutInput = {
...
 
 export const useSmartCheckout = ({
   checkout,
   provider,
   items,
-  amount,
+  
   tokenAddress,
 }: UseSmartCheckoutInput) => {
+  const amount = '0';
```
jhesgodi, line 28: Please ignore this patch
The entire hook gets remove in next PR as is not longer required for multi currency flow.

jwhardwick (dismissed):
Looks good 🔥

luads (approved):

dreamoftrees (approved):
Great work @jhesgodi

mimi-imtbl (approved):

jwhardwick (dismissed):

luads (approved):

luads (approved):

mimi-imtbl (approved):

dreamoftrees (approved):
👍

jwhardwick (approved):

PR #1753, opened 2024-05-09
feat: [CM-632][Sale Widget] Add configuration flat to enable/disable transaction finality confirmation

mimi-imtbl:

## Added 

Added `waitFulfillmentSettlements` to sale widget config to enable/disable wait for transaction settlement. `true` by default.

### DIFF
--- a/packages/checkout/widgets-lib/src/widgets/sale/SaleWidgetRoot.tsx
+++ b/packages/checkout/widgets-lib/src/widgets/sale/SaleWidgetRoot.tsx
@@ -132,6 +132,7 @@ export class Sale extends Base<WidgetType.SALE> {
                   excludePaymentTypes={this.parameters.excludePaymentTypes!}
                   language="en"
                   multicurrency={!!this.properties?.config?.multicurrency}
+                  waitForFulfillmentSettlements={!!this.properties?.config?.waitFulfillmentSettlements}
                 />
               </Suspense>
             </ConnectLoader>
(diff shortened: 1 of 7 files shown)


jwhardwick (dismissed):

packages/checkout/widgets-lib/src/widgets/sale/SaleWidgetRoot.tsx
```diff
@@ -132,6 +138,9 @@ export class Sale extends Base<WidgetType.SALE> {
                   excludePaymentTypes={this.parameters.excludePaymentTypes!}
                   language="en"
                   multicurrency={!!this.properties?.config?.multicurrency}
+                  waitFulfillmentSettlements={
```
jhesgodi, line 141: Question and suggestion?
I think we may want the wait to be enabled by default, right? If so then I suggest you use
```
waitFulfillmentSettlements={this.properties?.config?.waitFulfillmentSettlements ?? true}
```
this expresion will allow`false`, but if null or undefined will default to `true`

packages/checkout/widgets-lib/src/widgets/sale/context/SaleContextProvider.tsx
```diff
@@ -136,6 +136,7 @@ const SaleContext = createContext<SaleContextValues>({
   excludePaymentTypes: [],
   multicurrency: false,
   selectedCurrency: undefined,
+  waitFulfillmentSettlements: false,
```
jhesgodi, line 139: default to `true`

jhesgodi (changes requested):
Looks good, left comment re initial state

packages/checkout/widgets-lib/src/widgets/sale/context/SaleContextProvider.tsx
mimi-imtbl: good catch @J

jwhardwick (approved):

jhesgodi (approved):

PR #1772, opened 2024-05-13
feat:[CM-638][Sale Widget] Display Fees on Multi currency flow

jhesgodi:
# Summary
- Simplify fee copy, remove amounts to be paid copy
- Change items list size based on number of funding currencies available
- Display fees breakdown with fiat conversions
- Properly display Fess for multiple items or single items orders

### Screenshot
**Price & Fees on Coins select**
Screenshot 2024-05-15 at 5 20 08 PM
**Fees display for multi item**
Screenshot 2024-05-15 at 5 17 45 PM
**Fees display for single item**
Screenshot 2024-05-15 at 5 20 45 PM
**Fee breakdown**
Screenshot 2024-05-15 at 5 19 18 PM


dreamoftrees (dismissed):
Nice 👍 Just checking should the ~ also be ≈ or different meaning in this context?

jhesgodi:
> Nice 👍 Just checking should the ~ also be ≈ or different meaning in this context?

Nice one thanks.
From my understanding, the symbols are defined as:
- `~` means approximately amount
- `≈` means approximately equal to amount

hence, we should think on using them like this:

*** To display aprox. amounts without conversion:**
`Swap fee ~1 IMX`
read this as: The fee amount will be approximately 1 IMX.

*** To display an fixed amount to an aprox. amount after conversion**
```
Swap fee ~1 IMX
                 ≈ 2.01 USD    
```
read this as: The approximate 1 IMX fee amount is approximately equal to `2.01` USD conversion

jhesgodi:
Spliting into multiple PRs for faster review. Closing for now.

PR #1774, opened 2024-05-13
refactor:[SMR-2729] refactor getTxData

Benjimmutable:
# Summary
Refactor `getTxData` in bridge SDK, and other refactors.

## Added 
- `lib/contracts/createContract.ts` which has a `createContract` function to wrap all `ethers.Contract` creation nicely.
    - Use this function whenever an `ethers.Contract` gets created throughout the sdk.
- Add `isDeposit`, `isWithdraw`, `isWithdrawNotWrappedIMX`, `isWithdrawWrappedIMX`, `isValidDeposit`, and `isValidWithdraw` into `lib/utils.ts`.
    - Check the function natspecs for their respective explanations.
    - The purpose of these functions was to make it clear what the `if (...` statements were actually checking for. If you see more unclear if statements please call them out.

## Changed
- Refactored the `TokenBridge.getBridgeContract` function.
    - This did much more than just "get the bridge contract". It got the bridge contract and 3 other values.
    - It turns out we only need the bridge contract, and `contractMethods`.
    - I deleted the other two and separated the rest into two functions: `getBridgeContract` and `getBridgeMethods`
    - Cleaned up the logic a bit
    - More documentation


packages/internal/bridge/sdk/src/contracts/ABIs/createContract.ts
```diff
@@ -0,0 +1,11 @@
+import { Provider } from '@ethersproject/providers';
+import { BridgeErrorType, withBridgeError } from 'errors';
+import { Contract } from 'ethers';
+
+export async function createContract(address: string, iface: any, provider: Provider) {
+  return await withBridgeError<Contract>(
+    async () => new Contract(address, iface, provider),
+    BridgeErrorType.PROVIDER_ERROR,
+    // TODO what kind of error should this be? Should we just make a new "CONTRACT_CREATION_ERROR"?
+  );
```
Benjimmutable, line 10: opinions?

packages/internal/bridge/sdk/src/contracts/createContract.ts
```diff
@@ -0,0 +1,11 @@
+import { Provider } from '@ethersproject/providers';
+import { BridgeErrorType, withBridgeError } from 'errors';
+import { Contract } from 'ethers';
+
+export async function createContract(address: string, iface: any, provider: Provider) {
+  return await withBridgeError<Contract>(
+    async () => new Contract(address, iface, provider),
+    BridgeErrorType.PROVIDER_ERROR,
+    // TODO what kind of error should this be? Should we just make a new "CONTRACT_CREATION_ERROR"?
```
Benjimmutable, line 9: Opinions?

packages/internal/bridge/sdk/src/lib/utils.ts
```diff
@@ -72,6 +72,52 @@ export function isWrappedIMX(token: FungibleToken, source: string) {
...
+): boolean {
+  return isWithdraw(sourceChainId, bridgeInstance) && !isRootIMX(token, sourceChainId);
+}
+
+export function isWithdrawWrappedIMX(
+  token: FungibleToken,
+  sourceChainId: string,
+  bridgeInstance: BridgeInstance,
+): boolean {
+  return isWithdraw(sourceChainId, bridgeInstance) && isRootIMX(token, sourceChainId);
```
ermyas, line 102: can i confirm that checking `isRootIMX()` against `token` here correct? I would have assumed `token` here would be the address on L2, while the `isRootIMX()` fn compares against address on L1?

packages/internal/bridge/sdk/src/lib/utils.ts
```diff
@@ -72,6 +72,52 @@ export function isWrappedIMX(token: FungibleToken, source: string) {
...
+  return sourceChainId === bridgeInstance.rootChainID && destinationChainId === bridgeInstance.childChainID;
+}
+
+/**
+ * This is the same as `isWithdraw`, but takes an extra parameter - the `destinationChainId` - to check that both
+ * of the chain IDs are correct.
+ */
+export function isValidWithdraw(sourceChainId: string, destinationChainId: string, bridgeInstance: BridgeInstance) {
+  return sourceChainId === bridgeInstance.childChainID && destinationChainId === bridgeInstance.rootChainID;
+}
```
ermyas, line 119: is there a reason we wouldn't replace `isDeposit()` and `isWithdrawal()` with these? The additional verification of the destination chain is beneficial for integrity as well.

packages/internal/bridge/sdk/src/lib/utils.ts
```diff
@@ -72,6 +72,52 @@ export function isWrappedIMX(token: FungibleToken, source: string) {
   return token.toUpperCase() === getWrappedIMX(source).toUpperCase();
 }
 
+/**
+ * Assumes the destination chain is correct (i.e. it doesn't check the destination chain)
+ */
+export function isDeposit(sourceChainId: string, bridgeInstance: BridgeInstance): boolean {
```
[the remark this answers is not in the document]
Benjimmutable: Yeah good catch. I added a check to `req.action` here as well. We could also add a check against the `destinationChain` too. Not sure what the nicest way to go about this is.

packages/internal/bridge/sdk/src/lib/utils.ts
Benjimmutable: Yeah I tend to agree. I made the current changes because I was compartmentalising the code that currently existed in `tokenBridge.ts`. But thinking about it more, we should probably not do that here.

packages/internal/bridge/sdk/src/lib/utils.test.ts
```diff
@@ -24,4 +33,233 @@ describe('utils', () => {
       expect(result).toEqual('devnet');
     });
   });
+
+  describe('isWithdrawNotWrappedIMX', () => {
+    it('should return false for non-wrapped IMX but not withdraw', () => {
```
ermyas, line 38: nit: can we use a diff term for "non-wrapped"/"not-wrapped" 😅 ... this is def subjective, but i find it confusing ... maybe "native"?

Benjimmutable: Yeah lol good point

packages/internal/bridge/sdk/src/lib/utils.test.ts
```diff
@@ -24,4 +33,233 @@ describe('utils', () => {
...
+
+      const result = isWithdrawNotWrappedIMX(token, direction, bridgeInstance);
+      expect(result).toEqual(false);
+    });
+
+    it('should return true for not wrapped IMX', () => {
+      const token: FungibleToken = rootIMXs.mainnet; // Anything that isn't a child chain WIMX.
+      const bridgeInstance: BridgeInstance = {
+        rootChainID: '123',
+        childChainID: ETH_SEPOLIA_CHAIN_ID,
```
ermyas, line 76: Should the child chain ids here and elsewhere be `ZKEVM_TESTNET_CHAIN_ID`? 
```suggestion
        childChainID: ZKEVM_TESTNET_CHAIN_ID,
```

Benjimmutable: Given that these are just unit tests, it shouldn't make a difference how we configure the `BridgeInstance`

packages/internal/bridge/sdk/src/lib/utils.ts
```diff
@@ -72,6 +75,52 @@ export function isWrappedIMX(token: FungibleToken, source: string) {
   return token.toUpperCase() === getWrappedIMX(source).toUpperCase();
 }
 
+/**
+ * This is the same as `isDeposit`, but takes an extra parameter - the `destinationChainId` - to check that both
```
ermyas, line 79: nit: should we update comment, given there is no `isDeposit` fn anymore?

packages/internal/bridge/sdk/src/lib/utils.ts
```diff
@@ -72,6 +75,52 @@ export function isWrappedIMX(token: FungibleToken, source: string) {
...
+ *          indicating that this should be a deposit or withdraw finalisation, without doing validation.
+ * Useful for if we believe a transaction should be a deposit or withdraw finalisation,
+ * so you can construct relevant data types to then do the checks
+ */
+export function shouldBeDepositOrFinaliseWithdraw(sourceChainId: string, bridgeInstance: BridgeInstance) {
+  return sourceChainId === bridgeInstance.rootChainID;
+}
+
+/**
+ * This is the same as `isWithdraw`, but takes an extra parameter - the `destinationChainId` - to check that both
```
ermyas, line 99: nit: should we update comment, given there is no `isWithdraw` fn anymore?

packages/internal/bridge/sdk/src/lib/utils.test.ts
ermyas: Agreed. It just avoids possible confusion?

packages/internal/bridge/sdk/src/tokenBridge.ts
```diff
@@ -169,41 +181,79 @@ export class TokenBridge {
...
     }
 
     // Get source fee & bridge fee
+    let axelarGasLimit;
+    if (req.action === BridgeFeeActions.DEPOSIT) {
+      sourceChainFee = calculateGasFee(feeData, BridgeMethodsGasLimit.DEPOSIT_SOURCE);
+      axelarGasLimit = BridgeMethodsGasLimit.DEPOSIT_DESTINATION;
+    } else {
+      sourceChainFee = calculateGasFee(feeData, BridgeMethodsGasLimit.WITHDRAW_SOURCE);
+      axelarGasLimit = BridgeMethodsGasLimit.WITHDRAW_DESTINATION;
```
ermyas, line 226: I would have assumed we'd be using dynamic gas estimates on these?

packages/internal/bridge/sdk/src/tokenBridge.ts
```diff
@@ -551,6 +618,7 @@ export class TokenBridge {
     }
     // Withdraw request
```
wcgcyx, line 619: should we check if it is a valid withdraw?

packages/internal/bridge/sdk/src/lib/utils.test.ts
```diff
@@ -24,4 +33,233 @@ describe('utils', () => {
...
+  describe('isWithdrawNotWrappedIMX', () => {
+    it('should return false for non-wrapped IMX but not withdraw', () => {
+      const token: FungibleToken = rootIMXs.mainnet;
+      const bridgeInstance: BridgeInstance = {
+        rootChainID: '123',
+        childChainID: ETH_SEPOLIA_CHAIN_ID,
+      };
+
+      const direction: BridgeDirection = {
+        sourceChainId: bridgeInstance.childChainID,
```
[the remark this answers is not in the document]
Benjimmutable: You are correct. In this case, we are unit testing `isWithdrawNativeIMX` to make sure that it returns `false` if `action` is `DEPOSIT`

packages/internal/bridge/sdk/src/tokenBridge.ts
wcgcyx: Looks like `getDepositOrWithdrawFee` is used by `getFeePrivate` (the old interface), where we do the static estimation. Fees obtained with `getFeePrivate` are used for illustration purposes only so does not involve transaction building.

packages/internal/bridge/sdk/src/lib/validation.ts
```diff
@@ -153,12 +156,31 @@ export async function validateBridgeReqArgs(
...
+  };
+
+  // TODO do these functions get tested?
   if (
-    req.action === BridgeFeeActions.DEPOSIT
-    && (
-      req.sourceChainId !== config.bridgeInstance.rootChainID
-    || req.destinationChainId !== config.bridgeInstance.childChainID
-    )
+    direction.action === BridgeFeeActions.DEPOSIT
```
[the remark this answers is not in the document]
Benjimmutable: I suppose not. I think this points to a wider issue that this validation function should just check against both `isValidDeposit` and `isValidWithdraw`.

packages/internal/bridge/sdk/src/tokenBridge.ts
Benjimmutable: Yep good point. It could be neither a valid deposit or a valid withdrawal. In which case we'd want to throw an error I imagine

ermyas (approved):

wcgcyx (approved):

PR #1778, opened 2024-05-14
feat: [ID-1585] get signers after confirmation popup

scastleman-immutable:
# Summary
[Jira ticket](https://immutable.atlassian.net/browse/ID-1585)
Moved initialisation of starkex signers after the confirmation popup is initialised, in order to avoid possible popup blocking issues

### DIFF
--- a/packages/passport/sdk-sample-app/src/context/ImmutableProvider.tsx
+++ b/packages/passport/sdk-sample-app/src/context/ImmutableProvider.tsx
@@ -126,7 +126,7 @@ const getPassportConfig = (environment: EnvironmentNames): PassportModuleConfigu
           authenticationDomain: 'https://auth.dev.immutable.com',
           magicPublishableApiKey: 'pk_live_4058236363130CA9',
           magicProviderId: 'C9odf7hU4EQ5EufcfgYfcBaT5V6LhocXyiPRhIjw2EY=',
-          passportDomain: 'https://passport.dev.immutable.com',
+          passportDomain: 'passport.dev.immutable.com',
           imxPublicApiDomain: 'https://api.dev.immutable.com',
           imxApiClients: new ImxApiClients(createConfig({
             basePath: 'https://api.dev.immutable.com',
--- a/packages/passport/sdk/src/starkEx/workflows/workflows.test.ts
+++ b/packages/passport/sdk/src/starkEx/workflows/workflows.test.ts
@@ -39,7 +39,7 @@ describe('passportImxProvider auth tests', () => {
 
   const mockGuardianClient = {
     withDefaultConfirmationScreenTask: (task: () => any) => task,
-    withConfirmationScreenTask: (task: () => any) => (() => task()),
+    withConfirmationScreenTask: () => (task: () => any) => task,
   };
 
   const passportImxProvider = new PassportImxProvider({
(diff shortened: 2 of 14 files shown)


packages/passport/sdk/src/starkEx/workflows/workflows.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    it(`should return an error for ${methodName}`, async () => {
+      expect(async () => await passportImxProvider[methodName!](args))
+        .rejects
+        .toThrow(
+          new PassportError(
+            'User has been logged out',
+            PassportErrorType.NOT_LOGGED_IN_ERROR,
+          ),
+        );
+    });
```
imx-mikhala, line 77: When asserting errors with custom types, like here with `PassportErrorType.NOT_LOGGED_IN_ERROR` jest doesn't actually assert this type is what we expect and only validates the message is the same. You could pass in any `PassportErrorType` here and the test will pass.

E.g. the following will pass the test

```
expect(async () => await passportImxProvider[methodName!](args))
        .rejects
        .toThrow(
          new PassportError(
            'User has been logged out',
            'Some type' as PassportErrorType,
          ),
        );
```

If you want to assert the type the only way I've found works is using a catch block to assert the type is what is expected. Note that we need to assert outside of the catch, otherwise we cant be sure that the catch was evaluated as the test will pass if the function doesn't throw.

```
let message = '';
let type = '';
try {
        await passportImxProvider[methodName!](args);
} catch (e: any) {
        message = e.message;
        type = e.type;
}

expect(message).toEqual('User has been logged out');
expect(type).toEqual(PassportErrorType.NOT_LOGGED_IN_ERROR);
```

scastleman-immutable: Thanks for the callout. I just moved this test over from `passportImxProvider.test.ts`, but probably a good opportunity to fix it up while we're in the area.

packages/passport/sdk/src/starkEx/workflows/workflows.test.ts
```diff
@@ -0,0 +1,115 @@
...
+      passportEventEmitter.emit(PassportEvents.LOGGED_OUT);
+    });
+
+    it(`should return an error for ${methodName}`, async () => {
+      try {
+        await passportImxProvider[methodName!](args);
+      } catch (err: any) {
+        const { message, type } = err;
+        expect(message).toEqual('User has been logged out');
+        expect(type).toEqual(WorkflowErrorMap[methodName]);
```
imx-mikhala, line 86: Can we move the expect outside of the catch to ensure we are actually falling into the catch block? Otherwise if the function isnt actually throwing this test still passes. See my other comment for how to set this up.

oliyoung (approved):

packages/passport/sdk-sample-app/src/context/ImmutableProvider.tsx
```diff
@@ -126,7 +126,7 @@ const getPassportConfig = (environment: EnvironmentNames): PassportModuleConfigu
           authenticationDomain: 'https://auth.dev.immutable.com',
           magicPublishableApiKey: 'pk_live_4058236363130CA9',
           magicProviderId: 'C9odf7hU4EQ5EufcfgYfcBaT5V6LhocXyiPRhIjw2EY=',
-          passportDomain: 'https://passport.dev.immutable.com',
```
haydenfowler, line 129: Was this change intentional?

haydenfowler (changes requested):
The change to `ImmutableProvider.tsx` causes issues

image

PR #1781, opened 2024-05-14
refactor:[SMR-2730] rewrite get tx data

Benjimmutable:
# Summary
Rewrite `getTxData` in bridge SDK

If you do not think this is a better way to write this function, please voice this opinion. I personally believe it is easier to read.

## Changed
- Rewrite a bit of `getTxData` so you just set the function name and parameters then generate the calldata based on those, since you already have the correct bridge contract to do it on as well.


packages/internal/bridge/sdk/src/tokenBridge.ts
```diff
@@ -384,39 +384,40 @@ export class TokenBridge {
...
      * Handle bridge transaction for native token
      */
     if (token.toUpperCase() === NATIVE) {
       if (sender === recipient) {
         // Deposit or withdraw native token
-        return await withBridgeError<string>(async () => bridgeContract.interface.encodeFunctionData(
-          currentBridgeMethods.native,
-          [amount],
-        ), BridgeErrorType.INTERNAL_ERROR);
+        functionName = currentBridgeMethods.native;
```
Benjimmutable, line 395: the new way of doing it feels a lot easier to read to me

wcgcyx (approved):

packages/internal/bridge/sdk/src/tokenBridge.ts
```diff
@@ -384,39 +384,40 @@ export class TokenBridge {
     const currentBridgeMethods = await this.getBridgeMethods(direction);
     const bridgeContract = await this.getBridgeContract(direction);
 
+    let functionName: string;
+    let parameters: any[];
     /**
      * Handle bridge transaction for native token
      */
     if (token.toUpperCase() === NATIVE) {
       if (sender === recipient) {
```
ermyas, line 393: should we handle scenario where a checksummed version of an address might evaluate different from a non-checksummed version?

ermyas (approved):

packages/internal/bridge/sdk/src/tokenBridge.ts
Benjimmutable: Yep. Ticket created

PR #1783, opened 2024-05-15
signers now retrieved after confirmation popup is opened

scastleman-immutable:
# Summary
[Jira ticket](https://immutable.atlassian.net/browse/ID-1585)
Moved initialisation of starkex signers after the confirmation popup is initialised, in order to avoid possible popup blocking issues

### DIFF
--- a/packages/passport/sdk/src/starkEx/passportImxProvider.ts
+++ b/packages/passport/sdk/src/starkEx/passportImxProvider.ts
@@ -268,7 +268,6 @@ export class PassportImxProvider implements IMXProvider {
   async batchNftTransfer(
     request: NftTransferDetails[],
   ): Promise<imx.CreateTransferResponse> {
-  // eslint-disable-next-line function-paren-newline
     return this.guardianClient.withConfirmationScreenTask(
       { width: 480, height: 784 },
     )(async () => {
--- a/packages/passport/sdk/src/starkEx/passportImxProvider.ts
+++ b/packages/passport/sdk/src/starkEx/passportImxProvider.ts
@@ -22,7 +22,7 @@ import GuardianClient from '../guardian';
 import {
   PassportEventMap, PassportEvents, UserImx, User, IMXSigners, isUserImx,
 } from '../types';
-import { PassportError, PassportErrorType, withPassportError } from '../errors/passportError';
+import { PassportError, PassportErrorType } from '../errors/passportError';
 import {
   batchNftTransfer, cancelOrder, createOrder, createTrade, exchangeTransfer, transfer,
 } from './workflows';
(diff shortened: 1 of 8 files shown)


packages/passport/sdk/src/starkEx/workflows/order.ts
```diff
@@ -33,66 +32,64 @@ export async function createOrder({
...
-    const amountBuy = request.buy.type === ERC721 ? '1' : request.buy.amount;
-    const headers = { Authorization: `Bearer ${user.accessToken}` };
-
-    const getSignableOrderRequestV3: imx.GetSignableOrderRequest = {
-      user: ethAddress,
-      amount_buy: amountBuy,
-      token_buy: convertToSignableToken(request.buy),
-      amount_sell: amountSell,
-      token_sell: convertToSignableToken(request.sell),
+  const { ethAddress } = user.imx;
```
haydenfowler, line 35: For the sake of consistency, can we leave the calls to `withPassportError` inside of the workflow files please?
Otherwise we end up in a situation where some workflow methods are wrapped in `withPassportError`, and others aren't.

scastleman-immutable: So we want to keep `withDefaultConfirmationScreenTask` outside the workflow file, but the call to `withPassportError` inside it?

scastleman-immutable:
> It looks like you've updated for some of the workflows but not all of them?

The only one I didn't update was `exchangeTransfer`, which doesn't use the guardian confirmation screen as far as I can tell, so it wouldn't have been affected by the bug this ticket was to address

packages/passport/sdk/src/starkEx/passportImxProvider.ts
```diff
@@ -213,55 +217,71 @@ export class PassportImxProvider implements IMXProvider {
...
-    const { user, starkSigner } = await this.#getRegisteredImxUserAndSigners();
-
-    return batchNftTransfer({
-      request,
-      user,
-      starkSigner,
-      transfersApi: this.immutableXClient.transfersApi,
-      guardianClient: this.guardianClient,
-    });
+  // eslint-disable-next-line function-paren-newline
```
haydenfowler, line 271:
```suggestion
```
This doesn't seem necessary

scastleman-immutable: You're right - it was a holdover from when I moved it from the workflow file. I'll remove that

packages/passport/sdk/src/starkEx/workflows/order.ts
```diff
@@ -102,13 +102,13 @@ export async function cancelOrder({
...
 }: CancelOrderParams): Promise<imx.CancelOrderResponse> {
-  return withPassportError<imx.CancelOrderResponse>(guardianClient.withDefaultConfirmationScreenTask(async () => {
+  return withPassportError<imx.CancelOrderResponse>(async () => {
     const getSignableCancelOrderRequest: imx.GetSignableCancelOrderRequest = {
       order_id: request.order_id,
     };
 
     const headers = {
-      // eslint-disable-next-line @typescript-eslint/naming-convention
+    // eslint-disable-next-line @typescript-eslint/naming-convention
```
haydenfowler, line 111:
```suggestion
      // eslint-disable-next-line @typescript-eslint/naming-convention
```

haydenfowler (approved):

PR #1787, opened 2024-05-16
chore: [NO-JIRA] Updates to Checkout SDK sample app to support Passport provider

dreamoftrees:
# Summary
Small update to Checkout SDK sample app to support Passport provider

## Added 
Injected providers options and Passport provider to Checkout SDK sample app

### DIFF
--- a/packages/checkout/sdk-sample-app/src/components/CheckConnection.tsx
+++ b/packages/checkout/sdk-sample-app/src/components/CheckConnection.tsx
@@ -36,6 +36,7 @@ export default function CheckConnection(props: CheckConnectionProps) {
       const resp = await checkout.checkIsWalletConnected({
         provider,
       });
+      console.log('Connection result', resp);
       setResult(resp);
       setLoading(false);
     } catch (err: any) {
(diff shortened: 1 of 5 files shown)


jwhardwick (approved):

packages/checkout/sdk-sample-app/src/components/CheckConnection.tsx
```diff
@@ -36,6 +36,7 @@ export default function CheckConnection(props: CheckConnectionProps) {
       const resp = await checkout.checkIsWalletConnected({
         provider,
       });
+      console.log('Connection result', resp);
```
mimi-imtbl, line 39: do we want to log this? 👀

mimi-imtbl (approved):

packages/checkout/sdk-sample-app/src/components/CheckConnection.tsx
dreamoftrees: I thought it was helpful just because the UI is not really refreshing clearly.. I also figured it was fine as it's really our internal SDK sample app.. but happy to remove if you prefer @mimi-imtbl ?

PR #1797, opened 2024-05-17
fix: ID-1732 Fix background blur and append required font links to the head

imx-mikhala:

# Summary

- Fix issue with background not being blurred on Safari & iPhone
- Append font links to head to import Roboto

## Security

### DIFF
--- a/packages/passport/sdk/src/overlay/overlay.ts
+++ b/packages/passport/sdk/src/overlay/overlay.ts
@@ -15,13 +15,10 @@ export default class Overlay {
 
   private onCloseListener: (() => void) | undefined;
 
-  private addedFonts: boolean;
-
   constructor(popupOverlayOptions: PopupOverlayOptions, isBlockedOverlay: boolean = false) {
     this.disableBlockedPopupOverlay = popupOverlayOptions.disableBlockedPopupOverlay || false;
     this.disableGenericPopupOverlay = popupOverlayOptions.disableGenericPopupOverlay || false;
     this.isBlockedOverlay = isBlockedOverlay;
-    this.addedFonts = false;
   }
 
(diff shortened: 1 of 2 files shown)


packages/passport/sdk/src/overlay/overlay.ts
```diff
@@ -48,6 +48,22 @@ export default class Overlay {
...
+      const link2: HTMLLinkElement = document.createElement('link');
+      link2.href = 'https://fonts.gstatic.com';
+      link2.rel = 'preconnect';
+      link2.crossOrigin = 'anonymous';
+      document.head.appendChild(link2);
+
+      const link3: HTMLLinkElement = document.createElement('link');
+      link3.href = 'https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,400;0,500;0,700;1,400;1,500;1,700&display=swap';
+      link3.rel = 'stylesheet';
+      document.head.appendChild(link3);
```
imx-mikhala, line 65: this is all the links from the [stackblitz](https://stackblitz.com/edit/stackblitz-starters-tcunqf?file=index.html)

oliyoung (dismissed):

haydenfowler (changes requested):
The link elements are added to the head element each time the overlay is displayed.
Also, is a `preconnect` link accurate here if we're only adding the link at the time that the fonts are needed?

imx-mikhala:
> The link elements are added to the head element each time the overlay is displayed. Also, is a `preconnect` link accurate here if we're only adding the link at the time that the fonts are needed?

oh good catch i will patch that. not sure re the `preconnect`, if its not required i can remove

haydenfowler (approved):

yundifu (approved):

PR #1798, opened 2024-05-17
fix: update the checkout.Sell interface to allow undefined type property

Sam-Jeston:
# Summary

A previous change introduced a minor type break. This restores backwards compatibility by introducing a deprecated type that matches the previous type, and ensure the code behaves as expected when that type is provided

## Fixed
Backwards compatible SellItem type

### DIFF
--- a/packages/checkout/sdk/src/smartCheckout/sell/sell.ts
+++ b/packages/checkout/sdk/src/smartCheckout/sell/sell.ts
@@ -7,6 +7,7 @@ import {
   PrepareListingResponse,
   constants,
   ERC721Item as OrderbookERC721Item,
+  ERC1155Item as OrderbookERC1155Item,
 } from '@imtbl/orderbook';
 import { BigNumber, Contract, utils } from 'ethers';
 import {
@@ -132,9 +133,10 @@ export const sell = async (
     const { seaportContractAddress } = orderbook.config();
     spenderAddress = seaportContractAddress;
 
-    const sellItem = 'type' in sellToken && sellToken.type === ItemType.ERC1155
+    const hasType = 'type' in sellToken;
+    const sellItem: OrderbookERC721Item | OrderbookERC1155Item = hasType && sellToken.type === ItemType.ERC1155
       ? {
-        type: sellToken.type,
+        type: ItemType.ERC1155,
         contractAddress: sellToken.collectionAddress,
         tokenId: sellToken.id,
         amount: sellToken.amount,
@@ -143,7 +145,7 @@ export const sell = async (
         type: ItemType.ERC721,
         contractAddress: sellToken.collectionAddress,
         tokenId: sellToken.id,
-      } as OrderbookERC721Item;
+      };
 
     listing = await measureAsyncExecution<PrepareListingResponse>(
       config,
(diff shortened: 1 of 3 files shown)


naveen-imtb (dismissed):

naveen-imtb (dismissed):

packages/checkout/sdk/src/smartCheckout/sell/sell.ts
```diff
@@ -166,9 +168,9 @@ export const sell = async (
   }
 
   const itemRequirements = [
```
jhesgodi, line 170: Thanks for the fix. May I please get improve legibility of this code block. A ternary combined within an array is less clear. we can improve by defaulting a simple `if/else` statement.

```
  const itemRequirements = [];
  if (sellToken?.type === ItemType.ERC1155) {
      itemRequirements.push(getERC1155Requirement(sellToken.id, sellToken.collectionAddress, spenderAddress, sellToken.amount));
  } else {
      getERC721Requirement(sellToken.id, sellToken.collectionAddress, spenderAddress)
  }
```

packages/checkout/sdk/src/smartCheckout/sell/sell.ts
```diff
@@ -130,18 +131,19 @@ export const sell = async (
     orderbook = instance.createOrderbookInstance(config);
     const { seaportContractAddress } = orderbook.config();
     spenderAddress = seaportContractAddress;
-    const sellItem = sellToken.type === ItemType.ERC721
+
+    const sellItem = 'type' in sellToken && sellToken.type === ItemType.ERC1155
```
jhesgodi, line 135: Not a blocker.
This can benefit from a coercion op `sellToken?.type === ItemType.ERC1155`

packages/checkout/sdk/src/smartCheckout/sell/sell.ts
```diff
@@ -130,18 +131,19 @@ export const sell = async (
...
       ? {
         type: sellToken.type,
```
jhesgodi, line 137: given there are 2 distinctive types now, this benefits from explicitly setting the type to `ItemType.ERC1155`

packages/checkout/sdk/src/smartCheckout/sell/sell.ts
```diff
@@ -130,18 +131,19 @@ export const sell = async (
...
+        amount: sellToken.amount,
       }
       : {
-        type: sellToken.type,
+        type: ItemType.ERC721,
         contractAddress: sellToken.collectionAddress,
         tokenId: sellToken.id,
-        amount: sellToken.amount,
-      };
+      } as OrderbookERC721Item;
```
jhesgodi, line 146: Wonder why is necessary to force the type here? Wouldn't `type: ItemType.ERC721` get typescript to detect this as an `ERC721` already?

jhesgodi (changes requested):
Thanks for the improvements @Sam-Jeston , left some minor comments.

packages/checkout/sdk/src/smartCheckout/sell/sell.ts
Sam-Jeston: From what I can tell, the type guard is required because the deprecated ERC721 item type (which this PR restores support for) does not include a type property, so to explicit type guard is required, and we know the ERC1155 variant is guaranteed to have the property
Screenshot 2024-05-20 at 10 25 32 am

packages/checkout/sdk/src/smartCheckout/sell/sell.ts
Sam-Jeston: Yeh I think due to the ternary it thought otherwise `amount` as undefined, which strictly isnt on the orderbook ERC721 item type. I think its cleaner to not cast here and instead strictly type the declared const which I have now done

packages/checkout/sdk/src/smartCheckout/sell/sell.ts
Sam-Jeston: done!

packages/checkout/sdk/src/smartCheckout/sell/sell.ts
Sam-Jeston: done!

jhesgodi (approved):
LGTM 🔥 
Thanks for this fixes.

packages/checkout/sdk/src/smartCheckout/sell/sell.ts
jhesgodi: Makes sense thanks, Missed the deprecated type line

jiyounglee (approved):

PR #1809, opened 2024-05-21
feat:[CM-657] Adds order details to Sales analytics

jwhardwick:

# Summary

## Added 

* Adds Sales order details to analytic events
* Wait until we have this data to track payment methods page view so we can track conversion % by order size

### DIFF
--- a/packages/checkout/widgets-lib/src/widgets/sale/hooks/useSaleEvents.ts
+++ b/packages/checkout/widgets-lib/src/widgets/sale/hooks/useSaleEvents.ts
@@ -1,24 +1,24 @@
-import { useContext } from 'react';
-import { StandardAnalyticsActions } from '@imtbl/react-analytics';
 import { FundingItem, SalePaymentTypes } from '@imtbl/checkout-sdk';
+import { StandardAnalyticsActions } from '@imtbl/react-analytics';
+import { useContext } from 'react';
 import {
+  AnalyticsControlTypes,
   UserJourney,
   useAnalytics,
-  AnalyticsControlTypes,
 } from '../../../context/analytics-provider/SegmentAnalyticsProvider';
 import { EventTargetContext } from '../../../context/event-target-context/EventTargetContext';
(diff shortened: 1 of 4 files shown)


luads (approved):

packages/checkout/widgets-lib/src/widgets/sale/views/PaymentMethods.tsx
```diff
@@ -82,7 +83,12 @@ export function PaymentMethods() {
     }
   }, [paymentMethod]);
 
-  useEffect(() => sendPageView(SaleWidgetViews.PAYMENT_METHODS), []); // checkoutPrimarySalePaymentMethodsViewed
+  useEffect(() => {
+    // We need OrderQuote on this page view, but don't want to slow load of page by waiting for it
+    if (!orderQuote || orderQuote?.totalAmount?.USD?.amount === undefined) return;
```
jhesgodi, line 88: Given that USD is not guarantee to exist, and we don't know yet the settlement coin the user will select.
We may do better by sending a serialised `orderQuote: totalAmount` thru.

In most cases should be just like this, IMO:
`{ pricing: { USDC: x, GOG: x, ETH: x, USD: x }}`

packages/checkout/widgets-lib/src/widgets/sale/hooks/useSaleEvents.ts
```diff
@@ -27,6 +27,7 @@ export const useSaleEvent = () => {
     recipientAddress: userId,
     signResponse,
     paymentMethod,
+    orderQuote,
```
jhesgodi, line 30: orderQuote has too much information is not needed, can we change this to `totalAmount`.

ie: config.contractId should not be sent, as well as `products` is info that gets passed again towards the end on the success event

jhesgodi (changes requested):

jwhardwick:
Parking this one, will consider a different approach.

PR #1814, opened 2024-05-21
fix: [Sale Widget] Price conversion calculation

jhesgodi:
# Summary
Fix calculation for unknown coins prices from USD conversion rate


dreamoftrees (approved):
👍

luads (approved):

PR #1824, opened 2024-05-24
fix: [Sale Widget] Duplicated balances quotes

jhesgodi:
# Summary
Fix issue where balances where quoted for all currencies. Instead, it should only quote funding routes for the base currency.

**Before: Duplicated routes**
Screenshot 2024-05-24 at 4 48 25 PM

**After: Only routes for base currency**
Screenshot 2024-05-24 at 4 48 25 PM


alejoloaiza (approved):

dreamoftrees (approved):
👍

mimi-imtbl (approved):

PR #1836, opened 2024-05-29
DX-2821/update_deposits

kaihirota:
# Summary

Update deposit flows to drop the registration check. This is for StarkEx upgrade to v4.

From StarkEx V4, users are not required to be registered on-chain in order to deposit.


brayansdt (approved):
LGTM

mimi-imtbl (approved):

PR #1838, opened 2024-05-29
chore: ID-1632 Update oidc-client-ts dependency

haydenfowler:
# Summary
This PR bumps `oidc-client-ts` so that we can make use of a feature that was introduced in a later version. According to the [packages releases page](https://github.com/authts/oidc-client-ts/releases?page=1), no breaking changes have been implemented.

I've tested the login & logout flow locally and they appear to be working as intended.

## Changed
- Passport: Bumped `oidc-client-ts` dependency to 2.4.0


carmen0208 (dismissed):

sdk/package.json
CodeSchwert, line 1: @haydenfowler can you run build locally and see if that reorders the dependencies?

haydenfowler: @CodeSchwert after `yarn build`, the dependencies still appear in this order. Want me to revert the re-ordering?

CodeSchwert: Yes, revert the ordering please.

haydenfowler: Fixed @CodeSchwert

CodeSchwert (approved):

sharif9876 (approved):

PR #1843, opened 2024-05-30
fix: [Connect Widget] Wallet connect provider chain id validation

jhesgodi:
# Summary
When connecting with wallet connect the underlying provider chain id was always out of sync. Given the following issues:
1. Ethereum provider returned by wallet connect is not in sync with selected network in wallet. To fix it, one must ensure to request a network switch.
2. Chain id validation was incorrect for wallet connect, as  both`chainId` property and value returned from rpc call was returned as an hex value for desktop wallets, but as a number for wallet connect.

**Misc changes**:
- Add export for `validateProvider` in SDK. this function is handy to check and debug providers within widgets
- Add wallet connect config to `/sale` page in widgets sample app.

## Added 
- Connect widget: After connecting with Wallet connect trigger network switch if chain id mismatch

## Fixed
- Fix `getUnderlyingChainId` method to account for different chain id  as number or hexadecimal

### DIFF
--- a/packages/checkout/sdk/src/index.ts
+++ b/packages/checkout/sdk/src/index.ts
@@ -15,6 +15,7 @@ export { IMMUTABLE_API_BASE_URL } from './env';
 export {
   getPassportProviderDetail,
   getMetaMaskProviderDetail,
+  validateProvider,
 } from './provider';
 
 export {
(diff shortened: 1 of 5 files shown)


dreamoftrees (approved):
Awesome work @jhesgodi so good to snag those issues!

jwhardwick (approved):
Nice ✅

PR #1846, opened 2024-05-31
fix:[IGSGS-7] Widget input handling of decimal places

jwhardwick:

# Summary

Fixes bug in Swap when handling input values with > 6 decimal places.

## Fixed

* when input value is > 6dp, the value is ignored
* this was still triggering loading to be set to true, even though no quote was being fetched
* caused infinite loading spinner
* if the value isn't updated, we skip loading/attempting a refetch

https://immutable.atlassian.net/browse/IGSGS-7
Note - there's another change needed to full resolve this with the Sale Widget, coming in another PR.


packages/checkout/widgets-lib/src/widgets/bridge/components/BridgeForm.tsx
```diff
@@ -188,20 +188,15 @@ export function BridgeForm(props: BridgeFormProps) {
   };
 
   const handleBridgeAmountChange = (value: string) => {
-    // Ensure that starting with a decimal is formatted correctly
-    let inputValue = value;
```
jwhardwick, line 192: This is already happening inside `TextInputForm`.

packages/checkout/widgets-lib/src/widgets/bridge/components/BridgeForm.tsx
```diff
@@ -188,20 +188,15 @@ export function BridgeForm(props: BridgeFormProps) {
...
-    if (inputValue === '.') {
-      inputValue = '0.';
-    }
-    setFormAmount(inputValue);
```
jwhardwick, line 196: This is already happening inside `TextInputForm`, removing it here.

dreamoftrees (approved):
👍 Nice one

mimi-imtbl (approved):

jhesgodi (approved):

PR #1855, opened 2024-06-03
update

kaihirota:

# Summary

## Security


packages/x-provider/src/signable-actions/withdrawal-actions/completeERC20Withdrawal.ts
```diff
@@ -17,6 +23,80 @@ type CompleteERC20WithdrawalWorkflowParams = {
 
 const ERC20TokenType = 'ERC20';
 
+export async function executeRegisterAndWithdrawAllERC20(
```
brayansdt, line 26: Given this method will be used to withdraw ETH as well, we can rename ERC20 to Fungible instead. Or at least add a comment in all functions related to it explaining that this can also be used for ETH withdrawals.

brayansdt (approved):

sharif9876 (approved):

PR #1863, opened 2024-06-04
fix: safe guard process on

shineli1984:
# Summary
Safe guard the usage of `process.on`. Checkout widgets ended up importing minting backend code. This is to be defensive in calling `process.on`. A further investigation is needed to understand why Widgets will end up importing this module.


dreamoftrees (commented):
👍

PR #1878, opened 2024-06-07
feat:Sale Widget Sign Errors

jwhardwick:

# Summary

Better tracking of errors inside Sale Signing.

### DIFF
--- a/packages/checkout/widgets-lib/src/widgets/sale/hooks/useSignOrder.ts
+++ b/packages/checkout/widgets-lib/src/widgets/sale/hooks/useSignOrder.ts
@@ -186,9 +186,7 @@ export const useSignOrder = (input: SignOrderInput) => {
       data: string,
       gasLimit: number,
       method: string,
-    ): Promise<[hash: string | undefined, error: any]> => {
-      let transactionHash: string | undefined;
-
+    ): Promise<[hash: string | undefined, error?: SignOrderError]> => {
       try {
         const signer = provider?.getSigner();
         const gasPrice = await provider?.getGasPrice();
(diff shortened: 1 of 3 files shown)


mimi-imtbl (approved):

packages/checkout/widgets-lib/src/widgets/sale/hooks/useSignOrder.ts
```diff
@@ -205,13 +203,15 @@ export const useSignOrder = (input: SignOrderInput) => {
           await txnResponse?.wait();
         }
 
-        transactionHash = txnResponse?.hash || '';
+        const transactionHash = txnResponse?.hash;
+        if (!transactionHash) {
+          throw new Error('Transaction hash is undefined');
```
dreamoftrees, line 208: As this new Error is thrown is it worth capturing that case within the reason below in the try/catch with some more detail around missing txn hash?
Otherwise would it just be showing up as wallet_failed error for default, maybe thats enough?

dreamoftrees (approved):
Nice one @jwhardwick just let some comments.

packages/checkout/widgets-lib/src/widgets/sale/hooks/useSignOrder.ts
```diff
@@ -360,7 +358,7 @@ export const useSignOrder = (input: SignOrderInput) => {
         method,
       );
 
-      if (txnError || !hash) {
```
jwhardwick, line 363: here

packages/checkout/widgets-lib/src/widgets/sale/hooks/useSignOrder.ts
jwhardwick: Thanks @dreamoftrees. 
I'm not sure why `transactionHash` can be missing.
Before it was just being returned and if it was `undefined` the parent function would set an empty error (see https://github.com/immutable/ts-immutable-sdk/pull/1878/files#r1634210544), so I refactored it here so at least we have the generic error messaging if it does happen.

PR #1880, opened 2024-06-07
Docs: [DX-2907] Update README of "checkout" package

zaidarain1:
## Summary

Updates the documentation for the Checkout SDK folder

Preview [here](https://github.com/immutable/ts-immutable-sdk/blob/faeedc9713e6c449243827d0398f6e08cc5947f9/packages/checkout/README.md)

### DIFF
--- a/packages/checkout/README.md
+++ b/packages/checkout/README.md
@@ -1,122 +1,181 @@
-<div align="center">
-  <p align="center">
-    <a href="https://docs.x.immutable.com/docs">
-      <img src="https://cdn.dribbble.com/users/1299339/screenshots/7133657/media/837237d447d36581ebd59ec36d30daea.gif" width="280"/>
-    </a>
-  </p>
-  <h1>Immutable Checkout SDK</h1>
-</div>
+# Checkout
 
-# Overview
+## Table of Contents
 
-> **Warning** **IMMUTABLE CHECKOUT SDK IS UNSTABLE** <br/>
-> Since it has not hit the version 1.0 yet, its public interface should not be considered final. Future releases may include breaking changes without further notice. We will do our best to keep this documentation updated providing visibility on breaking changes planned.
+- [Pre-requisites](#pre-requisites)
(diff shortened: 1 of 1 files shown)


packages/checkout/README.md
```diff
@@ -1,123 +1,182 @@
...
-└───README.md
+```bash
+yarn install
 ```
 
-# Running locally
+### Checkout SDK Quick Start Guide
 
-## Dependencies
+All commands below need to be run in the context of the `checkout-sdk` package unless specified otherwise. Read more about context [here](../../../README.md#context).
```
wdinau, line 21: did you mean this is `checkout/sdk` package or is there another package called `checkout-sdk`?

packages/checkout/README.md
```diff
@@ -1,123 +1,182 @@
...
-| Dependency | Version |
-| ---------- | ------- |
-| Node       | v16+    |
-| Yarn       | v3.5.0  |
+```bash
+yarn start:dev
+```
 
-Node Module dependencies are still added to the package.json of each checkout module, however each module installs it's dependencies to a shared node_modules directory at the root level of the Unified SDK. This requires dependency versions to be aligned across all SDK submodules.
+Running in `sandbox`/`production` mode using the `immutable-testnet`/`mainnet` network based on the environment value set for the `SANDBOX_CONFIGURATION` config object in the `sdk.ts` file:
```
wdinau, line 29: this is the `src/sdk.ts` file, correct?
the doc mentioned "everything said below will be in context of checkout/sdk folder" so I wasn't sure if there suppose to be another sdk.ts 😆

packages/checkout/README.md
zaidarain1: The package names are used here instead of the directories. I've added the extra text at the end, for eg: "the context of the `checkout-sdk` package in `/packages/checkout/sdk`" to specify the folder too for that context sentence. I've made some changes throughout the documentation to be more clear, shouldn't be confusing anymore hopefully 😄

wdinau (approved):

jwhardwick (approved):

PR #1885, opened 2024-06-11
feat: bulk listing creation from orderbook SDK

Sam-Jeston:
# Summary

Implement bulk listing creation from the orderbook SDK. When the bulk listing methods are called with single items, the existing prepareListing and createListing methods are delegated to instead, reducing the impact on enriched confirmations for the majority of users who will not be listing in bulk.

## Added 

 - orderbook.prepareBulkListings


packages/orderbook/src/seaport/components.ts
```diff
@@ -9,3 +10,26 @@ export function getOrderComponentsFromMessage(orderMessage: string): OrderCompon
 
   return orderComponents;
 }
+
+export function getBulkOrderComponentsFromMessage(orderMessage: string): OrderComponents[] {
```
frankisawesome, line 14: Seems like bespoke and error prone logic here? Just observation

packages/orderbook/src/seaport/seaport.ts
```diff
@@ -327,6 +433,34 @@ export class Seaport {
...
 
+  private async getTypedDataFromBulkOrderComponents(
+    orderComponents: OrderComponents[],
+    numberOfOrders: number,
+  ): Promise<SignableAction['message']> {
+    const { chainId } = await this.provider.getNetwork();
+
+    const domainData = {
+      name: SEAPORT_CONTRACT_NAME,
+      version: SEAPORT_CONTRACT_VERSION_V1_5,
```
frankisawesome, line 444: Is this something we want to put in to a config/env? Especially now that we use it in multiple places, could be easy to forget if we move to 1.6

packages/orderbook/src/test/bulk-listings.e2e.ts
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    await contract.safeMint(offerer.address);
+
+    const { actions, createListings } = await sdk.prepareBulkListings({
+      makerAddress: offerer.address,
+      listingParams: orderParams,
+    });
+
+    const signatures = await actionAll(actions, offerer);
```
frankisawesome, line 63: Looks clean 👍

frankisawesome (dismissed):
lgtm, a few thoughts that are not blocking

packages/orderbook/src/seaport/components.ts
Sam-Jeston: Yeh undeniably not ideal - but unfortunately there is no real way around it because of how we expose the signable message to callers, rather than using the inbuilt seaport-js methods

packages/orderbook/src/seaport/seaport.ts
Sam-Jeston: Those constant values are all imported from seaport/constants.ts. If we move versions updating there will modify all the places we create this message

packages/orderbook/src/orderbook.ts
```diff
@@ -160,6 +163,104 @@ export class Orderbook {
...
+
+    // In the event of a single order, delegate to prepareListing as the signature is more
+    // gas efficient
+    if (listingParams.length === 1) {
+      const prepareListingResponse = await this.prepareListing({
+        buy: listingParams[0].buy,
+        makerAddress,
+        sell: listingParams[0].sell,
+        orderExpiry: listingParams[0].orderExpiry,
+      });
```
lfportal, line 196: **Suggestion:** For colocation clarity, could we consider lifting the internal call so the param manipulation matches the bulk operation below?

```suggestion
      const prepareListingResponse = this.seaport.prepareSeaportOrder(
        makerAddress,
        listingParams[0].sell,
        listingParams[0].buy,
        new Date(),
        listingParams[0].orderExpiry || new Date(Date.now() + 1000 * 60 * 60 * 24 * 365 * 2),
    );
```

lfportal (dismissed):

shineli1984 (approved):

CodeSchwert (approved):

lfportal (approved):

PR #1909, opened 2024-06-17
chore: make import path consistent

shineli1984:
# Summary
Some clean up on import paths.

This is not to make it consistent once for all but mainly for people that try out different approaches to build our sdk, especially when touching baseurl or removing tsconfig files, won't be bogged down by fixing all the imports.


keithbro-imx:
> This is to make import path consistent across all packages.

Feel like I'm missing some context. What's the motivation here?

What is going to stop it becoming "inconsistent" again straight away?

shineli1984:
@keithbro-imx 
This is not to make it consistent once for all but mainly for people trying out different approach to build our sdk, especially when touching baseurl or removing tsconfig giles, they won't be bogged down by fixing all the imports.

keithbro-imx (approved):

lfportal (approved):

rzmahmood (approved):

jiyounglee (approved):

jasonzwli (approved):

zaidarain1:
Can we set up an eslint rule to only allow relative imports to ensure this remains fixed?

Package [here](https://www.npmjs.com/package/eslint-plugin-import)
Rule [here](https://github.com/import-js/eslint-plugin-import/blob/HEAD/docs/rules/no-absolute-path.md)

shineli1984:
@zaidarain1 the baseUrl is their because I want to do the test fields separately in a different PR. also, that eslint rule is for absolute path tho. And I can't find a rule about "force relative path" in eslint.

zaidarain1:
> @zaidarain1 the baseUrl is there because I want to do the test file separately in a different PR. they still use non-relative paths. also, that eslint rule is for absolute path tho. And I can't find a rule about "force relative path" in eslint.

@shineli1984 
```
  "plugins": ["import"],
  "rules": {
    "import/no-unresolved": ["error"],
```

Running this rule with the above npm eslint-plugin-import packages errors out the imports for me as needed, but yeah the absolute path rule doesn't work. 
image

zaidarain1 (approved):

PR #1916, opened 2024-06-18
refactor:Refactor calldata function and create state object type

Benjimmutable:
# Summary
Refactoring the `getTxData` function to isolate some of the logic into a transaction-related file.
Create a `StateObject` type which is a bit friendlier to use than the double-nested `Record` required by Tenderly API.

## Added 
- Two types: `StateObject` and `StateDiff`.
- Function: `unwrapStateObjects` to convert `StateObject` array into `Record<string, Record<string, Record<string, string>>>`
- `lib/transactions.ts` and `lib/transactions.test.ts`.
- Function: `getBridgeTxCalldata`

## Changed
- Use `unwrapStateObjects` in `submitTenderlySimulations`.
- Change `getTxData` to `getConfigAndBridgeTxCalldata`
- Use `getBridgeTxCalldata` in `getConfigAndBridgeTxCalldata`


packages/internal/bridge/sdk/src/lib/tenderly.ts
```diff
@@ -22,13 +29,41 @@ import { getTenderlyEndpoint } from './utils';
  *    }
  *  }
  */
+export function unwrapStateObjects(
+  stateObjects: StateObject[],
+): Record<string, Record<string, Record<string, string>>> {
```
Benjimmutable, line 34: Alternatively, we could have a `StateObjects` class which essentially combines the storage of `StateObject` array and this `unwrap` function. LMK thoughts.

ermyas (approved):

PR #1919, opened 2024-06-18
fix: esm build compatibility

Sam-Jeston:
# Summary

`getBulkOrderTree` was recently imported from the seaport-js/lib and that broke the esm build fix. As there is no other way to reach this function at the moment, it and depending functions have been temporarily copied over


CodeSchwert (dismissed):

wdinau (commented):
@Sam-Jeston I did a keyword search and found 3 spots with `@opensea/seaport-js/lib/utils` 
one in component, one in transactions and one in tests.
could you please have a look at all three?

sdk/package.json
```diff
@@ -45,6 +45,7 @@
     "jwt-decode": "^3.1.2",
     "lru-memorise": "0.3.0",
     "magic-sdk": "^21.2.0",
+    "merkletreejs": "^0.3.11",
```
zaidarain1, line 48: Was wondering if the dependency needs to be added to the SDK package.json as well, if it's already listed in the orderbook package.json dependencies. Would the orderbook not already provide the dependency code when building, similar to other dependencies for other packages?

zaidarain1 (changes requested):

sdk/package.json
Sam-Jeston: This is a consequence of `yarn build`, I did not add it manually - any dependency in that package.json's of the packages gets lifted here

zaidarain1: I see, thank you for the clarification. Apologies, was not aware of this 😄

zaidarain1 (approved):

naveen-imtb (approved):

PR #1921, opened 2024-06-19
fix: [CM-726][Sale Widget] Remove duplicated event triggers

jhesgodi:
# Summary
Trigger top up view & summary screen load events once.


mimi-imtbl (approved):
nice `useMount`!

PR #1926, opened 2024-06-20
Passport - Add explicit Ethereum prefix for personal_sign signatures

Jon-Alonso:
# Summary
[BREAKING] Updated the Passport `personal_sign` method to explicitly add the 191 prefix to the message to be signed.


haydenfowler (approved):
Tested locally, LGTM ✅

PR #1929, opened 2024-06-21
fix: [DX-2945] Fix command for passport package running sdk and sample app in dev concurrently

zaidarain1:

# Summary

Fix concurrent passport dev commands and update README to represent change

### DIFF
--- a/packages/passport/README.md
+++ b/packages/passport/README.md
@@ -40,7 +40,7 @@ yarn dev
 yarn dev
 
 # Alternatively, you can run both commands in one terminal via the below concurrent yarn script
-# Context: passport in ./sdk
+# Context: passport-sdk-sample-app in ./sdk-sample-app
 yarn dev-with-sample
 ```
 
--- a/packages/passport/README.md
+++ b/packages/passport/README.md
@@ -39,8 +39,9 @@ yarn dev
 # Context: passport-sdk-sample-app in ./sdk-sample-app
 yarn dev
 
-# Alternatively, you can run both commands in one terminal via the below
-yarn workspace @imtbl/passport-sdk-sample-app dev & yarn workspace @imtbl/passport dev
+# Alternatively, you can run both commands in one terminal via the below concurrent yarn script
+# Context: passport in ./sdk
+yarn dev-with-sample
 ```
 
 Building to the `./sdk/dist` directory:
--- a/packages/passport/README.md
+++ b/packages/passport/README.md
@@ -41,7 +41,7 @@ yarn dev
 
 # Alternatively, you can run both commands in one terminal via the below concurrent yarn script
 # Context: passport-sdk-sample-app in ./sdk-sample-app
-yarn dev-with-sample
+yarn dev-with-sdk
 ```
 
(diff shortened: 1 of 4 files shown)


wdinau (approved):

packages/passport/sdk/package.json
```diff
@@ -61,6 +62,7 @@
   "scripts": {
     "build": "NODE_ENV=production rollup --config rollup.config.js",
     "dev": "rollup --config rollup.config.js -w",
```
haydenfowler, line 64: Can I suggest that we move this script to the Passport sample app please? When the SDK is published it won't contain the sample app, nor will the `concurrently` package be available, so this script may cause confusion for consumers.

zaidarain1: Yeap sounds like a much better idea. Moved 😄

packages/passport/README.md
```diff
@@ -39,8 +39,9 @@ yarn dev
 # Context: passport-sdk-sample-app in ./sdk-sample-app
 yarn dev
 
-# Alternatively, you can run both commands in one terminal via the below
-yarn workspace @imtbl/passport-sdk-sample-app dev & yarn workspace @imtbl/passport dev
+# Alternatively, you can run both commands in one terminal via the below concurrent yarn script
+# Context: passport-sdk-sample-app in ./sdk-sample-app
+yarn dev-with-sdk
```
haydenfowler, line 44: Can we move this to the sample app README please?

haydenfowler: Oh disregard me, didn't realise we'd previously merged both README's into a root README.

haydenfowler, line 44: Since we combined `passport/sdk/README` and `passport/sdk-sample-app/README` into `/passport/README`, the commands mentioned in this file would fail, as no `/passport/package.json` exists.

Can we update these commands to include the workspace please? E.g:
```
yarn workspace @imtbl/passport dev
...
yarn workspace @imtbl/passport-sdk-sample-app dev
...
yarn workspace @imtbl/passport-sdk-sample-app dev-with-sdk
```

zaidarain1: The context that's mentioned above specifies the context where the commands needs to be run. I've added a link to what this means at the start of the quick start section. The context tells you where to run it, and you decide how to then run it within that context, since devs may prefer their own way, either cding or yarn workspace.

For eg, with that above command, if the command needs to run in the sample app context as mentioned, then devs can do either 
```bash
yarn workspace @imtbl/passport-sdk-sample-app dev-with-sdk
```
or
```bash
cd packages/passport/sdk-sample-app
yarn dev-with-sdk
```

I've chosen this approach since it allows for flexibility rather than forcing a specific way of doing things, that can be annoying. I also went with this since I've seen devs using both ways rather than one way being much more consistent for all/most devs

zaidarain1: Another approach would be to just set up these dev commands in the root package.json for each package, but then that would bloat that with a ton of scripts and I felt that wouldn't have been the best approach.

wdinau (approved):

haydenfowler (approved):

PR #1934, opened 2024-06-22
fix: no russian nesting doll

shineli1984:
# Summary
Make esm module not replicating each other as much as possible.
The size of this approach is ~31 MB vs 37.8MB on npm

## Changed
Group esm package inputs into a rollup config so that roll up can reason about their relationship.

### DIFF
--- a/sdk/rollup.config.js
+++ b/sdk/rollup.config.js
@@ -36,52 +36,46 @@ const getFilesToBuild = () => {
   return [...files, ...returnModules];
 };
 
-const getFileBuild = (inputFilename) => [
-  {
-    input: `./src/${inputFilename}.ts`,
-    output: {
-      dir: 'dist',
-      format: 'es',
-    },
-    plugins: [
-      nodeResolve({
-        resolveOnly: getPackages(),
-      }),
-      json(),
-      commonJs(),
-      typescript({
-        declaration: true,
-        declarationDir: './dist/types',
-      }),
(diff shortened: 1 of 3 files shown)


zaidarain1:
The image below is the node modules install folder when installing the sdk. We have some extra stuff we can remove to reduce the size even more when bundling with rollup. For eg, we could remove the src folders of each and the sdk scripts folder, as well as some of the other files in the root of the repo that are not needed. Worth the effort? Can save another half to one mb maybe.

image

sdk/rollup.config.js
```diff
@@ -36,52 +36,57 @@ const getFilesToBuild = () => {
...
+        typescript({
+          declaration: true,
+          declarationDir: './dist/types',
+        }),
+        replace({
+          exclude: 'node_modules/**',
+          preventAssignment: true,
+          __SDK_VERSION__: pkg.version,
+        })
+      ],
```
zaidarain1, line 72: adding `terser()` in between these two lines will reduce the bundle size by a further tested 1.6MB (32.8MB to 31.2MB). Is there any reason for why we wouldn't want to minify?

```suggestion
        }),
        terser(),
      ],
```

zaidarain1 (changes requested):

zaidarain1 (approved):

keithbro-imx (approved):

PR #1938, opened 2024-06-24
refactor: SDK workflow builds

CodeSchwert:

# Summary

Refactor the Build, Lint, Typecheck, Test workflow re-use build and `node_module` artefacts across steps rather than re-running the install and build steps on every job.

- No impact to external customers

## Changed
- Cache build and `node_modules` in Build, Lint, Typecheck, Test workflow

- More caching to follow up after this PR is merged


wdinau (commented):
Hi Nik, just some newbie questions
Q1: is this the only place to use cache (compare to all other places in .github/workflows) ?
Q2: we put cache node modules on the top in `build`, and all subsequent steps are using that cache.would there ever be cases where people re-run part of the job, (say `build-passport`) without running the `build`, and they'd get to use the latest build of their change?  
Q3: ` build-cache-sdk` (in build-sdk.yaml) is that a step `build` job (of build-lint-typecheck-test.yaml) can re-use instead of copy paste same logic again?

CodeSchwert:
> Hi Nik, just some newbie questions Q1: is this the only place to use cache (compare to all other places in .github/workflows) ? Q2: we put cache node modules on the top in `build`, and all subsequent steps are using that cache.would there ever be cases where people re-run part of the job, (say `build-passport`) without running the `build`, and they'd get to use the latest build of their change? Q3: ` build-cache-sdk` (in build-sdk.yaml) is that a step `build` job (of build-lint-typecheck-test.yaml) can re-use instead of copy paste same logic again?

Q1: yes, currently only the build-lint-typecheck workflow is using the cache. We can expand to more workflows, but makes the most sense to update this one first since it has multiple `build` steps in the same workflow.
Q2: The node_modules cache is based off the yarn.lock changing, which is unlikely to change during the workflow. Infact, the workflow is designed to error out if the `yarn.lock` file changes.
Q3: I'm planning to use the `build-sdk` workflow to share the build artefacts across workflows. Apparently it needs to be in the main branch first before I can do that... so including it in this PR for future use. We can potentially reuse both `node_modules` and `build` artefacts across all the workflow runs in the PR.

.github/workflows/build-sdk.yaml
```diff
@@ -0,0 +1,41 @@
...
+          cache: 'yarn'
+
+      - name: Install dependencies
+        run: yarn install --immutable
+
+      - name: Cache build artifacts
+        uses: actions/cache@v4
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-build-cache-deps-${{ hashFiles('yarn.lock') }}
```
wdinau, line 30: where is `build-cache-deps` coming from?
I don't know enough but i can see a pattern here the key is build-cache-xxx and the job is called `build-cache-sdk` is there another key/job exist with the word `deps` ?

CodeSchwert: Not yet, that will be a follow up PR to use that cache once this workflow is merged to main.

wdinau (approved):

PR #1950, opened 2024-06-28
fix: ID-2067 Remove use of Buffer in personal sign

imx-mikhala:

# Summary
- Removes Buffer to fix personal sign when Buffer is not defined

[Here](https://github.com/imx-mikhala/passport-browser-bundle-sdk) is a sample-app put together to test Passport using browser bundle

There is an SDK alpha release of this branch [here](https://cdn.jsdelivr.net/npm/@imtbl/sdk@1.45.0-alpha.4/dist/index.browser.min.js)

When Buffer was missing, the confirmation app displays the hex string of the message:

telegram-cloud-photo-size-5-6332289006158593993-y

The following is with the patch applied:

Screenshot 2024-06-27 at 1 23 13 PM

### DIFF
--- a/packages/passport/sdk/src/utils/string.ts
+++ b/packages/passport/sdk/src/utils/string.ts
@@ -1,27 +1,14 @@
-/* eslint-disable no-console */
 import { utils } from 'ethers';
 
 export const hexToString = (hex: string) => {
   if (!hex) return hex;
 
-  console.log('hex', hex);
-
   try {
     const stripped = utils.stripZeros(utils.arrayify(hex));
     const buff = Buffer.from(stripped);
 
-    console.log('stripped', stripped);
-    console.log('buff', buff);
-
-    if (buff.length === 32) {
-      console.log('buff.length === 32', hex);
-    } else {
-      console.log('buff.length !== 32', utils.toUtf8String(stripped));
-    }
-
     return buff.length === 32 ? hex : utils.toUtf8String(stripped);
   } catch (e) {
-    console.log('error from hexToString', e);
     return hex;
   }
 };
--- a/packages/passport/sdk/src/utils/string.ts
+++ b/packages/passport/sdk/src/utils/string.ts
@@ -5,7 +5,7 @@ export const hexToString = (hex: string) => {
 
   try {
     const stripped = utils.stripZeros(utils.arrayify(hex));
-    return stripped.length === 32 ? hex : utils.toUtf8String(stripped);
+    return utils.toUtf8String(stripped);
   } catch (e) {
     return hex;
   }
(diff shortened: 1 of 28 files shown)


packages/passport/sdk/src/zkEvm/personalSign.ts
```diff
@@ -38,7 +38,12 @@ export const personalSign = async ({
     throw new JsonRpcError(RpcErrorCode.INVALID_PARAMS, 'personal_sign requires the signer to be the from address');
   }
 
-  // Convert message into a string if it's a hex
+  if (typeof window !== 'undefined' && !window.Buffer) {
```
keithbro-imx, line 41: 👀 very interesting.. why does this work?

imx-mikhala: The `hexToString` function called after this if statement uses `Buffer` which causes an issue with using the browser bundle as `Buffer` doesn't exist. Adding it to the window if Buffer isnt defined fixes the issue.

packages/passport/sdk/src/utils/string.ts
```diff
@@ -1,18 +1,11 @@
...
-    // Ensures Buffer is available when using the browser bundle of the SDK
-    if (typeof window !== 'undefined' && !window.Buffer) {
-      const bufferModule = await import('buffer');
-      window.Buffer = bufferModule.Buffer;
-    }
-
     const stripped = utils.stripZeros(utils.arrayify(hex));
-    const buff = Buffer.from(stripped);
-    return buff.length === 32 ? hex : utils.toUtf8String(stripped);
+    return utils.toUtf8String(stripped);
```
imx-mikhala, line 8: todo: confirm purpose of `buff.length === 32 `

packages/passport/sdk/src/utils/string.ts
```diff
@@ -5,8 +5,7 @@ export const hexToString = (hex: string) => {
 
   try {
     const stripped = utils.stripZeros(utils.arrayify(hex));
-    const buff = Buffer.from(stripped);
-    return buff.length === 32 ? hex : utils.toUtf8String(stripped);
+    return stripped.length === 32 ? hex : utils.toUtf8String(stripped);
```
imx-mikhala, line 8: @Jon-Alonso @haydenfowler  Can someone confirm what the `length === 32` check is for here?

imx-mikhala: Thanks for the info. We can probably just remove the check then if other implementations result in the same behaviour.

packages/passport/sdk/src/utils/string.ts
```diff
@@ -5,8 +5,7 @@ export const hexToString = (hex: string) => {
 
   try {
     const stripped = utils.stripZeros(utils.arrayify(hex));
-    const buff = Buffer.from(stripped);
```
keithbro-imx, line 8: Love it when the fix to a bug is deleting code 😂

keithbro-imx (approved):

PR #1952, opened 2024-07-01
feat: [CM-728] Tracks popup blocked event

alejoloaiza:
# Summary
Adding an event when the Passport popups are blocked

## Security

### DIFF
--- a/packages/passport/sdk/src/confirmation/confirmation.ts
+++ b/packages/passport/sdk/src/confirmation/confirmation.ts
@@ -256,7 +256,6 @@ export default class ConfirmationScreen {
         width: this.popupOptions?.width || CONFIRMATION_WINDOW_WIDTH,
         height: this.popupOptions?.height || CONFIRMATION_WINDOW_HEIGHT,
       });
-
       this.timer = setInterval(timerCallback, CONFIRMATION_WINDOW_CLOSED_POLLING_DURATION);
     } catch { /* Empty */ }
   }
--- a/packages/passport/sdk/src/confirmation/confirmation.ts
+++ b/packages/passport/sdk/src/confirmation/confirmation.ts
@@ -258,6 +258,6 @@ export default class ConfirmationScreen {
       });
 
       this.timer = setInterval(timerCallback, CONFIRMATION_WINDOW_CLOSED_POLLING_DURATION);
-    } catch { /* Catch */ }
+    } catch { /* Empty */ }
   }
 }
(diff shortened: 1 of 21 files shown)


packages/passport/sdk/src/confirmation/confirmation.ts
```diff
@@ -180,8 +182,10 @@ export default class ConfirmationScreen {
         height: popupOptions?.height || CONFIRMATION_WINDOW_HEIGHT,
       });
       this.overlay = new Overlay(this.config.popupOverlayOptions);
-    } catch (e) {
+    } catch (error) {
       // If an error is thrown here then the popup is blocked
+      const errorMessage = error instanceof Error ? error.message : String(error);
+      trackError('passport', 'PassportLoginPopupDenied', new Error(errorMessage));
```
haydenfowler, line 188: A couple of things here:

1. We can omit `Passport` in the event name, as we set the module name to `passport`
2. The existing event names use camelCase, not PascalCase
3. `confirmation.ts` isn't used for login's, it's used for confirmation windows (e.g. sending a transaction, or approving a signature). You're probably after the blocked popup logic in `authManager.ts` ([here](https://github.com/immutable/ts-immutable-sdk/blob/4927bc384b51ab6905a2d57c3a20cf2b5c0d1d4e/packages/passport/sdk/src/authManager.ts#L196))

alejoloaiza: hi @haydenfowler thanks for the review, we actually want to track `confirmation` , the naming was wrong from the beggining

packages/passport/sdk/src/confirmation/confirmation.ts
```diff
@@ -180,8 +182,10 @@ export default class ConfirmationScreen {
         height: popupOptions?.height || CONFIRMATION_WINDOW_HEIGHT,
       });
       this.overlay = new Overlay(this.config.popupOverlayOptions);
-    } catch (e) {
+    } catch (error) {
       // If an error is thrown here then the popup is blocked
+      const errorMessage = error instanceof Error ? error.message : String(error);
+      trackError('passport', 'confirmationPopupDenied', new Error(errorMessage));
```
haydenfowler, line 188: One non-blocking suggestion, but we typically refer to the popup as being "blocked" rather than denied. It would be good to keep the naming consistent here.
```suggestion
      trackError('passport', 'confirmationPopupBlocked', new Error(errorMessage));
```

haydenfowler (approved):

PR #1957, opened 2024-07-02
[DX-2988] faster and simpler watch mode and build in ci

shineli1984:
# Summary
Introduce a caching enabled watch mode on **root** level like below:

```bash
yarn dev PACKAGE_NAME # e.g. yarn w @imtbl/passport
```

or just 
```bash
yarn dev # this will prompt user to select a package
```

This command will do the following:
- build once the specified package. (including any dependencies it needs)
- watch for file changes on the specified package and the packages it depends on. When changes are detected, it will rebuild the specified package, including any dependencies it needs.
- cache any unchanged artefacts locally and potentially remotely and reuse them in building process.

## changed
- Introduce NX for the build and dev scripts
- Utilising NX for caching purpose in both local and CI.


packages/blockchain-data/sdk/package.json
```diff
@@ -34,7 +34,7 @@
   "repository": "immutable/ts-immutable-sdk.git",
   "scripts": {
     "build": "NODE_ENV=production rollup --config rollup.config.js",
-    "dev": "rollup --config rollup.config.js -w",
+    "d": "swc src -d dist --strip-leading-paths --ignore '**/*.test.*'",
```
shineli1984, line 37: We can not keep using `dev` as the script name because `nx` treat `dev` script output as un-cacheable.

README.md
```diff
@@ -82,38 +82,20 @@ Most packages in the workspace have their own README file that provides specific
...
+Build a package and start watch mode for a package:
 
 ```bash
-# Context: Specified package
-yarn start
+yarn w YOUR_PACKAGE_NAME
 ```
 
-Some packages also come with a sample app that you can run in development mode. To run the sample app, you first need to run the SDK it depends on in development mode, and then run the sample app development mode in a seperate terminal. For example, to run the `checkout-sdk-sample-app` located at `/packages/checkout/sdk-sample-app`, you can use the following commands:
-
```
wdinau, line 99: this part explained what to do with sample app. did you delete them because we know it'd change soon anyway in sample code proj?

README.md
```diff
@@ -82,38 +82,20 @@ Most packages in the workspace have their own README file that provides specific
...
 
-### Development Mode
+### Development watch mode
 
-Development mode allows you to run a package in a local development environment. This is useful for testing changes and new features before deploying them to production. Any changes made will be reflected in real-time on save, so you can see the effects of your changes immediately. These are propogated in real time to the sample app, if it is running, for that SDK too.
-
-Development mode for sdk packages will tend to use rollup and its watch mode to rebuild the package on file changes. This build output, stored in the dist folder of the sdk package, is what is exposed to any other packages that depend on it, including the sample app.
-
-To run a package in development mode, you can use the following command in the context of that package:
+Build a package and start watch mode for a package:
```
wdinau, line 87: i like this. clean and clear

README.md
```diff
@@ -82,38 +82,20 @@ Most packages in the workspace have their own README file that provides specific
...
-Some packages also come with a sample app that you can run in development mode. To run the sample app, you first need to run the SDK it depends on in development mode, and then run the sample app development mode in a seperate terminal. For example, to run the `checkout-sdk-sample-app` located at `/packages/checkout/sdk-sample-app`, you can use the following commands:
-
+e.g. for @imtbl/passport package
 ```bash
-yarn workspace @imtbl/checkout-sdk start
-
-# Or from the root of the monorepo
-cd packages/checkout/sdk
-yarn start
+yarn w @imtbl/passport
```
wdinau, line 95: any example in readme use `d` ?  assume instead of yarn dev we'd use yarn d for each package to be built and watch?

README.md
shineli1984: The sample apps are well understood by each team. because they do whatever they like in the sample apps

README.md
shineli1984: The use case for `yarn d` should be gone now with this PR.

EdwardJES (approved):

naveen-imtb (approved):

hiep-immutable (approved):

yundifu (commented):
🥳

yundifu (dismissed):

README.md
keithbro-imx, line 95: I tried this and got

```
~/dev/ts-immutable-sdk try-out-nx-watch-mode                                                                                                                                                          
❯ yarn w @imtbl/dex-sdk
Usage Error: Couldn't find a script named "w".

$ yarn run [--inspect] [--inspect-brk] [-T,--top-level] [-B,--binaries-only] <scriptName> ...
```

Did I do it wrong?

keithbro-imx: `yarn run dev @imtbl/dex-sdk` seems to be the correct command.

keithbro-imx: When running this command, I made a change in dex-sdk to introduce a type error. The watcher rebuilt dex-sdk, but said said everything was fine. Is it supposed to fail?

keithbro-imx: My last comment is: this is great, I can see what problem it solves, thank you. Maybe you're already planning this, but what I REALLY want is to be able to run the unit tests for a package in watch mode, with all dependencies also being rebuilt. That would be a thing of beauty.

nx-cloud.example.env
```diff
@@ -0,0 +1,4 @@
+# Create a file named nx-cloud.env in the root of the repo and obtain the contents from the Platform Engineering Shared 1Password vault
```
keithbro-imx, line 1: Under what circumstance would I want to do this?

README.md
shineli1984: I'll fix the doc and see if I can reproduce your problem in dex-sdk. which file you changed btw? @keithbro-imx 

for test and linting, they are coming.

keithbro-imx: You can change any file to see the behaviour. I think the problem is that SWC transpiles the code but doesn't typecheck it - so in a sense this is working as expected given the choice of using SWC.

sharif9876 (approved):

nx-cloud.example.env
shineli1984: if you want to save cache into the cloud from local

shineli1984: I'll make that clear.

README.md
shineli1984: I think the IDE will catch it and the typecheck script on CI will catch it. case edge enough for us to not cater for this scenario?

nx-cloud.example.env
keithbro-imx: What would be the use case for that?

README.md
keithbro-imx: Yeah I think that's fine

nx-cloud.example.env
shineli1984: engineer 1 has a branch and build locally in development -> cache is stored in the cloud.
engineer 2 checkout the branch and build it locally -> cache will be from the cloud.

keithbro-imx: Is the intent of engineer 1 to make engineer 2's builds faster? So for example, I would write to nx-cloud so that my team mate can build the same branch faster?

keithbro-imx (approved):

nx-cloud.example.env
shineli1984: yes. that's the idea.

PR #1965, opened 2024-07-10
ID-1897 WIP

matthewmuscat:

# Summary

## Security

### DIFF
--- a/packages/passport/sdk/src/zkEvm/zkEvmProvider.ts
+++ b/packages/passport/sdk/src/zkEvm/zkEvmProvider.ts
@@ -1,6 +1,5 @@
-import { StaticJsonRpcProvider, Web3Provider } from '@ethersproject/providers';
+import { StaticJsonRpcProvider } from '@ethersproject/providers';
 import { MultiRollupApiClients } from '@imtbl/generated-clients';
-import { Signer } from '@ethersproject/abstract-signer';
 import { utils } from 'ethers';
 import {
   Flow, identify, trackError, trackFlow,
@@ -66,16 +65,6 @@ export class ZkEvmProvider implements Provider {
 
   readonly #relayerClient: RelayerClient;
 
-  /**
-   * This property is set during `#initialiseEthSigner` and stores the signer in a promise.
-   * This property is not meant to be accessed directly, but through the
-   * `#getSigner` method.
-   * @see getSigner
-   */
-  #ethSigner?: Promise<Signer | undefined> | undefined;
-
-  #signerInitialisationError: unknown | undefined;
-
   public readonly isPassport: boolean = true;
 
   constructor({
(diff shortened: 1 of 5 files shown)


packages/passport/sdk/src/zkEvm/zkEvmProvider.ts
```diff
@@ -256,28 +243,25 @@ export class ZkEvmProvider implements Provider {
         }
       }
       case 'eth_accounts': {
-        const zkEvmAddress = await this.#getZkEvmAddress();
+        const zkEvmAddress = zkEvmUser?.zkEvm.ethAddress;
```
haydenfowler, line 246: If `zkEvmUser` is null, then `zkevm` will also be null, and so this will throw a `cannot read properties of undefined` error

matthewmuscat: just missing an optional chain, pushed https://github.com/immutable/ts-immutable-sdk/pull/1965/commits/b9397aa3454878abf3dcbd718e98437982a80ff3

packages/passport/sdk/src/zkEvm/zkEvmProvider.ts
```diff
@@ -201,77 +131,86 @@ export class ZkEvmProvider implements Provider {
...
-        const flow = trackFlow('passport', 'ethRequestAccounts');
-
-        try {
-          const user = await this.#authManager.getUserOrLogin();
-          flow.addEvent('endGetUserOrLogin');
-
-          this.#initialiseEthSigner(user);
+    // Get user from local storage
+    const user = await this.#authManager.getUser();
+    const zkEvmUser = this.#getZkEvmUser(user);
```
haydenfowler, line 147: This will have unintended consequences on the passthrough methods when `getUser` is called. Two scenarios that I can think of:

1. If a token refresh is triggered, then we'll need to wait for that process to resolve before we can pass the method along to the node
2. If a token refresh is triggered and it fails, then this error will be surfaced to the caller

matthewmuscat: >If a token refresh is triggered, then we'll need to wait for that process to resolve before we can pass the method along to the node

`await this.#authManager.getUser();` already awaits the refreshing promise so I think this is fine?

> If a token refresh is triggered and it fails, then this error will be surfaced to the caller

[Looks like this is fine](https://github.com/immutable/ts-immutable-sdk/blob/feature/ID-1897-magic-session-management/packages/passport/sdk/src/zkEvm/zkEvmProvider.ts#L245) for `eth_requestAccounts`, `personal_sign`, etc as we surface any errors anyway to the caller. 

However for `eth_accounts` and a few other RPC calls, this would throw when it should just return `[]` as we don't handle error handling. I'll give this some thought.

matthewmuscat: https://github.com/immutable/ts-immutable-sdk/pull/1965/commits/ac1424aec118caa4427220242b68b2e06ef69c1c

PR #1966, opened 2024-07-11
fix: allow use of localstorage to be optional in metrics

cajames:

# Summary

Allows the use of localstorage to be optional, some environments (e.g. game bridge) doesn't have robust local storage implementations, so disable it for those environments.

## Security


nattb8 (dismissed):

cajames:
Not using this will switch to indexed-based storage due to the low write performance of local storage in chromium.

PR #1967, opened 2024-07-11
refactor: game bridge to use metrics trackError and trackDuration methods

CodeSchwert:

# Summary

[DX-2970](https://immutable.atlassian.net/browse/DX-2970)
[DX-2971](https://immutable.atlassian.net/browse/DX-2971)

Switch to the new metrics `trackError` and `trackDuration` methods.

Game bridge functions have also been refactored to throw errors instead of setting a string error message in the callback function, so all errors are now handled in the `catch` block.

- Error stack should give better output as to which function call raised the error.

## Changed
- Game bridge function errors now handled consistently

[DX-2970]: https://immutable.atlassian.net/browse/DX-2970?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DX-2971]: https://immutable.atlassian.net/browse/DX-2971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/game-bridge/src/index.ts
```diff
@@ -78,13 +83,29 @@ declare function blu_event(event: string, data: string): void;
...
 declare function UnityPostMessage(message: string): void;
 
-const callbackToGame = (data: object) => {
+interface CallbackToGameData {
+  responseFor: string;
+  requestId: string;
+  success: boolean;
+  error?: string;
+  errorType?: string | null;
+  code?: string;
```
nattb8, line 92: `code`, `deviceCode`, `url` and `interval` are only returned for initDeviceFlow. Could they not be part of this CallbackToGameData? Same for `accounts`

CodeSchwert: Added a union type instead so it's clearer

YermekG (approved):

packages/game-bridge/src/index.ts
```diff
@@ -78,13 +83,36 @@ declare function blu_event(event: string, data: string): void;
...
-const callbackToGame = (data: object) => {
+// Type for common callback data
+type CommonCallbackData = {
+  responseFor: string;
+  requestId: string;
+  success: boolean;
+  error: string | null;
+  errorType?: string | null;
+  result?: any;
+  accounts?: string[];
```
nattb8, line 94: accounts is not comment too, only used by request accounts. Should it be another type?

CodeSchwert: Added another type for request accounts

nattb8 (approved):

PR #1968, opened 2024-07-11
feat: [DX-2998] Integrate NX for builds

zaidarain1:
## Description

This PR sets up the initial integration of NX into the Typescript SDK. For now this only adds support for caching the build artifacts, including using the NX cloud remote caching, however additional caching of more steps, as well as fixing workflow hierarchy will come shortly in the future.


feenaIMX (approved):

sharif9876 (dismissed):

yundifu (approved):

nx.json
```diff
@@ -0,0 +1,25 @@
...
+  "defaultBase": "main",
+  "plugins": [
+    {
+      "plugin": "@nx/rollup/plugin",
+      "options": {
+        "buildTargetName": "rollup:build"
+      }
+    }
+  ],
+  "nxCloudAccessToken": "Mzg3ZGY1MWUtYmYyNy00ZmE4LTkyNDAtYjYxZmJmYmE4NWQ3fHJlYWQ="
```
haydenfowler, line 24: Should this be committed? Is this a read only access token?

zaidarain1: Yeap this is a read only fallback token.

I had a chat with Tim, and we have a read/write access token in 1pass now, the nx-cloud.example.env now details how to get that one set up. We also have a read/write token set up for the github workflow set up in the secrets. Since our repo source code is public anyway the only thing this will allow people to access is the build output artifacts that they can get from manually building, and won't be able to push anything malicious to the cache server.

haydenfowler: Good to know, thanks @zaidarain1!

PR #1978, opened 2024-07-16
[DX-3013] feat: passport indexed db storage for unity sdk

nattb8:

# Summary

The device code authentication works well on Android, iOS, and Mac, but the Unity SDK encounters issues on Windows. After logging in, the user's OIDC data stored in localStorage often disappears when reopening the game. This happens because Windows initially stores localStorage items in memory, unlike IndexedDB, which writes directly to disk. Upon looking at the Chromium source code for version 122.0.6261.130, local storage items only get committed after every [5 seconds](https://source.chromium.org/chromium/chromium/src/+/refs/tags/122.0.6261.130:components/services/storage/dom_storage/local_storage_impl.cc;drc=81b3c99740a33d3ed7cf86781031637417c3aa38;l=154). Additionally, this [line](https://source.chromium.org/chromium/chromium/src/+/main:components/services/storage/dom_storage/local_storage_impl.cc;l=194;drc=fa1e0cf5238e2600039f60af6909aefbd4cf003c) is also the other cause of the issue because the metrics package was using local storage a lot.

To ensure better compatibility with game engines, this PR updates game SDKs to use IndexedDB instead of local storage for storing OIDC data.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
   "scripts": {
     "build": "yarn workspace @imtbl/sdk updateDependencies && NODE_OPTIONS=--max-old-space-size=14366 nx run-many --target=build --projects=@imtbl/sdk,@imtbl/checkout-widgets && yarn syncpack:format && yarn wsrun -p @imtbl/sdk -a -m copyBrowserBundles",
     "build:onlysdk": "NODE_OPTIONS=--max-old-space-size=14366 nx run-many --target=build --projects=@imtbl/sdk && yarn syncpack:format",
+    "dev": "./dev.sh",
     "docs:build": "typedoc",
     "docs:serve": "http-server ./docs --cors -p 8080 -c-1",
     "lint": "wsrun --exclude-missing -e lint --no-error-on-unmatched-pattern",
@@ -64,8 +65,7 @@
     "test:checkout:sdk:coverage": "wsrun -p @imtbl/checkout-sdk --exclude-missing -e test:coverage",
     "test:examples": "cd examples/ts-immutable-sample && yarn test && yarn test:e2e",
     "test:vpn": "RUN_VPN_TESTS=1 wsrun --exclude-missing -e test",
-    "typecheck": "wsrun --exclude-missing typecheck",
-    "dev": "./dev.sh"
+    "typecheck": "wsrun --exclude-missing typecheck"
   },
   "workspaces": {
     "packages": [
--- a/packages/passport/sdk/src/authManager.ts
+++ b/packages/passport/sdk/src/authManager.ts
@@ -11,6 +11,7 @@ import axios from 'axios';
 import * as crypto from 'crypto';
 import jwt_decode from 'jwt-decode';
 import { getDetail, Detail } from '@imtbl/metrics';
+import localForage from 'localforage';
 import DeviceCredentialsManager from './storage/device_credentials_manager';
 import logger from './utils/logger';
 import { isTokenExpired } from './utils/token';
@@ -31,6 +32,7 @@ import {
 } from './types';
 import { PassportConfiguration } from './config';
 import Overlay from './overlay';
+import { LocalForageAsyncStorage } from './storage/LocalForageAsyncStorage';
 
 const formUrlEncodedHeader = {
   headers: {
@@ -44,7 +46,14 @@ const authorizeEndpoint = '/authorize';
 const getAuthConfiguration = (config: PassportConfiguration): UserManagerSettings => {
   const { authenticationDomain, oidcConfiguration } = config;
 
-  const store = typeof window !== 'undefined' ? window.localStorage : new InMemoryWebStorage();
+  let store;
+  if (config.crossSdkBridgeEnabled) {
+    store = new LocalForageAsyncStorage('ImmutableSDKPassport', localForage.INDEXEDDB);
+  } else if (typeof window !== 'undefined') {
+    store = window.localStorage;
+  } else {
+    store = new InMemoryWebStorage();
+  }
   const userStore = new WebStorageStateStore({ store });
 
   const endSessionEndpoint = new URL(logoutEndpoint, authenticationDomain.replace(/^(?:https?:\/\/)?(.*)/, 'https://$1'));
--- a/packages/passport/sdk/package.json
+++ b/packages/passport/sdk/package.json
@@ -21,6 +21,7 @@
     "ethers": "^5.7.2",
     "events": "^3.3.0",
     "jwt-decode": "^3.1.2",
+    "localforage": "^1.10.0",
     "magic-sdk": "^21.2.0",
     "oidc-client-ts": "2.4.0",
     "uuid": "^8.3.2"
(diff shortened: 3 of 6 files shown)


YermekG (approved):

packages/passport/sdk/src/authManager.ts
```diff
@@ -44,7 +45,12 @@ const authorizeEndpoint = '/authorize';
 const getAuthConfiguration = (config: PassportConfiguration): UserManagerSettings => {
   const { authenticationDomain, oidcConfiguration } = config;
 
-  const store = typeof window !== 'undefined' ? window.localStorage : new InMemoryWebStorage();
+  let store;
+  if (config.crossSdkBridgeEnabled) {
+    store = new IndexedDbStorage();
+  } else {
+    store = typeof window !== 'undefined' ? window.localStorage : new InMemoryWebStorage();
+  }
```
haydenfowler, line 53: This should be more readable if we avoid nesting a ternary within an `if`:
```suggestion
  if (config.crossSdkBridgeEnabled) {
    store = new IndexedDbStorage();
  } else if (typeof window !== 'undefined') {
    store = window.localStorage;
  } else {
    store = new InMemoryWebStorage();
  }
```

YermekG (approved):

packages/passport/sdk/src/storage/IndexedDbStorage.ts
```diff
@@ -0,0 +1,173 @@
...
+        reject(new Error('Failed to open IndexedDB database'));
+      };
+
+      request.onsuccess = () => {
+        this.db = request.result;
+        resolve();
+      };
+
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      request.onupgradeneeded = (event) => {
```
haydenfowler, line 23:
```suggestion
      request.onupgradeneeded = () => {
```

packages/passport/sdk/src/storage/IndexedDbStorage.ts
```diff
@@ -0,0 +1,173 @@
...
+          resolve();
+        };
+        clearRequest.onerror = () => {
+          reject(new Error('Failed to clear object store'));
+        };
+      } else {
+        reject(new Error('Object store not available'));
+      }
+    });
+  }
```
haydenfowler, line 82:
```suggestion
  async clear(): Promise<void> {
    const store = await this.getObjectStore('readwrite');

    return new Promise<void>((resolve, reject) => {
      if (store) {
        const clearRequest = store.clear();
        clearRequest.onsuccess = () => {
          resolve();
        };
        clearRequest.onerror = () => {
          reject(new Error('Failed to clear object store'));
        };
      } else {
        reject(new Error('Object store not available'));
      }
    });
  }
```

haydenfowler:
Can we make sure that all of the functions that return a promise are marked as `async` please? E.g. `key, `length`, ...

packages/passport/sdk/src/storage/IndexedDbStorage.ts
```diff
@@ -0,0 +1,173 @@
...
+          resolve();
+        };
+        deleteRequest.onerror = () => {
+          reject(new Error(`Failed to delete item with key '${key}'`));
+        };
+      } else {
+        reject(new Error('Object store not available'));
+      }
+    });
+  }
```
haydenfowler, line 154:
```suggestion
  async removeItem(key: string): Promise<void> {
    const store = await this.getObjectStore('readwrite');

    return new Promise<void>((resolve, reject) => {
      if (store) {
        const deleteRequest = store.delete(key);
        deleteRequest.onsuccess = () => {
          resolve();
        };
        deleteRequest.onerror = () => {
          reject(new Error(`Failed to delete item with key '${key}'`));
        };
      } else {
        reject(new Error('Object store not available'));
      }
    });
  }
```

packages/passport/sdk/src/storage/IndexedDbStorage.ts
```diff
@@ -0,0 +1,173 @@
...
+          resolve();
+        };
+        putRequest.onerror = () => {
+          reject(new Error(`Failed to set item with key '${key}'`));
+        };
+      } else {
+        reject(new Error('Object store not available'));
+      }
+    });
+  }
```
haydenfowler, line 172:
```suggestion
  async setItem(key: string, value: string): Promise<void> {
    const store = await this.getObjectStore('readwrite');

    return new Promise<void>(async (resolve, reject) => {
      if (store) {
        const putRequest = store.put(value, key);
        putRequest.onsuccess = () => {
          resolve();
        };
        putRequest.onerror = () => {
          reject(new Error(`Failed to set item with key '${key}'`));
        };
      } else {
        reject(new Error('Object store not available'));
      }
    });
  }
```

packages/passport/sdk/src/storage/IndexedDbStorage.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  private async getObjectStore(
+    mode: IDBTransactionMode = 'readonly',
+  ): Promise<IDBObjectStore | null> {
+    await this.initDatabase();
+    if (!this.db) {
+      throw new Error('IndexedDB database is not initialised');
+      return null;
```
haydenfowler, line 38: This will never be executed due to the `throw new Error` on the line above
```suggestion
```

packages/passport/sdk/src/storage/IndexedDbStorage.ts
```diff
@@ -0,0 +1,173 @@
+import { AsyncStorage } from 'oidc-client-ts';
+
+const databaseName = 'ImmutableSDK';
+const storeName = 'Passport';
+
+export class IndexedDbStorage implements AsyncStorage {
```
haydenfowler, line 6: Given that this class is used for maintaining user state, it's pretty crucial that this continues to work as expected. Can we please add some tests?

nattb8: It just uses `localForage` now. Do we need tests?

package.json
```diff
@@ -5,6 +5,7 @@
   "author": "Immutable",
   "bugs": "https://github.com/immutable/ts-immutable-sdk/issues",
   "dependencies": {
+    "localforage": "^1.10.0",
```
haydenfowler, line 8: This dependency should be in `/sdk/package.json` & `/packages/passport/sdk/package.json`

packages/passport/sdk/src/storage/IndexedDbStorage.ts
```diff
@@ -0,0 +1,37 @@
+import { AsyncStorage } from 'oidc-client-ts';
+import localForage from 'localforage';
+
+export class IndexedDbStorage implements AsyncStorage {
```
haydenfowler, line 4: Can we update the name to indicate that this is a `localForage` wrapper please? Maybe `LocalForageAsyncStorage`?

haydenfowler: Following the same train of thought, can we either:

- include IndexedDB in the name, OR
- pass the driver (and the name) as an argument to the constructor

haydenfowler (approved):

YermekG (approved):

PR #1981, opened 2024-07-18
feat: Add settlementType to Segment events

sharif9876:
# Summary

Add the settlementType of SUFFICIENT or SWAP into the Segment events for ProceedToPay in the Checkout Sales Widget

https://immutable.atlassian.net/browse/CM-775


jhesgodi (approved):
Awesome, thanks!

PR #1984, opened 2024-07-18
ID-1986 feat: make linkWallet method accessible

aerim-yi:
# Summary
This PR makes the linkWallet method accessible and also updates the sample app.

Also regenerated the mr clients.

## Added 

- added `linkExternalWallet` method to Passport that calls the link-wallet-v2 public API ([reference](https://docs.immutable.com/api/zkevm/reference/#/operations/link-wallet-v2))

- ### 200 Status Code (Success):

https://github.com/user-attachments/assets/69625540-6056-4f3d-a24c-f3e66c0e0722

- ### 400 & 500 Status Code (Error handling when messages are available)

https://github.com/user-attachments/assets/2fd7c17c-d3fe-4c5e-959c-19ee4b8c0a2b


naveen-imtb (approved):

hiep-immutable (dismissed):

packages/passport/sdk-sample-app/src/components/LinkWallet.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const handleSubmit = async (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    e.stopPropagation();
+
+    const form = e.currentTarget;
+    if (form.checkValidity()) {
+      setInvalid(false);
+      setIsSubmitting(true);
+      try {
+        const nonce = generateNonce();
```
dom-murray, line 40: The nonce will need to be provided as input as well. It's in the typed data that gets signed so it will be in the signature.

matthewmuscat (dismissed):

zaidarain1 (approved):

zaidarain1 (approved):

dom-murray (approved):

packages/passport/sdk/src/Passport.ts
```diff
@@ -299,4 +297,31 @@ export class Passport {
     const getUserInfoResult = await this.multiRollupApiClients.passportProfileApi.getUserInfo({ headers });
     return getUserInfoResult.data.linked_addresses;
   }
+
+  public async linkExternalWallet(
+    type: string,
+    walletAddress: string,
+    signature: string,
+    nonce: string,
+  ): Promise<LinkedWallet> {
```
haydenfowler, line 306: Can we update this function to accept a single object as an argument, rather than 4 different arguments please?

packages/passport/sdk/src/Passport.ts
```diff
@@ -299,4 +297,26 @@ export class Passport {
...
+      signature: params.signature,
+      nonce: params.nonce,
+    };
+    const linkWalletV2Result = await this.multiRollupApiClients
+      .passportProfileApi
+      .linkWalletV2(
+        { linkWalletV2Request },
+        { headers },
+      );
+    return { ...linkWalletV2Result.data };
```
haydenfowler, line 320: Should we check if the user is registered with zkEvm or IMX before allowing the request to be sent?
And how are errors handled here? What type of error is thrown from this function if the request fails? If it's an Axios error, then we probably want to wrap it and handle accordingly.

packages/passport/sdk/src/Passport.test.ts
```diff
@@ -371,23 +377,45 @@ describe('Passport', () => {
...
+
+    it('should throw error if user is not registered with IMX or ZkEvm', async () => {
+      getUserMock.mockReturnValue(mockUser);
+
+      await expect(passport.linkExternalWallet(linkWalletParams)).rejects.toThrow(
+        new PassportError('User has not been registered', PassportErrorType.USER_NOT_REGISTERED_ERROR),
+      );
+    });
+
+    it('should handle errors from the linkWalletV2 API call', async () => {
```
dom-murray, line 412: Can we add a couple tests to check that the status code and body of the response are still accessible here? E.g. the end user will need to be able to handle a few different error codes for 400s: (ALREADY_LINKED, MAX_WALLETS_LINKED, DUPLICATE_NONCE, VALIDATION_ERROR).

image

packages/passport/sdk/src/Passport.ts
```diff
@@ -299,4 +302,55 @@ export class Passport {
...
+    track('passport', 'linkWallet', { type: params.type });
+
+    const user = await this.authManager.getUser();
+    if (!user) {
+      throw new PassportError('User is not logged in', PassportErrorType.NOT_LOGGED_IN_ERROR);
+    }
+
+    const isRegisteredWithIMX = isUserImx(user);
+    const isRegisteredWithZkEvm = isUserZkEvm(user);
+    if (!isRegisteredWithIMX && !isRegisteredWithZkEvm) {
```
dom-murray, line 316: Was this intentionally changed back to &&? With the focus on zkEVM I think a lot of users will end up not being registered for IMX.

haydenfowler:
The sample app seems to get stuck in a loading state after successfully linking a wallet (note the output and the loading spinner in the top right):
image

packages/passport/sdk/src/Passport.ts
```diff
@@ -299,4 +302,55 @@ export class Passport {
...
+              throw new PassportError(message, PassportErrorType.LINK_WALLET_DUPLICATE_NONCE_ERROR);
+            case 'VALIDATION_ERROR':
+              throw new PassportError(message, PassportErrorType.LINK_WALLET_VALIDATION_ERROR);
+            default:
+              throw new PassportError(message, PassportErrorType.LINK_WALLET_GENERIC_ERROR);
+          }
+        } else {
+          // Handle unexpected bad request with a generic error message
+          throw new PassportError('LINK_WALLET error', PassportErrorType.LINK_WALLET_GENERIC_ERROR);
+        }
```
haydenfowler, line 351: Can we also use the `trackError` function from `@imtbl/metrics` when an error is thrown please? Should help us keep track of any issues in the future

packages/passport/sdk/src/Passport.ts
```diff
@@ -299,4 +302,55 @@ export class Passport {
...
+      signature: params.signature,
+      nonce: params.nonce,
+    };
+
+    return await withPassportError(async () => {
+      const linkWalletV2Result = await this.multiRollupApiClients
+        .passportProfileApi.linkWalletV2({ linkWalletV2Request }, { headers });
+
+      // Throw PassportError for Bad Request.
+      if (linkWalletV2Result.status === 400) {
```
haydenfowler, line 333: According to [the docs](https://docs.immutable.com/api/zkevm/reference/#/operations/link-wallet-v2) this endpoint can return a `400`, `401`, `403`, or a `500`. Can we invert the logic so that we're handling these status codes please? It may also be good to log the status code if it's something other than an APIError, e.g:
```
if (linkWalletV2Result.status === 200) {
  return linkWalletV2Result.data;
} else {
  if (isAPIError(linkWalletV2Result.data)) {
    const { code, message } = linkWalletV2Result.data;
    switch (code) {
      ...
    }
  } else {
    throw new PassportError(`Link wallet request failed with status code ${linkWalletV2Result.status}`, PassportErrorType.LINK_WALLET_GENERIC_ERROR);
  }
}
```

packages/passport/sdk/src/Passport.ts
```diff
@@ -299,4 +303,58 @@ export class Passport {
...
+            throw new PassportError(message, PassportErrorType.LINK_WALLET_VALIDATION_ERROR);
+          default:
+            throw new PassportError(message, PassportErrorType.LINK_WALLET_GENERIC_ERROR);
+        }
+      } else {
+        // Handle unexpected error with a generic error message
+        throw new PassportError(
+          `Link wallet request failed with status code ${(error as any).response.status}`,
+          PassportErrorType.LINK_WALLET_GENERIC_ERROR,
+        );
```
haydenfowler, line 356: We should avoid using `any` unless we have no alternative. It also looks like this will fail if an error is thrown that doesn't have `error.response.status`. You can reproduce this by initiating a wallet link in the sample app, right click the request in dev tools and click `Block request URL`, and then attempt to link a wallet again - you'll get a `Cannot read properties of undefined (reading 'status') type: undefined` error. 

I think we can fix both of these issues by structuring the logic as follows:
```
      if (isAxiosError(error) && error.response) {
        if (error.response.data && isAPIError(error.response.data)) {
          const { code, message } = error.response.data;

          switch (code) {
            case 'ALREADY_LINKED':
              throw new PassportError(message, PassportErrorType.LINK_WALLET_ALREADY_LINKED_ERROR);
            case 'MAX_WALLETS_LINKED':
              throw new PassportError(message, PassportErrorType.LINK_WALLET_MAX_WALLETS_LINKED_ERROR);
            case 'DUPLICATE_NONCE':
              throw new PassportError(message, PassportErrorType.LINK_WALLET_DUPLICATE_NONCE_ERROR);
            case 'VALIDATION_ERROR':
              throw new PassportError(message, PassportErrorType.LINK_WALLET_VALIDATION_ERROR);
            default:
              throw new PassportError(message, PassportErrorType.LINK_WALLET_GENERIC_ERROR);
          }
        } else if (error.response.status) {
          throw new PassportError(
            `Link wallet request failed with status code ${error.response.status}}`,
            PassportErrorType.LINK_WALLET_GENERIC_ERROR,
          );
        }
      }

      let message: string = 'Link wallet request failed';
      if (error instanceof Error) {
        message += `: ${error.message}`;
      }

      throw new PassportError(
        `Link wallet request failed: ${message}`,
        PassportErrorType.LINK_WALLET_GENERIC_ERROR,
      );
```

haydenfowler (approved):

dom-murray (approved):

PR #1986, opened 2024-07-19
fix dev mod for passport sample sdk

shineli1984:

# Summary

## Security

### DIFF
--- a/dev.sh
+++ b/dev.sh
@@ -22,7 +22,7 @@ else
 fi
 
 if echo "$PACKAGE_NAME" | grep -q "sample"; then
-  echo -e "You are targetting a sample app. Dev mode only supports SDK packages. \nPlease run development script in your sample app instead."
+  echo -e "You are targeting a sample app. Dev mode only supports SDK packages. \nPlease run development script in your sample app instead."
   exit 1
 fi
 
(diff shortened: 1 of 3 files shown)


packages/passport/sdk-sample-app/package.json
```diff
@@ -34,8 +34,8 @@
   "private": true,
   "scripts": {
     "build": "next build",
-    "d": "next dev",
-    "dev-with-sdk": "concurrently 'yarn workspace @imtbl/passport-sdk-sample-app dev' 'yarn workspace @imtbl/passport dev'",
+    "dev": "next dev",
```
matthewmuscat, line 37: Just double check for me but I think the `dev` script in `passport/sdk/package.json` is also `d` by mistake as well?

dev.sh
```diff
@@ -21,6 +21,11 @@ else
   PACKAGE_NAME=$1
 fi
 
+if echo "$PACKAGE_NAME" | grep -q "sample"; then
+  echo -e "You are targetting a sample app. Dev mode only supports SDK packages. \nPlease run development script in your sample app instead."
```
matthewmuscat, line 25: targeting

packages/passport/sdk-sample-app/package.json
shineli1984: yes. because "next dev" is itself a watch script already.

matthewmuscat (approved):

PR #1990, opened 2024-07-22
feat: [EXC-23] CICD and playwright for nextjs passport example

zaidarain1:


.github/workflows/build-and-test-examples.yaml
```diff
@@ -0,0 +1,54 @@
...
+        uses: actions/checkout@v4
+
+      - name: Use Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: ".nvmrc"
+          registry-url: https://registry.npmjs.org/
+          cache: "yarn"
+
+      - name: Find all projects
```
andrew-yangy, line 26: I feel like it can be relatively easier done by using yarn workspace cmd, or wsrun, or nx Shine has been working on.
- https://classic.yarnpkg.com/lang/en/docs/cli/workspaces/#toc-yarn-workspaces-run
- https://github.com/hfour/wsrun
- https://nx.dev/nx-api/nx/documents/run

zaidarain1: The problem I found was that these are separate apps, not a part of the root workspace, so yarn workspaces won't work here. NX might also need a bit more of a set up here

That's also why these have their own separate yarn.lock file too. I tried to use `yarn --cwd examples/../.. build` but it would just build in the root directory, seems like that's buggy and doesn't work correctly, at least in the current version of yarn we're using. It was fixed apparently according to an issue 3 weeks ago.

shineli1984 (dismissed):

shineli1984 (approved):

PR #1993, opened 2024-07-23
feat:[CM-708] Checkout Widget Setup

jwhardwick:

# Summary

## Added 

* Adds Checkout Widget to `widgets-lib`, will load the Checkout App iFrame
* Only create Widget on SANDBOX
* Adds to widgets-sample-app


packages/checkout/widgets-lib/src/factory.ts
```diff
@@ -92,6 +94,15 @@ export class WidgetsFactory implements IWidgetsFactory {
           provider,
         }) as Widget<WidgetType.SALE> as Widget<T>;
       }
+      case WidgetType.CHECKOUT: {
+        if (this.sdk.config.environment === Environment.SANDBOX) {
```
jhesgodi, line 98: @jwhardwick can we please reverse the logic here to only throw if env is prod, because we have to allow dev deployments as well.

packages/checkout/widgets-lib/src/widgets/checkout/CheckoutWidget.tsx
```diff
@@ -0,0 +1,82 @@
...
+} from '../../context/view-context/ViewContext';
+import { StrongCheckoutWidgetsConfig } from '../../lib/withDefaultWidgetConfig';
+import { ErrorView } from '../../views/error/ErrorView';
+import { LoadingView } from '../../views/loading/LoadingView';
+
+export type CheckoutWidgetInputs = CheckoutWidgetParams & {
+  config: StrongCheckoutWidgetsConfig,
+};
+
+export default function CheckoutWidget(props: CheckoutWidgetInputs) {
```
jhesgodi, line 27: @jwhardwick lets clean up this component to be empty, since this will just load the iframe, dont require any of the existing providers.

Elements we can keep are:
- config props
- checkout and provider instances

and just have it render an empty div.

jwhardwick:
Thanks @jhesgodi comments addressed.

jhesgodi (approved):

PR #2033, opened 2024-07-30
fix: [DX-3046] Enhance game bridge logging

CodeSchwert:

# Summary

Send tx data with track duration in transfer and transaction calls for better logging.

Don't minify the game bridge build artefact to help debugging.

Should help debug transfer and transaction errors from our end.

## Added 
- Additional logging to game bridge transfer and transaction methods.


YermekG (approved):

packages/game-bridge/package.json
```diff
@@ -14,8 +14,8 @@
     "parcel": "^2.8.3"
   },
   "scripts": {
-    "build": "parcel build --no-cache --no-scope-hoist && yarn updateSdkVersion",
     "d": "swc src -d dist --strip-leading-paths --ignore '**/*.test.*'",
+    "build": "parcel build --no-cache --no-scope-hoist --no-optimize  && yarn updateSdkVersion",
```
wdinau, line 18: could you help me with what does no-optimize solve?  and perhaps the double space is a mistake here?

wdinau (commented):
Nothing against current logs. Through troubleshooting with ILV, i found we have no trace of `RequestID` from the system (NR log) anywhere.
Imagining we are troubleshoot with ILV again, they claim no call came in to this function. Is there way to attach a few more logs or tracking info near the batchTrasnfer fucntion?

packages/game-bridge/package.json
CodeSchwert: `--no-optimize` will stop not minify the game bridge build artefact so it'll be easier to see which bit of code triggered the error in the stack trace. Otherwise the JS output is just one giant line making it harder to parse.

wdinau (approved):

PR #2043, opened 2024-07-31
feat: track passport starkex transfer errors

CodeSchwert:

# Summary
Use the metrics package to track starkex transfer errors in the Passport module.

The change doesn't change the behaviour of the StarkEx transfer methods, just adds additional error tracking.

## Added 
- metrics package error tracking for Passport starkEx transfer errors.


packages/passport/sdk/src/errors/passportError.ts
```diff
@@ -52,6 +53,9 @@ export const withPassportError = async <T>(
       errorMessage = (error as Error).message;
     }
 
-    throw new PassportError(errorMessage, customErrorType);
+    const passportError = new PassportError(errorMessage, customErrorType);
+    trackError('passport', customErrorType, passportError);
```
haydenfowler, line 57: I don't believe we should be using `customErrorType` as the second argument. `trackError` accepts a `eventName` as the second argument, and it also prepends `trackError_` to the `eventName`, so in the context of transfer this should probably be something like `imxTransfer` to produce an event called `trackError_imxTransfer`

CodeSchwert: Yes, shows up like this in the query results: `metrics_eventName:trackErrorTRANSFERERROR`. I'll give it a better event name

packages/passport/sdk/src/errors/handleTrackError.ts
```diff
@@ -0,0 +1,10 @@
+import { trackError } from '@imtbl/metrics';
+
+const moduleName = 'passport';
+
+export const handleError = (method: string, error: unknown) => {
+  const wrappedError = new Error(`${moduleName} ${method} error`, { cause: error });
+
+  trackError('passport', method, wrappedError);
```
imx-mikhala, line 8: Given this is called within a `withPassportError` arent we already calling `trackError` there?

packages/passport/sdk/src/errors/handleTrackError.ts
```diff
@@ -0,0 +1,10 @@
+import { trackError } from '@imtbl/metrics';
+
+const moduleName = 'passport';
+
+export const handleError = (method: string, error: unknown) => {
```
imx-mikhala, line 5: nit: given this is used as `throw handleError(...)` it might read better if this was named wrappedError so it reads as `throwWrappedError` or similar

packages/passport/sdk/src/errors/handleTrackError.ts
CodeSchwert: I put it in both places to see which would be preferable. If Passport team is ok with modifying the `withPassportError` function, then I'll remove the `handleError` function from the transfer methods.

imx-mikhala: Ah ok, I think it makes sense to keep it all contained within `withPassportError` since that does all the error handling already

packages/passport/sdk/src/errors/passportError.ts
```diff
@@ -52,6 +54,9 @@ export const withPassportError = async <T>(
       errorMessage = (error as Error).message;
     }
 
-    throw new PassportError(errorMessage, customErrorType);
+    const passportError = new PassportError(errorMessage, customErrorType);
+    trackError('passport', customEventName || '', passportError);
```
haydenfowler, line 58: It looks like there are 15 places that call `withPassportError` without specifying a `customEventName`. Can we either:
- make `customEventName` mandatory, OR
- only call `trackError` if `customEventName` is set

That way we won't end up with a whole barrage of `passport.trackError_` logs.

haydenfowler (approved):

imx-mikhala (approved):

packages/passport/sdk/src/errors/passportError.ts
haydenfowler: Have logged [ID-2222](https://immutable.atlassian.net/browse/ID-2222) for the Passport team to action at a later date.

PR #2055, opened 2024-08-02
[SRE-2556] Remove gitleaks

dhf22:
Removing use of Gitleaks action as it has been monetised and now requires license. Secret scanning is now done at an org level.


package.json
zaidarain1, line 1: Could the scan:secrets from line 40 under lint-staged also be removed since that command isn't valid anymore?

zaidarain1 (changes requested):

zaidarain1 (approved):

zaidarain1 (approved):

PR #2056, opened 2024-08-02
[NO-CHANGELOG][Checkout Widget] Update Checkout widget to use post message handler

jiyounglee:

# Summary

## Security


packages/checkout/sdk/src/postMessageHandler/postMessageHandler.ts
```diff
@@ -6,6 +6,7 @@ export type PostMessageHandlerConfiguration = {
 // todo put these in a types file
 export enum PostMessageHandlerEventType {
   PROVIDER_RELAY = 'PROVIDER_RELAY',
```
jhesgodi, line 8:
```suggestion
IMTBL_PROVIDER_RELAY
```

jhesgodi: since these are post message events, I think we should prefix them all to not collide with other apps

jhesgodi (dismissed):

jhesgodi (approved):

PR #2068, opened 2024-08-05
:feat add passport react context and hooks

shineli1984:
# Summary
Add passport react context and hooks to reduce the code user need to write in a react app.

## Added 
Passport context and hooks for react app


packages/passport/sdk/src/react/PassportContext.tsx
```diff
@@ -0,0 +1,242 @@
...
+type PassportProviderProps = {
+  config: PassportModuleConfiguration;
+  children: React.ReactNode;
+};
+
+export function PassportProvider({ children, config }: PassportProviderProps) {
+  const [isLoggedIn, setLoggedIn] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
+  const [accounts, setAccounts] = useState<string[]>([]);
+  const [error, setError] = useState<Error | null>(null);
```
keithbro-imx, line 43: Hmmm no... I don't think you want to say this is Error or null when you're putting `any` in to it.

packages/passport/sdk/src/react/PassportContext.tsx
```diff
@@ -0,0 +1,242 @@
...
+  const [idToken, setIdToken] = useState<string | null>(null);
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState<Error | null>(null);
+
+  const getIdToken = () => {
+    if (isLoggedIn) {
+      setLoading(true);
+      passportInstance
+        .getIdToken()
+        .then((t) => setIdToken(t || null))
```
keithbro-imx, line 138: I think we'd want to handle `t` being undefined here in some way. What would cause this? And what would we expect to get back from `useIdToken` in this scenario?

packages/passport/sdk/src/react/PassportContext.tsx
```diff
@@ -0,0 +1,242 @@
...
+import { Passport } from '../Passport';
+import { PassportModuleConfiguration, UserProfile } from '../types';
+
+type PassportContextType = {
+  passportInstance: Passport;
+  isLoggedIn: boolean;
+  login: () => Promise<string[]>;
+  logout: () => Promise<void>;
+  loginWithoutWallet: () => Promise<UserProfile | null>;
+  loginWithEthersjs: () => Promise<{ accounts: string[], provider: Web3Provider | null }>;
```
keithbro-imx, line 16: When would provider be null here?

keithbro-imx (commented):
Some initial comments, but i'll try this in a real app and provide feedback

packages/passport/sdk/src/react/PassportContext.tsx
```diff
@@ -0,0 +1,242 @@
...
+  isLoading: boolean;
+  error: Error | null;
+  accounts: string[];
+};
+
+const PassportContext = createContext<PassportContextType>({
```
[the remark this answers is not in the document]
shineli1984: yeah. I can add that

packages/passport/sdk/src/react/PassportContext.tsx
shineli1984: this is from calling oidcClient on this line: https://github.com/immutable/ts-immutable-sdk/blob/f4f6eee9ab99b61a2923f1625ada30d565098af6/packages/passport/sdk/src/authManager.ts#L525
So we don't really have control over.

packages/passport/sdk/src/react/PassportContext.tsx
shineli1984: I'll move provider onto the context but it's null when `provider.send('eth_requestAccounts', [])` rejects.

packages/passport/sdk/src/react/PassportContext.tsx
```diff
@@ -0,0 +1,242 @@
...
+        setLoggedIn(true);
+        return profile;
+      } catch (e: any) {
+        setError(e);
+        return null;
+      } finally {
+        setIsLoading(false);
+      }
+    },
+    loginWithEthersjs: async () => {
```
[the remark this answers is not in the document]
shineli1984: Passport sdk package already has `@ethersproject/providers` as an dependency tho. Should I change it to peer?

packages/passport/sdk/src/react/PassportContext.tsx
```diff
@@ -0,0 +1,242 @@
...
+  const [accounts, setAccounts] = useState<string[]>([]);
+  const [error, setError] = useState<Error | null>(null);
+  const passportInstance = new Passport(config);
+  const v = useMemo(() => ({
+    passportInstance,
+    isLoggedIn,
+    isLoading,
+    accounts,
+    error,
+    login: async () => {
```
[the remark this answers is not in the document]
shineli1984: I'm going with our documentation to separate 
- login: https://docs.immutable.com/products/zkEVM/passport/identity/login#2-initialise-the-provider-and-login-users
- and login without wallet: https://docs.immutable.com/products/zkEVM/passport/identity/login#2-initialise-the-provider-and-login-users

I assume we want user to use the first one? what are you suggesting? name this function as connectEVM and the other one as login?

packages/passport/sdk/src/react/PassportContext.tsx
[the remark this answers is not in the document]
shineli1984: I'll do this.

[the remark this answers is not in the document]
shineli1984: Added

packages/passport/sdk/src/react/PassportContext.tsx
keithbro-imx: Hmm but in what scenario would this happen? If `isLoggedIn` is true and `getIdToken` returns undefined, should that set an error for example? Consider the result of useIdToken - it would be `{ idToken: null, error: null, loading: false }` - not helpful

packages/passport/sdk/src/react/PassportContext.tsx
```diff
@@ -0,0 +1,274 @@
...
+};
+
+const PassportContext = createContext<PassportContextType | null>(null);
+
+type PassportProviderProps = {
+  config: PassportModuleConfiguration;
+  children: React.ReactNode;
+};
+
+export function PassportProvider({ children, config }: PassportProviderProps) {
```
keithbro-imx, line 33: I'd recommend calling this something like ReactProvider to avoid confusion with the provider on line 17

packages/passport/sdk/src/react/PassportContext.tsx
shineli1984: casted to Error in catch statements

packages/passport/sdk/src/react/PassportContext.tsx
keithbro-imx, line 33: It would be useful if we could pass in the passport instance, rather than the config.

shineli1984: Considering the usage side, maybe `PassportReactProvider`?

keithbro-imx: Normally you're going to be doing

```
import { passport } from '@imtbl/sdk';
```

so the double "passport" in

```
passport.PassportReactProvider
```

would be unnecessary IMO

keithbro-imx: Reason: when using checkout, you want to be able to pass the passport instance to checkout. Obviously you can still get it by calling `usePassport` but it makes it more tricky as you would want to initialise checkout and passport together

packages/passport/sdk/src/react/PassportContext.tsx
shineli1984: This should be equivalent to an internal error / bug scenario. I can set the error in this case. however, it complex things a bit more because now we need to consider if we need to clear up all the other fields.

packages/passport/sdk/src/react/PassportContext.tsx
```diff
@@ -0,0 +1,279 @@
...
+  isLoading: boolean;
+  error: Error | null;
+  accounts: string[];
+  passportProvider: Provider | null;
+  web3Provider: Web3Provider | null;
+  login: () => Promise<string[]>;
+  logout: () => Promise<void>;
+  loginWithoutWallet: () => Promise<UserProfile | null>;
+  loginWithEthersjs: () => Promise<string[]>;
+  loginCallback: () => Promise<void>;
```
haydenfowler, line 23: Can we have a separate hook for generic Passport functionality (everything in `Passport.ts`), and zkEvm functionality? For a consumer, it's not clear if `accounts` relates to IMX or zkEvm

packages/passport/sdk/src/react/PassportContext.tsx
```diff
@@ -0,0 +1,279 @@
+/* eslint-disable @typescript-eslint/naming-convention */
```
haydenfowler, line 1: There's quite a bit of functionality within this module - can we please add some tests?

packages/passport/sdk/src/react/PassportContext.tsx
shineli1984: I'm also confused. I have this screenshot attached. They are methods on Passport. Are you saying I should have separate Providers, one for imx and one for zkevm?
image

packages/passport/sdk/src/react/PassportContext.tsx
shineli1984: I'll need to address all the feedback first.

packages/passport/sdk/src/react/PassportContext.tsx
shineli1984: Now I think about it. I should just do this for zkevm.

haydenfowler: I think ideally we'd have three: one for generic Passport (everything in `Passport.ts` except `connectImx` & `connectEvm`), one for exposing IMX functionality, and one for zkEVM. I agree that we can probably disregard the IMX provider

packages/passport/sdk/src/react/PassportContext.tsx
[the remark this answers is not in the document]
haydenfowler: We use some of the modules from Ethers internally, but we do not expose Ethers to consumers of Passport. By exposing Ethers from Passport like this, we'll need to continue to support it indefinitely, or make a breaking change if we decide to remove it.

This could also open up the door for partners to request that we add methods for other wallet connection libraries (e.g. Viem), so I think it's best if we just expose the raw Passport provider and let consumers wrap it themselves.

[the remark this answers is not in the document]
shineli1984: I'll try remove this and see what kind of experience it is with 3rd party provider.

packages/passport/sdk/src/react/PassportContext.tsx
```diff
@@ -0,0 +1,244 @@
...
+
+export function usePassportProvider() {
+  const { passportProvider, isLoading, error } = usePassport();
+  return { passportProvider, isLoading, error };
+}
+
+export function useProfile() {
+  const { profile, isLoading, error } = usePassport();
+  return { profile, isLoading, error };
+}
```
imx-mikhala, line 244: These three hooks seem redundant if they can just get them from `usePassport`

packages/passport/sdk/src/react/PassportContext.tsx
```diff
@@ -0,0 +1,244 @@
...
+      } catch (e) {
+        setError(e as Error);
+      } finally {
+        setIsLoading(false);
+      }
+    },
+    loginCallback: async () => {
+      await passportInstance.loginCallback();
+    },
+    linkExternalWallet: passportInstance.linkExternalWallet.bind(passportInstance),
```
haydenfowler, line 96: Is the `bind` call here necessary? It's not clear to me why we can't just call `linkExternalWallet` directly:
```suggestion
    linkExternalWallet: passportInstance.linkExternalWallet,
```

shineli1984: `linkExternalWallet` uses `this` in its function body tho. Without the bind, it'll work?

packages/passport/sdk/src/react/PassportContext.tsx
shineli1984: This is mostly for convenience.

packages/passport/sdk/src/react/PassportContext.tsx
```diff
@@ -0,0 +1,244 @@
...
+          const acc = await provider.request({ method: 'eth_requestAccounts' });
+          setLoggedIn(true);
+          setAccounts(acc);
+        }
+      } catch (e) {
+        setError(e as Error);
+      } finally {
+        setIsLoading(false);
+      }
+    },
```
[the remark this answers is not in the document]
shineli1984: I'm only supporting zkevm here. I'll change the component name to reflect on that.
In that case, I think it makes more sense to have a login for both auth + wallet, which is our preferred use case? Then treating auth without wallet as edge case.

packages/react/passport/package.json
zaidarain1, line 1: Regarding the package.json:
- Could the "d" command be updated to be a rollup config build without the NODE_ENV=production, and have the rollup.config.js to conditionally use typescript or swc when rebuilding to match the change that was merged in yesterday? 
- The unplugin-swc dependency will then need to be added to this package
- The exports object needs to be set up for the conditional types for live types locally, matching the other package.jsons that were updated
- I'll drop the tsconfig change needed here; the customConditions value also needs to be added to support the live types.

You can use the passport sdk package.json/rollup.config/tsconfig.json as a point of reference

zaidarain1 (changes requested):

packages/react/passport/src/PassportContext.tsx
```diff
@@ -0,0 +1,187 @@
...
+import {
+  Passport, PassportModuleConfiguration,
+  UserProfile,
+} from '@imtbl/passport';
+
+type ZkEvmReactContext = {
+  passportInstance: Passport;
+  isLoggedIn: boolean;
+  isLoading: boolean;
+  withWallet: boolean;
```
haydenfowler, line 16: What's the use case for `withWallet`? If the consumer specifies whether they want the wallet initialised or not, then the consumer would already know if the wallet has been initialised

packages/react/passport/src/PassportContext.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+type ZkEvmReactProviderProps = {
+  config: PassportModuleConfiguration;
+  children: React.ReactNode;
+};
+
+export function ZkEvmReactProvider({ children, config }: ZkEvmReactProviderProps) {
+  setPassportClientId(config.clientId);
+  track('passport', 'ZkEvmProviderInitialised');
+  const [isLoggedIn, setLoggedIn] = useState(false);
```
haydenfowler, line 43: Can we update the setter to `setIsLoggedIn` please?

packages/react/passport/src/PassportContext.tsx
```diff
@@ -0,0 +1,187 @@
...
+  const [error, setError] = useState<Error | null>(null);
+  const [withWallet, setWithWallet] = useState(true);
+  const passportInstance = new Passport(config);
+
+  const checkLogggedIn = useCallback(async () => {
+    const p = await passportInstance.login({
+      useCachedSession: true,
+    });
+    setLoggedIn(!!p);
+  }, [passportInstance]);
```
haydenfowler, line 54: This has the potential to cause re-rendering issues for consumers. My assumption is that we expect marketplaces to consume the provider like this:

```
function MyComponent() {
  const { isLoading, isLoggedIn } = reactPassport.usePassport();
  
  if (isLoading) {
    return <LoadingAnimation />;
  }
  
  if (isLoggedIn) {
    return <SomeComponentForAuthenticatedUsers />;
  }

  return <PromptUserToLogin />;
}
```

In the above scenario, if `login` triggers a token refresh, `PromptUserToLogin` would render momentarily. Once the token refresh finishes and `login` returns true, `SomeComponentForAuthenticatedUsers` would then render.

packages/react/passport/src/PassportContext.tsx
```diff
@@ -0,0 +1,187 @@
...
+        const provider = passportInstance.connectEvm();
+        const accounts = await provider.request({ method: 'eth_requestAccounts' });
+        return accounts;
+      } catch (e: any) {
+        setError(e as Error);
+      } finally {
+        setIsLoading(false);
+      }
+      return [];
+    },
```
haydenfowler, line 146: `eth_requestAccounts` initialises the wallet, so if we're going to keep the `withWallet` property, then this should also call `setWithWallet`

packages/react/passport/src/PassportContext.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+type ZkEvmReactContext = {
+  passportInstance: Passport;
+  isLoggedIn: boolean;
+  isLoading: boolean;
+  withWallet: boolean;
+  error: Error | null;
+  getIdToken: Passport['getIdToken'];
+  getAccessToken: Passport['getAccessToken'];
+  getAccounts: () => Promise<string[]>;
```
haydenfowler, line 20: Can I suggest that we expose the connectEvm method instead of `getAccounts` like this? This interface would make it difficult to support something like app-chains in the future

PR #2081, opened 2024-08-07
bugfix: TD-1577 bulk listing support with multi pop ups for smart contract wallets.

naveen-imtb:

- [X] `feat:`, `fix:`, `refactor:`, `docs:`, or `chore:`.

# Summary

This PR fixes the bulk listings creation failure when using smart contract wallets.

## Deprecated
`completeListings(signature: string): Promise<BulkListingsResult>` in the `PrepareBulkListingsResponse` type is now superseded by `completeListings(signatures: string[]): Promise<BulkListingsResult>`

## Fixed
Failed bulk listings creation using smart contract wallets.

## Security

With this change for bulk listings creation using smart contract wallets, a user will be required to sign multiple listing confirmations (as many as the number of orders attempted to be created).


packages/orderbook/src/orderbook.ts
```diff
@@ -217,28 +228,95 @@ export class Orderbook {
...
+        new Date(),
+        listing.orderExpiry || Orderbook.defaultOrderExpiry(),
+      )));
+
+      const pendingApproval: string[] = [];
+      const actions = prepareListingResponses.flatMap((response) => {
+        // de-dupe approval transactions to ensure every contract has
+        // a maximum of 1 approval transaction
+        const dedupedActions: Action[] = [];
+        response.actions.forEach((action) => {
```
Sam-Jeston, line 250: idiomatically I think this would be better as a `for (const action of response.actions)` as its side-effecting

naveen-imtb: @Sam-Jeston eslint doesn't like updating the loop and shouts with,
`ESLint: iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations.(no-restricted-syntax)`

packages/orderbook/src/orderbook.ts
```diff
@@ -99,6 +100,11 @@ export class Orderbook {
     );
   }
 
+  // Default order expiry to 2 years from now
+  static defaultOrderExpiry(): Date {
+    return new Date(Date.now() + 1000 * 60 * 60 * 24 * 365 * 2);
```
frankisawesome, line 105: Is this what we've always used?

frankisawesome (approved):
Discussed offline - testing coverage unfortunately only achievable through e2e flows manually.

packages/orderbook/src/orderbook.ts
frankisawesome: i see that it is

naveen-imtb: Yep but it was hard coded in multiple places, I've consolidated it to be a static function.

CodeSchwert (approved):

PR #2084, opened 2024-08-09
docs: Examples/readme updates

proletesseract:
Updates to the readme for the code examples after dogfooding them with Passport.

Checkout the rendered markdown here: https://github.com/immutable/ts-immutable-sdk/blob/examples/readme-updates/examples/README.md


proletesseract:
Done another pass of this from top to bottom and fixed up a few things, happy with this now from my side.

examples/README.md
```diff
@@ -267,3 +278,73 @@ Add the test runner to the scripts in your package.json
...
+
+In your `docs` branch modify the file `/src/remark/import-code.mjs` and update the `BRANCH` constant from `main` to the name of your branch on `ts-immutable-sdk` e.g.
+
+```
+const BRANCH = 'DVR-193-passport-signing-example';
+```
+
+Now your `docs` branch will be pulling the code examples from your branch in `ts-immutable-sdk` and you can use them locally in your `docs` branch to make sure they make sense in the page.
+
+Make sure to change this back to `main` before merging your `docs` branch or it will break the docs site.
```
matthewmuscat, line 301: Might make sense to bold this or make it a little more obvious as it is a disclaimer

examples/README.md
```diff
@@ -267,3 +278,73 @@ Add the test runner to the scripts in your package.json
...
+
+## Creating code snippets
+
+In your code examples in `tsimmutable-sdk` find the parts of the code you want to use as code snippets and wrap them in the `#doc` and `#enddoc` comments while providing a label.
```
matthewmuscat, line 305:
```suggestion
In your code examples in `ts-immutable-sdk` find the parts of the code you want to use as code snippets and wrap them in the `#doc` and `#enddoc` comments while providing a label.
```

examples/README.md
```diff
@@ -267,3 +278,73 @@ Add the test runner to the scripts in your package.json
...
+```
+````
+Or if you want to display the whole file just don't include a `#` label at the end of the file reference e.g.
+
+````
+```tsx reference=examples/passport/wallets-connect-with-nextjs/src/app/connect-with-eip1193/page.tsx title="Create the Passport Provider"
+```
+````
+
+Just like regular code snippets, you can set language for syntax highlighting, in this example we are setting it to `tsx`.
```
matthewmuscat, line 334: It's not clear to me what languages this supports out of the box. Is there a link we can refer to or add another row in the table below? Not a biggie though as I don't imagine this edge case will be encountered often

matthewmuscat (approved):

examples/README.md
```diff
@@ -44,18 +96,55 @@ examples
...
-## Creating a New Example
+### Setup example app
 
-create the nextjs project
+If you want to make a new NextJS app from scratch this is the process;
 
+Create a branch in `ts-immutable-sdk` to add your example to.
+
+In the console navigate to the product directory where your example will live (or create one if it doesn't exist). Then use `yarn dlx` to create the app.
 ```
```
zaidarain1, line 142: Can we add syntax highlighting to all the codeblocks? For eg here we could do ```bash, and below in the code snippets for the config files for eg we could add ```ts/```js

zaidarain1 (changes requested):

examples/README.md
```diff
@@ -267,3 +279,89 @@ Add the test runner to the scripts in your package.json
...
+
+## Creating code snippets
+
+In your code examples in `ts-immutable-sdk` find the parts of the code you want to use as code snippets and wrap them in the `#doc` and `#enddoc` comments while providing a label.
+
+Labels have to be unique within a file so you should use a simple naming convention to avoid clashes within the file e.g.
+
+e.g. `<library>-<tag>`
+
+```
```
zaidarain1, line 311: Can we add syntax highlighting to all the codeblocks? For eg here and in similar places we could do ```ts/```js, and in other parts we could also add the ```bash syntax highlighting.

zaidarain1 (changes requested):

examples/README.md
```diff
@@ -44,18 +96,55 @@ examples
...
+
+## Creating a new example app
+
+The process may differ if you're using a different Javascript Framework, but this assumes you will be creating a new example app using NextJS. These steps which were implemented to create the [Passport Connect](/examples/passport/wallets-connect-with-nextjs) example. It is a good reference point in terms of route structure and page layouts. If you need to copy code from there to help with your layouts then please go ahead.
 
-## Creating a New Example
+### Setup example app
 
-create the nextjs project
+If you want to make a new NextJS app from scratch this is the process;
```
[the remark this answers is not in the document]
proletesseract: We can, i don't think it should block this PR though. Perhaps its something we add when im rotating around the teams if they want to create a sample app using vite?

examples/README.md
```diff
@@ -1,15 +1,67 @@
...
+The example apps in this examples directory are compiled and tested as part of our CI CD pipeline. The goal of this is to ensure the examples found here are always able to run against the latest version of the Immutable Typescript SDK. 
+
+Selected portions of code from these example apps is then displayed as code snippets in our docs site, which means our code snippets in the docs are inherently always accurate as well. How to include the code from here in the docs site is explained below. 
+
+These example apps can also be used directly as a reference and run locally to test and develop with.
+
+# Running examples locally
+
+First you need to clone the `ts-immutable-sdk` repo. There is no need to install the root node modules or build the SDK since the examples have their own `package.json` files and pull the node modules from npm rather than the local build.
```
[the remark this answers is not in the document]
proletesseract: all good. i'll re word this. the reason why i thought that was probably because i alraeady had the other dependencies installed so it was fast, but also when i run `yarn install` it seems to create node_modules folders inside the identity-with-nextjs and wallets-transactions-with-nextjs which i would not have expected if they were being installed to the root.

examples/README.md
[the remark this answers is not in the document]
zaidarain1: Yeap makes sense, agreed.

examples/README.md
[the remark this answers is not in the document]
zaidarain1: I think it will still create node_modules to the apps if those dependencies are used only in that app and no where else

proletesseract:
@zaidarain1 ive applied your feedback, LMK if i missed anything

zaidarain1 (approved):

PR #2087, opened 2024-08-13
feat: SMR-3051 return withdrawal state

wcgcyx:
# Summary

Make bridge fee estimation to return additional withdrawal state.

## Added 

Return the following additional fields in `getUnsignedBridgeBundledTx`:
```
{
  ...
  delayWithdrawalLargeAmount: boolean | null;  // If withdrawal gets queued due to large amount.
  delayWithdrawalUnknownToken: boolean | null;  // If withdrawal gets queued due to unknown token.
  withdrawalQueueActivated: boolean | null;  // If withdrawal gets queued due to activated queue.
  largeTransferThresholds: number | null;  // The configured large transfer threshold for given withdrawal token.
}
```


packages/internal/bridge/sdk/src/lib/tenderly.ts
```diff
@@ -117,8 +121,52 @@ export async function submitTenderlySimulations(
...
     }
-    gas.push(simResults[i].gasUsed);
+    if (simResults[i].logs === undefined) {
+      throw new BridgeError(
+        'Estimating gas did not return logs',
+        BridgeErrorType.TENDERLY_GAS_ESTIMATE_FAILED,
+      );
+    }
+    // Attempt to extract event.
+    for (let j = 0; j < simResults[i].logs.length; j++) {
```
ermyas, line 131: would this be unnecessary work when estimates are for deposits? if so, should this be moved out from this function and performed by the calling fn, which would have the right context.

ermyas (dismissed):

Benjimmutable (approved):

PR #2100, opened 2024-08-20
fix: [DX-3140] Fix react is not defined widgets sample app error

zaidarain1:


jhesgodi (approved):

jwhardwick (commented):
Thanks @zaidarain1, that fixes the issue we had

PR #2102, opened 2024-08-20
feat: Fetch token allowlists from verified tokens API

luads:
# Summary

The Checkout widgets have maintained an internal token allowlist historically. That list was used for filtering out tokens for Swaps, Bridging, and parts of balance checks including Smart Checkout. That list is maintained manually, making it difficult to scale as more and more tokens are added to Immutable zkEVM.

To help our widgets keep up to date with new tokens, we're replacing that list with our Tokens API's [dynamic list](https://api.immutable.com/v1/chains/imtbl-zkevm-mainnet/tokens?verification_status=verified&is_canonical=true) of canonical tokens.

This PR encompasses:
- Replacing the existing allowlist logic with the tokens API endpoint.
- Extending balance filtering to zkEVM — previously any ERC20 token would be displayed, regardless of it being allow listed or not.
- A block list for DEX tokens, used by the Swap widgets. Some tokens may need to be disabled from swaps due to legal reasons or, cases like lack of liquidity pools.


packages/checkout/sdk/src/config/remoteTokensFetcher.ts
```diff
@@ -0,0 +1,177 @@
...
+type TokensEndpointResponse = {
+  result: TokensEndpointResult[];
+};
+
+export type RemoteConfigParams = {
+  isDevelopment: boolean;
+  isProduction: boolean;
+};
+
+export class RemoteTokensFetcher {
```
jhesgodi, line 41: It'll sound picky:

What should be a better name for this service? Maybe we can find out how we refer to this api across the company and use a name that matches?

jhesgodi: feel free to disregard

jwhardwick (approved):

jwhardwick (approved):

PR #2110, opened 2024-08-23
[ID-2334] Defer preload for magic in Passport SDK

matthewmuscat:
### ID-2334

# Summary
Magic SDK version we use [already handles preload when SDK is instantiated. ](https://github.com/magiclabs/magic-js/releases/tag/magic-sdk%4021.0.0)

We can remove the need for this.

## Removed
- Lazy load of Magic SDK client
- Magic SDK preload

## Security


packages/passport/sdk/src/magicAdapter.test.ts
```diff
@@ -75,15 +72,27 @@ describe('MagicWallet', () => {
...
+        const magicAdapter = new MagicAdapter(config);
+
+        await expect(async () => {
+          await magicAdapter.login(idToken);
+        }).rejects.toThrow(
+          new PassportError(
+            'Cannot perform this action outside of the browser',
+            PassportErrorType.WALLET_CONNECTION_ERROR,
+          ),
+        );
```
imx-mikhala, line 89: `rejects.toThrow` doesnt actually assert the PassportErrorType properly and can be anything and the test will pass. Instead need to do something like:

```
it('should throw a browser error for loginWithOIDC', async () => {
        const magicAdapter = new MagicAdapter(config);

        let type = '';
        let message = '';
        try {
          await magicAdapter.login(idToken);
        } catch (e: any) {
          type = e.type;
          message = e.message;
        }
        expect(type).toEqual(PassportErrorType.WALLET_CONNECTION_ERROR);
        expect(message).toEqual('Cannot perform this action outside of the browser');
      });
```

imx-mikhala (approved):

haydenfowler (approved):

PR #2122, opened 2024-08-29
feat: Adjust confirmation screen on free mints

luads:
# Summary

Primary Sales can work with free items, but the UI doesn't reflect it 100% correctly. This change will make that more obvious to avoid any confusion.

image

## Security


alejoloaiza (approved):
lgtm

jwhardwick (approved):

jhesgodi (approved):

PR #2126, opened 2024-08-29
feat: Add logout mode option to Game Bridge

CodeSchwert:

# Summary

Add the `logoutMode` option to be passed into the Game Bridge init function from game SDKs.

The new option is optional and only affects the Unreal and Unity SDKs.

## Added 
- Optional `logoutMode` option in game bridge `init` function

### DIFF
--- a/packages/game-bridge/src/index.ts
+++ b/packages/game-bridge/src/index.ts
@@ -220,6 +220,7 @@ window.callFunction = async (jsonData: string) => {
       case PASSPORT_FUNCTIONS.init: {
         const request = JSON.parse(data);
         const redirect: string | null = request?.redirectUri;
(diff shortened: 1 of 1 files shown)


nattb8 (approved):

packages/game-bridge/src/index.ts
```diff
@@ -231,6 +232,7 @@ window.callFunction = async (jsonData: string) => {
             redirectUri: redirect ?? redirectUri,
             logoutRedirectUri: request?.logoutRedirectUri,
             crossSdkBridgeEnabled: true,
+            logoutMode,
```
nattb8, line 235: If the logout mode is `silent` for device code auth, how does it clear the session from the browser without opening it?

CodeSchwert: Looks like if paired with HardLogout from the Game SDKs, that should force a Passport login to happen.

PR #2128, opened 2024-08-30
[NO CHANGELOG] fix: Error with NPM deploy attestations step

jhesgodi:
# Summary
Fix Error: Failed to get ID token: Error message: Unable to get ACTIONS_ID_TOKEN_REQUEST_URL env variable

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -30,9 +30,6 @@ on:
     branches:
       - main
 
-permissions:
-  attestations: write # Required for GitHub Attestation
-
 env:
   RELEASE_TYPE: ${{ github.event.inputs.release_type || 'alpha' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -31,6 +31,7 @@ on:
       - main
 
 permissions:
```
immutable-art, line 33: I'd put thus under the job itself, right before `steps`.

immutable-art (dismissed):

.github/workflows/publish.yaml
jhesgodi: done, thank you

immutable-art (approved):

zaidarain1 (approved):

PR #2133, opened 2024-09-02
[ID-2146] chore: Add tracking metrics to root Passport class

matthewmuscat:
# Summary
https://immutable.atlassian.net/browse/ID-2146

Adds further observability around root Passport SDK methods.

## Security

### DIFF
--- a/packages/passport/sdk/src/Passport.ts
+++ b/packages/passport/sdk/src/Passport.ts
@@ -138,7 +138,6 @@ export class Passport {
     const flow = trackFlow('passport', 'connectImxSilent');
 
     try {
-      flow.addEvent('startConnectImxSilent');
       return await this.passportImxProviderFactory.getProviderSilent();
     } catch (error) {
       if (error instanceof Error) {
@@ -147,7 +146,7 @@ export class Passport {
       flow.addEvent('Error');
       throw error;
     } finally {
-      flow.addEvent('endConnectImxSilent');
+      flow.addEvent('End');
     }
   }
 
@@ -155,7 +154,6 @@ export class Passport {
     const flow = trackFlow('passport', 'connectImx');
 
     try {
-      flow.addEvent('startConnectImx');
       return await this.passportImxProviderFactory.getProvider();
     } catch (error) {
       if (error instanceof Error) {
@@ -164,7 +162,7 @@ export class Passport {
       flow.addEvent('Error');
       throw error;
     } finally {
-      flow.addEvent('endConnectImx');
+      flow.addEvent('End');
     }
   }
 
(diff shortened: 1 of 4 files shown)


packages/passport/sdk/src/Passport.ts
```diff
@@ -134,18 +135,29 @@ export class Passport {
...
+    flow.addEvent('endConnectImxSilent');
+    return provider;
   }
 
   public async connectImx(): Promise<IMXProvider> {
-    return this.passportImxProviderFactory.getProvider();
+    const flow = trackFlow('passport', 'connectImx');
+    flow.addEvent('startConnectImx');
+    const provider = this.passportImxProviderFactory.getProvider();
+    flow.addEvent('endConnectImx');
```
haydenfowler, line 149: Can we follow the same pattern that we're using in `zkEvmProvider.ts` please? E.g.:
```
const flow = trackFlow('passport', FLOW_NAME);
try {
} catch (error) {
  if (error instanceof Error) {
    trackError('passport', FLOW_NAME, error);
  }
  flow.addEvent('errored');
  throw error;
} finally { 
  flow.addEvent('end');
}
```

packages/passport/sdk/src/Passport.ts
```diff
@@ -134,35 +135,74 @@ export class Passport {
    * `connectImx` instead.
    */
   public async connectImxSilent(): Promise<IMXProvider | null> {
-    return this.passportImxProviderFactory.getProviderSilent();
+    const flow = trackFlow('passport', 'connectImxSilent');
+
+    try {
+      flow.addEvent('startConnectImxSilent');
```
haydenfowler, line 141: The `start` event seems to be triggered when `trackFlow` is called (see [here](https://github.com/immutable/ts-immutable-sdk/blob/d4aa5e257dbdf681cdd241decbe689f891ca4476/packages/internal/metrics/src/flow.ts#L104)), so we should be able to omit all of these `flow.addEvent('startXXXXX');` calls

haydenfowler (commented):
LGTM 👍 happy to approve once the "start" events are removed

packages/passport/sdk/src/Passport.ts
matthewmuscat: Ahh good catch, thanks!

haydenfowler (dismissed):

packages/passport/sdk/src/utils/metrics.ts
```diff
@@ -0,0 +1,39 @@
+import { trackError, trackFlow } from '@imtbl/metrics';
+
+export const withMetrics = <T>(
+  fn: () => T,
+  event: string,
+): T => {
+  const flow = trackFlow('passport', event);
+
+  try {
+    return fn();
```
haydenfowler, line 10: LGTM 👍 
We may want to extend these implementations in the future to pass the flow to the function being executed, e.g:

```suggestion
  fn: (flow: Flow) => T,
  event: string,
): T => {
  const flow = trackFlow('passport', event);

  try {
    return fn(flow);
```

haydenfowler (approved):

PR #2136, opened 2024-09-02
[ID-2222] chore: Add tracking errors for IMX workflows

matthewmuscat:
# Summary
https://immutable.atlassian.net/browse/ID-2222

Adds error tracking to IMX workflows

## Security


packages/passport/sdk/src/starkEx/workflows/registration.ts
```diff
@@ -10,32 +11,34 @@ export default async function registerPassport(
   { ethSigner, starkSigner, imxApiClients }: RegisterPassportParams,
   authorization: string,
 ): Promise<imx.RegisterUserResponse> {
-  const [userAddress, starkPublicKey] = await Promise.all([
-    ethSigner.getAddress(),
-    starkSigner.getAddress(),
-  ]);
+  return withPassportError<imx.RegisterUserResponse>(async () => {
```
matthewmuscat, line 14: I think this change makes sense, even though we weren't doing this before?

haydenfowler (approved):

PR #2141, opened 2024-09-04
[NO CHANGELOG][Wallet Widget] remove unexpected props from img component

jhesgodi:
# Summary
TokenImage its use by MenuItem a Biome design system component that will forward a `responsiveSizes` props that should not be forwarded to `img` tag.

ie:
```tsx
      <MenuItem.FramedImage
        circularFrame
        use={(
          <TokenImage src={icon} name={name} defaultImage={defaultTokenImage} />
        )}
      />
```

### DIFF
--- a/packages/checkout/widgets-lib/src/components/TokenImage/TokenImage.tsx
+++ b/packages/checkout/widgets-lib/src/components/TokenImage/TokenImage.tsx
@@ -32,12 +32,17 @@ export function TokenImage({
     [src, error],
   );
 
+  const { ...cleanedProps } = forwardedProps as Record<string, unknown>;
+  if (Object.prototype.hasOwnProperty.call(cleanedProps, 'responsiveSizes')) {
+    delete cleanedProps.responsiveSizes;
+  }
+
   return (
     <img
       src={url}
       alt={name}
       onError={() => setError(true)}
-      {...forwardedProps}
+      {...cleanedProps}
     />
   );
 }


packages/checkout/widgets-lib/src/components/TokenImage/TokenImage.tsx
```diff
@@ -32,12 +32,17 @@ export function TokenImage({
     [src, error],
   );
 
+  const { ...cleanedProps } = forwardedProps as Record<string, unknown>;
+  if (Object.prototype.hasOwnProperty.call(cleanedProps, 'responsiveSizes')) {
+    delete cleanedProps.responsiveSizes;
+  }
+
```
mimi-imtbl, line 39: the `TokenImage` is a react component and is not using Biome, right? Why do we have to explicitly filter out `responsiveSizes`?

mimi-imtbl: ah, is it because `<MenuItem.FramedImage>` component `uses` it?

jhesgodi: **Good question**
This component is always used as the token image fallback for the MenuItem from Biome, and its this one sending the extra unexpected props.

```tsx
      <MenuItem.FramedImage
        circularFrame
        use={(
          <TokenImage src={icon} name={name} defaultImage={defaultTokenImage} />
        )}
      />
      ```

mimi-imtbl: makes sense, thanks J!

mimi-imtbl (approved):

PR #2149, opened 2024-09-05
feat: Add a new Passport config option to automatically request a transaction when SCW is not deployed

Jon-Alonso:
# Summary
Added a new passport config option to request a transaction to be submitted before signing an ERC191 message when the user's SCW is not deployed.

# Demo

https://github.com/user-attachments/assets/23bcb5fc-3b4d-47fc-8adf-d2e801f4f874

# Testing
Add a new `forceScwDeployBeforeMessageSignature` option in the sample app config to enable/test this functionality: https://github.com/immutable/ts-immutable-sdk/blob/a3b429ee6411cbceae9409486fcd51fdaad96e4b/packages/passport/sdk-sample-app/src/context/ImmutableProvider.tsx#L96

```ts
const sharedConfigurationValues = {
    scope: SCOPE,
    audience: AUDIENCE,
    redirectUri: REDIRECT_URI,
    logoutMode: LOGOUT_MODE,
    logoutRedirectUri: LOGOUT_MODE === 'silent'
      ? SILENT_LOGOUT_REDIRECT_URI
      : LOGOUT_REDIRECT_URI,
    forceScwDeployBeforeMessageSignature: true,
  };
```


packages/passport/sdk/src/zkEvm/zkEvmProvider.ts
```diff
@@ -332,6 +334,24 @@ export class ZkEvmProvider implements Provider {
...
+                  params: request.params || [],
+                  ethSigner,
+                  zkEvmAddress,
+                  rpcProvider: this.#rpcProvider,
+                  guardianClient: this.#guardianClient,
+                  relayerClient: this.#relayerClient,
+                  flow,
+                });
+              }
+            }
```
haydenfowler, line 353: I think we should implement this functionality for EIP-712 signatures as well.

From a user's perspective, I don't believe that there is any good argument for this functionality to only exist for EIP-191 signatures, and by doing this check for EIP-191 signatures and not EIP-712 signatures, we're creating another quirk that consumers need to be aware of

packages/passport/sdk/src/zkEvm/transactionHelpers.ts
```diff
@@ -0,0 +1,210 @@
...
+    ethSigner,
+  );
+
+  const [, signedTransactions] = await Promise.all([
+    validateEVMTransactionPromise,
+    getSignedMetaTransactionsPromise,
+  ]);
+
+  flow.addEvent('endValidateEVMTransaction');
+  flow.addEvent('endGetSignedMetaTransactions');
```
haydenfowler, line 204: What's the reason for firing these events once the `Promise.all` completes?

packages/passport/sdk/src/zkEvm/sendDeployTransactionAndPersonalSign.ts
```diff
@@ -0,0 +1,47 @@
...
+    guardianClient,
+    relayerClient,
+    zkEvmAddress,
+    flow,
+  });
+
+  return guardianClient.withConfirmationScreen({
+    width: 480,
+    height: 720,
+  })(async () => {
```
haydenfowler, line 32: These are the default height and width values, so they can be omitted:
```suggestion
  return guardianClient.withConfirmationScreen()(async () => {
```

haydenfowler (dismissed):

haydenfowler (approved):

PR #2161, opened 2024-09-11
chore: DVR186 Add sample switch network app

ImmutableAustinG:

# Summary

Added sample app for switching networks.

Screenshot 2024-09-11 at 10 44 29 AM


proletesseract (approved):
LGTM

examples/checkout/sdk-switch-network-with-nextjs/yarn.lock
zaidarain1, line 1: New examples should be a part of the root yarn workspace and shouldnt have their own lockfile. 

Could you please delete this lock file, add the examples/checkout/sdk-switch-network-with-nextjs directory in the root package.json workspaces field and commit the updated root yarn.lock file after running yarn install?

examples/checkout/sdk-switch-network-with-nextjs/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "lint": "next lint",
+    "test": "playwright test"
+  },
+  "dependencies": {
+    "@biom3/react": "^0.25.21",
+    "@ethersproject/providers": "^5.7.2",
+    "@imtbl/sdk": "^1.48.1",
```
zaidarain1, line 15: IMO it would be best to switch the version to "latest" so we can always have this tested and running against the latest sdk version.

examples/checkout/sdk-switch-network-with-nextjs/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "next/core-web-vitals"
+}
```
zaidarain1, line 3: a root: true key value should be added to the config to ensure it's running eslint independently to the sdk repo's one

zaidarain1 (changes requested):

examples/checkout/sdk-switch-network-with-nextjs/yarn.lock
ImmutableAustinG: Would it be better to do examples/checkout/** ? I see the connect-with-nextjs has its own yarn.lock as well.

zaidarain1: It would definitely be nice if we could try and integrate that example too within the workspace. Maybe you can remove the lock file for that too and add that as well as a workspace via the directory entry you've suggested and see how it goes. 

If it's quite broken or is causing issues and is taking too long to resolve then feel free to just specify this new example and we can come back to the other example later down the line. What do you think?

examples/checkout/sdk-switch-network-with-nextjs/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "sdk-connect-with-nextjs",
```
zaidarain1, line 2: Examples should all start with @examples/ for their package.json name. Maybe something like `@examples/sdk-connect-with-nextjs`?

zaidarain1 (changes requested):

examples/checkout/sdk-switch-network-with-nextjs/yarn.lock
ImmutableAustinG: Integrated - didn't cause any issues

zaidarain1 (approved):

PR #2167, opened 2024-09-12
[NO CHANGELOG][Checkout Widget][CM-884]feat: Relay WidgetProviderEvent

jwhardwick:

# Summary

* We need to notify the child of any `WidgetProviderEvent`, which triggers a recreation of the provider in widget global state
* Enables network and account detection/switching, even when done in a browser extension


packages/checkout/widgets-lib/src/widgets/checkout/hooks/useWidgetProviderEventRelayer.ts
```diff
@@ -0,0 +1,24 @@
+import { PostMessageHandlerEventType } from '@imtbl/checkout-sdk';
+import { useCallback, useEffect } from 'react';
+import { baseWidgetProviderEvent } from '../../../lib';
```
jhesgodi, line 3: Can we instead use `PostMessageHandlerEventType.WIDGET_PROVIDER_EVENT` here?

jhesgodi (approved):

PR #2170, opened 2024-09-12
[SDK-3238] silent signin option

andrew-yangy:
# Summary

Add silent sign in option to the login method so it can trigger a silent request.

## Security


packages/passport/sdk/src/Passport.ts
```diff
@@ -198,7 +200,11 @@ export class Passport {
         logger.warn('Failed to retrieve a cached user session', error);
       }
       if (!user && !useCachedSession) {
-        user = await this.authManager.login(options?.anonymousId);
+        if (silent) {
+          user = await this.authManager.forceUserRefresh();
```
haydenfowler, line 204: If the user has an expired application session and an expired refresh token, then this will trigger two requests to use the refresh token (one when `this.refreshTokenAndUpdatePromise()` is called from `getUser()`, and one in `forceUserRefresh`)

andrew-yangy: I would be the same case for the current login, right?
`await this.authManager.login(options?.anonymousId);`

packages/passport/sdk/src/Passport.ts
```diff
@@ -176,14 +176,16 @@ export class Passport {
...
    * prompted to log in and the Promise will resolve with a null value.
    * @param options.anonymousId - If provided, Passport internal metrics will be enriched with this value.
+   * @param options.silent - If true, will trigger a silent login request.
    * @returns {Promise<UserProfile | null>} the user profile if the user is logged in, otherwise null
    */
   public async login(options?: {
-    useCachedSession: boolean;
+    useCachedSession?: boolean;
     anonymousId?: string;
+    silent?: boolean;
```
haydenfowler, line 185: Can we rename the `silent` argument to `useSilentLogin` please? It would be good to add more detail to the JSDoc as well, e.g:
```
   * @param options.useSilentLogin - If true, and no active session exists, then the user will not be prompted to log in. Instead, we will attempt to authenticate the user silently. This approach will fail if the user does not have an active session with the authentication server, or if user input is required (for example, consent is required).
```

packages/passport/sdk/src/Passport.ts
```diff
@@ -198,7 +200,11 @@ export class Passport {
         logger.warn('Failed to retrieve a cached user session', error);
       }
       if (!user && !useCachedSession) {
-        user = await this.authManager.login(options?.anonymousId);
+        if (silent) {
+          user = await this.authManager.forceUserRefresh();
+        } else {
+          user = await this.authManager.login(options?.anonymousId);
+        }
```
haydenfowler, line 207: I wonder if it makes sense to do this check before calling `login` 🤔 That way, consumers can use `useCachedSession` & `useSilentLogin` to decide whether or not they want to fallback to displaying the login popup if the silent sign-in fails. E.g:
```
// ...
if (!user && useSilentLogin`) {
  user = await this.authManager.forceUserRefresh();
}

if (!user && !useCachedSession) {
  user = await this.authManager.login(options?.anonymousId);
}
// ...
```

packages/passport/sdk/src/Passport.ts
```diff
@@ -176,14 +176,16 @@ export class Passport {
    * @param options.useCachedSession = false - If true, and no active session exists, then the user will not be
    * prompted to log in and the Promise will resolve with a null value.
    * @param options.anonymousId - If provided, Passport internal metrics will be enriched with this value.
+   * @param options.attemptSilentLogin - If true, and no active session exists, then we will attempt to log in the user silently. If no active session exists with the authentication server, or if user input is required (for example, consent is required), then the silent login request this will fail.
```
haydenfowler, line 179:
```suggestion
   * @param options.attemptSilentLogin - If true, and no active session exists, then we will attempt to log in the user silently. If no active session exists with the authentication server, or if user input is required (for example, consent is required), then the silent login request will fail.
```

haydenfowler (approved):

PR #2211, opened 2024-09-24
ITSEC-2205 Add Dependency Review workflow

immutable-art:
# Summary
Adds Dependency Review GitHub feature https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-dependency-review#about-dependency-review for new Critical vulnerabilities only.

No new PRs will be blocked yet. This will be running in the monitoring mode until we're happy that it works flawlessly.

### DIFF
--- a/.github/workflows/dependency-review.yaml
+++ b/.github/workflows/dependency-review.yaml
@@ -1,0 +1,26 @@
+name: 'Dependency Review'
+on:
+  push:
+    branches:
+      - main
+  pull_request:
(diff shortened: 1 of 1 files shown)


immutable-art:
Current runs are ✅  https://github.com/immutable/ts-immutable-sdk/actions/workflows/dependency-review.yaml

zaidarain1 (approved):

PR #2212, opened 2024-09-24
[NO CHANGELOG][OnRamp Widget] Spike: Add option to skip connect screen

jhesgodi:
# Summary
Ramping tokens with Transak does not require users to connect their wallet first, as they can provide the address later in Transak's form. However, connecting the wallet beforehand is preferable to ensure transfer of funds to the correct wallet.

In addition, the tracking events require wallet address to be set.

The changes required to support an on-ramp flow without wallet connection are as follows:

- Conditionally wrap the `onRamp` component with a `ConnectLoader`, using a `skipConnect` flag to control this behaviour.
- Eliminate the dependency on the `ConnectLoader` state by passing `checkout` and `provider` instances directly from the widget root, instead of relying on the `ConnectLoader` state.
- Introduce `toWalletAddress` as a parameter, as it is needed for analytics and ensures that the wallet address is available upfront. The option to skip the connect step will only be allowed if the user is already logged into the application initiating the on-ramp. In cases where the wallet address is not provided by the app, connecting will remain a requirement.


packages/checkout/widgets-sample-app/src/components/ui/on-ramp/onRamp.tsx
```diff
@@ -1,38 +1,86 @@
...
-    })
+      console.log("FAILURE", data);
+    });
     onRamp.addListener(OnRampEventType.CLOSE_WIDGET, () => {
-      unmount()
-    })
-  }, []);
+      unmount();
+    });
+  }, [skipConnect]);
```
jwhardwick, line 48: @jhesgodi does `toWalletAddress` need to be a dependancy here, as we're waiting for it to be set?

jhesgodi: nice one, thank you!

PR #2215, opened 2024-09-24
fix: Enable session activity check for existing users via game bridge

keithbro-imx:

# Summary
Fixes session activity for existing users coming via game bridge

## Security

### DIFF
--- a/packages/passport/sdk/src/zkEvm/zkEvmProvider.test.ts
+++ b/packages/passport/sdk/src/zkEvm/zkEvmProvider.test.ts
@@ -101,6 +101,45 @@ describe('ZkEvmProvider', () => {
       });
     });
 
+    it('should emit accountsRequested event when the user is not authenticated', async () => {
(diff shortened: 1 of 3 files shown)


packages/passport/sdk/src/zkEvm/zkEvmProvider.test.ts
```diff
@@ -101,6 +101,45 @@ describe('ZkEvmProvider', () => {
       });
     });
 
+    it('should emit accountsRequested event when the user is not authenticated', async () => {
```
jeetparikh, line 104: is there a specific event for 'session activity' that could be tested over this?

keithbro-imx: I had the same thought that we might be able to test it a little bit more specifically. But I followed the test above which is similar to this. Lemme have a go though

keithbro-imx:
Superseded by https://github.com/immutable/ts-immutable-sdk/pull/2218

PR #2219, opened 2024-09-24
[NO CHANGELOG] Upgrade biome design-tokens to 0.4.4 and react to 0.27.2

jiyounglee:
Upgrade biome design-tokens to 0.4.4 and react to 0.27.2


packages/checkout/widgets-lib/src/views/top-up/TopUpMenuItem.tsx
```diff
@@ -25,15 +27,19 @@ export function TopUpMenuItem({
...
+          />
+        ) : (
+          <MenuItem.Icon icon={icon} />
+        )}
         <MenuItem.Label size="medium">
           {heading}
         </MenuItem.Label>
         <MenuItem.IntentIcon
-          icon={intentIcon}
+          icon={intentIcon as AllSingleVariantIconKeys}
```
glomotion, line 42: so it kinda depends on how many icons we want to support in here. If we want to support Dual and Single icon variants properly, then we should maybe do as above and use the typeGuard: `isDualVariantIcon()`

jhesgodi (dismissed):

zaidarain1 (dismissed):

packages/passport/sdk-sample-app/src/components/Message.tsx
```diff
@@ -17,7 +17,7 @@ function Message() {
   return (
     <CardStack title="Message">
       <Sticker
```
haydenfowler, line 19: @glomotion should this version upgrade have any impact on the Sticker component?
Left is after this version bump, right is prior

image

glomotion: @haydenfowler so stickers have not changed, but Sticker.FramedIcon has. 😅 Think we just need one extra prop here. Ill fix.

glomotion: OK this commit should fix it: https://github.com/immutable/ts-immutable-sdk/pull/2219/commits/819e41e162858f81a93b6d3f55860ba1e4b9bdda

haydenfowler: @glomotion Same result unfortunately 😅

glomotion: @haydenfowler OK should definitely be fixed this time. Sorry for the false start. 🤪

haydenfowler (approved):

zaidarain1 (approved):

jhesgodi (approved):

PR #2220, opened 2024-09-24
chore: [DX-3280] Fix missing dependencies, flaky configs and tests

zaidarain1:
Some packages import from dependencies without having them listed in their dependencies in their respective package.jsons. Add these to the package.json dependencies as needed.

Some packages have configs and/or tests that are sensitive to breakage since they aren’t explicit or don't handle certain situations. Update these to be more resilient and handle any workflows that can introduce issues (for eg .yalc artifacts resulting in problems with jest)


allan-almeida-imtbl (approved):

sergerad (approved):

lfportal (approved):

jeetparikh (approved):

packages/internal/dex/sdk/package.json
```diff
@@ -62,7 +68,7 @@
     "lint:fix": "cd ../../../.. && yarn wsrun -p @imtbl/dex-sdk -c lint --fix",
     "test": "jest test",
     "test:watch": "jest --watch",
-    "typecheck": "tsc --noEmit"
+    "typecheck": "tsc --customConditions default --noEmit"
```
keithbro-imx, line 71: What does this do?

keithbro-imx (approved):

build-dependents.js
```diff
@@ -23,7 +23,7 @@ try {
 
   if (isDependent || changedProject === currentProject) {
     // Rebuild the current project
-    const command = `nx run-many --target=d --projects=${currentProject} --parallel=5 --no-cloud`;
+    const command = `nx run-many --target=d --projects=${currentProject} --no-cloud`;
```
jeetparikh, line 26: why `no-cloud` for this? is it to skip caching hit?

jeetparikh: reduce parallel count for flakiness?

packages/internal/dex/sdk/package.json
zaidarain1: Recently live types were added, so you can have live typechecking across packages without needing to build/save, as if the code lives within one monolithic package. We use custom export conditions to enable that behaviour, while also allowing us to switch back to declaration files when building for publishing. 

The problem we can face sometimes is without falling back to the declaration files for types from children workspace packages when doing a full typecheck during build or via the above typechecking command, it can produce errors as the parent tsconfig might not be be configured correctly to handle the children package's ts files, one example being jsx/tsx. This just ensures that we fallback to the default conditions correctly during the typechecking process. The only time we should be using the src .ts files directly for types is for the ts server in the IDE to provide cross package live types. 

This gives us the best of both worlds.

build-dependents.js
zaidarain1: no-cloud here will keep local cache, but skip the remote cache. With the dev mode being used mostly for constant changes, trying to hit the cloud again resulted in wasted resources and added some time to the dev rebuild loop. The no-cloud was added a bit ago to save the few seconds that were being wasted hitting the cloud at the start of each dev loop unnecessarily.

The parallel count was just moved to a common value in the nx.json rather than being hard coded everywhere, and was bumped to 8 to maximize the 8 core usage of our runners in the more workload heavy workflows 😄

matthewmuscat (approved):

alejoloaiza (approved):

PR #2228, opened 2024-09-25
feat: CM-901 - Implement custom order data for primary sales sign call

luads:
# Summary

## Security


packages/checkout/sdk/src/widgets/definitions/parameters/sale.ts
```diff
@@ -28,6 +28,8 @@ export type SaleWidgetParams = {
   excludeFiatCurrencies?: string[];
   /** Preferred currency, replacing the backend's base currency */
   preferredCurrency?: string;
+  /** Custom key-value pairs to be passed to the order */
+  customOrderData?: Record<string, any>;
```
jwhardwick, line 32: @luads Better to use `unknown` here

luads: Fixed!

jwhardwick (approved):

PR #2237, opened 2024-09-26
[NO CHANGELOG] Add funds/update use execute

jiyounglee:
Create isSquidNativeToken function
Update useExecute
- Approve transaction manually due to bug in squid sdk
- add convertToNetworkChangeableProvider to be used to change network without error
- Execute transaction to return Transaction Receipt


packages/checkout/widgets-lib/src/widgets/add-funds/functions/isSquidNativeToken.ts
```diff
@@ -0,0 +1,3 @@
+export const SQUID_NATIVE_TOKEN = '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee';
```
mimi-imtbl, line 1: let's put this in `config.ts`?

mimi-imtbl (approved):

packages/checkout/widgets-lib/src/widgets/add-funds/functions/isSquidNativeToken.ts
jiyounglee: Remove export to only be used in this function

PR #2242, opened 2024-09-30
[NO CHANGELOG][Checkout Widget] Add types for add funds widget

jhesgodi:
# Summary
Add missing types to integrate add funds flow into Checkout Widget

### DIFF
--- a/packages/checkout/sdk/src/widgets/definitions/events/addFunds.ts
+++ b/packages/checkout/sdk/src/widgets/definitions/events/addFunds.ts
@@ -8,3 +8,24 @@ export enum AddFundsEventType {
   REQUEST_ONRAMP = 'request-onramp',
   REQUEST_SWAP = 'request-swap',
 }
+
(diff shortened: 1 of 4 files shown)


packages/checkout/sdk/src/widgets/definitions/events/addFunds.ts
```diff
@@ -8,3 +8,24 @@ export enum AddFundsEventType {
...
+  transactionHash: string;
+};
+
+/**
+ * Type representing a add funds failure
+ * @property {string} reason
+ * @property {number} timestamp
+ */
+export type AddFundsFailed = {
+  /** The reason why the on-ramp failed. */
```
jiyounglee, line 27: Instead of **on-ramp** it should be either **add funds or (bridge or swap)**

jiyounglee (approved):

jiyounglee (approved):

PR #2251, opened 2024-09-30
[NO CHANGELOG] [Add Funds Widget]Add funds/review view

jiyounglee:
Create Review View
Create Squid Icon SVG
Update useRoute getRoute to include fromAddress which is needed when quoteOnly set to false


packages/checkout/widgets-lib/src/widgets/add-funds/hooks/useRoutes.ts
```diff
@@ -101,6 +101,7 @@ export const useRoutes = () => {
     toToken: Token,
     toAmount: string,
     toAddress: string,
+    fromAddress:string | undefined = undefined,
```
mimi-imtbl, line 104: I think this could just be written as `fromAddress?: string `

mimi-imtbl (approved):

PR #2278, opened 2024-10-07
[NO CHANGELOG] [Add Funds Widget] Get token to calculate USD amount

jiyounglee:
Get token to calculate USD amount


mimi-imtbl (approved):

packages/checkout/widgets-lib/src/widgets/add-funds/views/AddFunds.tsx
```diff
@@ -230,6 +231,31 @@ export function AddFunds({
   //   console.log('handle review click');
   // }, []);
 
+  useEffect(() => {
```
jhesgodi, line 234: Doing this calculation in react, reduces the testability and performance of the code.

I suggest you move these to a `getFiatAmount` utility fn:

```export function getFiatAmount(amount: string, token: TokenInfo, tokens: TokenInfo[]): Number {
  if (!amount || !token?.address || !tokens) {
    return 0;
  }

  const address =
    token.address === "native" ? SQUID_NATIVE_TOKEN : token?.address;

  const toToken = tokens.find(
    (value) =>
      value.address.toLowerCase() === address.toLowerCase() &&
      value.chainId === ChainId.IMTBL_ZKEVM_MAINNET.toString()
  );

  if (!toToken) {
    return 0;
  }

  return Number(amount) * toToken.usdPrice;
}
```
and set the value using a `useMemo` hook.

```
const fiatAmount = useMemo(
  () => getFiatAmount(inputValue, currentToTokenAddress, tokens);
}, [inputValue, currentToTokenAddress, tokens]);
```

packages/checkout/widgets-lib/src/widgets/add-funds/views/AddFunds.tsx
```diff
@@ -230,6 +231,31 @@ export function AddFunds({
...
+        && value.chainId === ChainId.IMTBL_ZKEVM_MAINNET.toString(),
+    );
+
+    if (!toToken) {
+      return;
+    }
+
+    const tokenBalance = Number(inputValue) * toToken.usdPrice;
+    if (tokenBalance > 0) {
+      setAmountInUSD(`USD $${tokenBalance.toFixed(2)}`);
```
jhesgodi, line 255: Please avoid `USD $${tokenBalance.toFixed(2)}` as the amounts format will be added by the translation interpolation.

ie:
`t(`ADD_FUNDS_FIAT_AMOUNT`, { fiatAmount })` where `amount` is `number | string`

this allows the language to set the format for the currency template, also notice rename it to `fiatAmount` as anchor currency could change

jhesgodi (changes requested):
review it with some comments

jhesgodi (approved):

packages/checkout/widgets-lib/src/widgets/add-funds/views/AddFunds.tsx
jiyounglee: Done

packages/checkout/widgets-lib/src/widgets/add-funds/views/AddFunds.tsx
jiyounglee: Done

mimi-imtbl (dismissed):

mimi-imtbl (approved):

PR #2312, opened 2024-10-16
[NO CHANGELOG][Add Funds Widget] Add validation for selected deliver to and pay with wallets

jhesgodi:
# Summary
- Prevent user from being able to send funds in between different accounts using the same wallet type
  - ie: Metamask 0x123 to Metamask 0x321 wont be supported, likewise for Passport
- Validate pay with wallet address hasn't change before allowing to proceed in review screen


packages/checkout/widgets-lib/src/widgets/add-funds/views/Review.tsx
```diff
@@ -661,7 +672,7 @@ export function Review({
           </>
         )}
 
-        {!route && <LoadingView loadingText="Securing quote" />}
+        {!route && !showAddressMissmatchDrawer && <LoadingView loadingText="Securing quote" />}
```
jhesgodi, line 675: this prevents the loading from ever showing on top of the drawer

jiyounglee (approved):

PR #2315, opened 2024-10-16
feat: Add IMX wrapping to the Checkout SDK

luads:
# Summary

This method introduces a way for wrapping IMX into wIMX and the other way around.

## Security


alejoloaiza (approved):

alejoloaiza (approved):

jwhardwick (commented):
Looks good @luads .

Thoughts on making `getUnsignedWrapUnwrapTx` public and creating an `Exchange` instance to use it? https://github.com/immutable/ts-immutable-sdk/blob/7067a430908813c77fd0c8070e97fea571f39846/packages/internal/dex/sdk/src/exchange.ts#L211

Benefit is having this logic live in one place, and that code already handles fee quotes.

alejoloaiza (changes requested):
Not needed anymore

PR #2333, opened 2024-10-21
[NO CHANGELOG] [Add Funds Widget] Update to calculate from amount from route response when route to amou…

jiyounglee:
Update to calculate from amount from route response when route to amount is less than input to amount

### DIFF
--- a/packages/checkout/widgets-lib/src/widgets/add-funds/hooks/useRoutes.ts
+++ b/packages/checkout/widgets-lib/src/widgets/add-funds/hooks/useRoutes.ts
@@ -52,6 +52,15 @@ export const useRoutes = () => {
     return fromAmountWithBuffer.toString();
   };
 
+  const calculateFromAmountFromRoute = (
+    exchangeRate: string,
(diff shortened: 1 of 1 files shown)


packages/checkout/widgets-lib/src/widgets/add-funds/hooks/useRoutes.ts
```diff
@@ -52,6 +55,15 @@ export const useRoutes = () => {
     return fromAmountWithBuffer.toString();
   };
 
+  const calculateFromAmountFromRoute = (
+    exchangeRate: string,
+    toAmount: string,
+  ) => {
+    const fromAmount = Number(toAmount) / Number(exchangeRate);
+    const fromAmountWithBuffer = fromAmount * 1.02;
```
mimi-imtbl, line 63: should we exract the 1.02 number as a const on top? seems like it's being used in both `calculateFromAmountFromRoute` and `calculateFromAmount` 👀

mimi-imtbl (approved):

PR #2342, opened 2024-10-22
refactor: Refactored to use Biome and separated apps

Calvin-Lieu:
App now uses biome and has been split into two apps, one for login with nextjs and one for passport methods.

https://immutable.atlassian.net/jira/software/projects/DVR/boards/2479?selectedIssue=DVR-196


proletesseract:
Outdated

PR #2346, opened 2024-10-23
[CM-1017][CM-986] feat: Simplify widget e2e 

jwhardwick:

Discovered that resolving breaking changes is difficult, as the example app was being built with both the latest production SDK and the local branch’s SDK. E2E tests were running against both, in the case of a breaking we can’t get both of those tests to pass. 

Based on this, we'll need separate apps for each case.
* moves auto update testing to `/test`
* keeps widget sample app in `/examples`

Some updated details of this process are here https://immutable.atlassian.net/wiki/spaces/PR/pages/2796814550/draft+Checkout+Widgets+Auto-Updating+Docs

https://immutable.atlassian.net/browse/CM-986


jwhardwick:
Had a discussion with @zaidarain1 - decided the best approach is to have two separate apps for this.
1. keep the examples app
2. copy this over to `/tests/func-tests` to test the auto update logic there

zaidarain1 (approved):

PR #2357, opened 2024-10-30
[NO CHANGELOG] [Add Tokens Widget]  Rename to add tokens widget

mimi-imtbl:

# Summary

## Security


jiyounglee (approved):

jwhardwick (changes requested):
@mimi-imtbl @jiyounglee I suspect this will break autoupdating as the name of the widget is changing. We don't have a great way to test for Add Funds as it's not on Sandbox.

I removed this from the queue for now. I'll see if we can still test AddTokens without it live on sandbox.

mimi-imtbl:
> @mimi-imtbl @jiyounglee I suspect this will break autoupdating as the name of the widget is changing. We don't have a great way to test for Add Funds as it's not on Sandbox.
> 
> I removed this from the queue for now. I'll see if we can still test AddTokens without it live on sandbox.

Hey Jimmy, I've updated the name change everywhere including within Commerce Widget. This was tested on both `sandbox` and `prod`.

jwhardwick:
@mimi-imtbl Add Funds is live on Toolkit right? https://toolkit.immutable.com/add-tokens/

With auto updates, Toolkit will fetch the latest bundle of the widgets code at runtime. It will try to call `factory.create(WidgetType.ADD_FUNDS)` however that will fail as that WidgetType is no longer supported in this change.

https://github.com/immutable/checkout-playground/blob/86bb30ae3967799d5d6917f310c655443fc407d4/src/context/WidgetContext.tsx#L134

jwhardwick:
@mimi-imtbl @jiyounglee This PR (https://github.com/immutable/ts-immutable-sdk/pull/2358) has an extra test for add funds which should catch this, ty for your patience.

You'll need to to stop auto updates at this version https://immutable.atlassian.net/wiki/spaces/PR/pages/2796814550/Commerce+Widgets+Auto-Updating+Docs, then should be good to merge this PR. Let me know if you need any help.

jwhardwick (approved):

PR #2376, opened 2024-11-07
feat: ID-2774 Handle 403s from Guardian evaluate endpoint

imx-mikhala:

# Summary

- Handle 403s from Guardian evaluate endpoint

## Security


packages/passport/sdk/src/confirmation/confirmation.ts
```diff
@@ -166,6 +166,14 @@ export default class ConfirmationScreen {
     });
   }
 
+  showServiceUnavailable() {
+    this.showConfirmationScreen(
+      this.getHref('unavailable'),
+      () => this.closeWindow(),
+      () => this.closeWindow(),
+    );
+  }
```
haydenfowler, line 175: With this setup, the Promise is rejected but the overlay is still in the DOM and the popup is still open. This means that consumers don't have any way to know when the popup has been closed, so I think we should update the implementation to only throw the error once the popup is closed 🤔

haydenfowler:
An unhandled error seems to be thrown once the popup is closed:
image

haydenfowler (commented):
Can we add some tests please? Apart from that LGTM ✅

haydenfowler (approved):

PR #2399, opened 2024-11-14
fix: ID-2650 Handle unknown or invalid refresh token error

imx-mikhala:

# Summary

- Handle unknown or invalid refresh token error when initialising a ZkEvmProvider 

## Security

### DIFF
--- a/packages/passport/sdk/src/Passport.ts
+++ b/packages/passport/sdk/src/Passport.ts
@@ -192,10 +192,8 @@ export class Passport {
       try {
         user = await this.authManager.getUser();
       } catch (error) {
-        if (error instanceof Error) {
-          trackError('passport', 'login', error);
-        }
         if (useCachedSession) {
+          if (error instanceof Error) trackError('passport', 'login', error);
           throw error;
         }
         logger.warn('Failed to retrieve a cached user session', error);
--- a/packages/passport/sdk/src/Passport.ts
+++ b/packages/passport/sdk/src/Passport.ts
@@ -192,8 +192,10 @@ export class Passport {
       try {
         user = await this.authManager.getUser();
       } catch (error) {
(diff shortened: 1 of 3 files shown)


packages/passport/sdk/src/zkEvm/zkEvmProvider.ts
```diff
@@ -123,7 +123,9 @@ export class ZkEvmProvider implements Provider {
           this.#callSessionActivity(user.zkEvm.ethAddress);
         }
       }
-    }).catch(); // User does not exist, don't initialise an eth signer
+    }).catch(() => {
+      // User does not exist, don't initialise an eth signer
+    });
```
imx-mikhala, line 128: This was where the error was not being caught - catch requires a function to handle the error

packages/passport/sdk/src/Passport.ts
```diff
@@ -192,10 +192,8 @@ export class Passport {
       try {
         user = await this.authManager.getUser();
       } catch (error) {
-        if (error instanceof Error) {
-          trackError('passport', 'login', error);
-        }
         if (useCachedSession) {
+          if (error instanceof Error) trackError('passport', 'login', error);
```
haydenfowler, line 196: I think where the `trackError` call was before made sense, but I think the right solution is to no longer surface unknown or invalid refresh token "errors" as errors. Maybe something like this:
```
} catch (error) {
  // pseudo
  if (errorIsUnknownOrInvalidRefreshToken) {
    flow.addEvent('unknown or invalid refresh token');
  } else {
    trackError('passport', 'login', error);
  }
```

haydenfowler: Yeah that's correct, that way we'll know if there are any other errors being being thrown regularly

haydenfowler (approved):

PR #2454, opened 2024-12-05
ID-3085 - Purchase add post hooks logic

alejoloaiza:
# Summary
 - Adding post hooks logic

## Security


luads (approved):

packages/checkout/widgets-lib/src/lib/squid/functions/fetchBalances.ts
```diff
@@ -11,21 +11,18 @@ export const fetchBalances = async (
...
+      chains: [chainId],
+      userAddress: address,
     });
     promises.push(balancePromise);
   }
 
   const balances = await Promise.all(promises);
   return balances
-    .flatMap((balance) => balance.evmBalances ?? [])
+    .flatMap((balance) => balance)
```
mimi-imtbl, line 26: I think we don't need a `flatMap` here anymore

mimi-imtbl (approved):

PR #2478, opened 2024-12-18
fix: change session activity to use local date

shineli1984:
# Summary
Change session activity to use local date so that D1 makes more sense in user context.

### DIFF
--- a/packages/passport/sdk/src/zkEvm/sessionActivity/sessionActivity.ts
+++ b/packages/passport/sdk/src/zkEvm/sessionActivity/sessionActivity.ts
@@ -19,8 +19,12 @@ const syncSendCount = () => {
   sendCount = getItem(SESSION_ACTIVITY_COUNT_KEY) || {};
   const sendDay = getItem(SESSION_ACTIVITY_DAY_KEY);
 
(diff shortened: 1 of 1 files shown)


packages/passport/sdk/src/zkEvm/sessionActivity/sessionActivity.ts
```diff
@@ -19,8 +19,12 @@ const syncSendCount = () => {
...
   const sendDay = getItem(SESSION_ACTIVITY_DAY_KEY);
 
-  // If no day, set count to zero. If not today, reset sendCount to 0
-  const today = new Date().toISOString().split('T')[0];
+  // If no day, init sendCount. If not today, reset sendCount
+  const date = new Date();
+  const yyyy = date.getFullYear();
+  const mm = `${date.getMonth() + 1}`.padStart(2, '0');
+  const dd = `${date.getDate()}`.padStart(2, '0');
+  const today = `${yyyy}-${mm}-${dd}`;
```
haydenfowler, line 27: Looks like this can be simplified:
```suggestion
  const today = (new Date()).toISOString().substr(0, 10);
```

haydenfowler (approved):

PR #2479, opened 2024-12-19
fix: ID-3129 Prevent Unknown or invalid refresh token requests from being triggered multiple times

imx-mikhala:

# Summary

- Prevent Unknown or invalid refresh token requests being triggered multiple times

## Security

### DIFF
--- a/packages/passport/sdk/src/authManager.ts
+++ b/packages/passport/sdk/src/authManager.ts
@@ -485,6 +485,7 @@ export default class AuthManager {
 
         if (err instanceof ErrorTimeout) {
           passportErrorType = PassportErrorType.SILENT_LOGIN_ERROR;
+          errorMessage = `${errorMessage}: ${err.message}`;
           removeUser = false;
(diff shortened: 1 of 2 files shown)


packages/passport/sdk/src/authManager.ts
```diff
@@ -481,18 +481,30 @@ export default class AuthManager {
       } catch (err) {
         let passportErrorType = PassportErrorType.AUTHENTICATION_ERROR;
         let errorMessage = 'Failed to refresh token';
+        let removeUser = true;
 
         if (err instanceof ErrorTimeout) {
           passportErrorType = PassportErrorType.SILENT_LOGIN_ERROR;
+          removeUser = false;
```
CodeSchwert, line 488: Should we set the `errorMessage` in this block similar to the other blocks?

CodeSchwert (dismissed):

CodeSchwert (approved):

PR #2480, opened 2024-12-20
chore: add Security to .github as approvers

immutable-art:
# Summary
Adds Security as approvers in CODEOWNERS to `.github`, along with the SDK team

PR #2481, opened 2024-12-20
[SRE-2781] Set publishing workflows to use prod env

dhf22:
TS_IMMUTABLE_SDK_NPM_TOKEN secret will now be stored as an environment secret in prod environment, meaning it will only be accessible from workflows running against prod env. This is to ensure only protected main brain with required reviews is able to publish to NPM to mitigate potential security incidents. 

Prod env will require reviewers from the sdk team, and these workflows will only be executable against main.

### DIFF
--- a/.github/workflows/publish-major-version.yaml
+++ b/.github/workflows/publish-major-version.yaml
@@ -30,6 +30,7 @@ on:
 jobs:
   PublishMajorVersion:
     name: Publish Workflow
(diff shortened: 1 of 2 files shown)


immutable-art (approved):

immutable-art:
Relates to https://github.com/immutable/ts-immutable-sdk/pull/2480

PR #2498, opened 2025-01-15
feat: Adding nonce spaces for session activity requests.

shirren:

# Summary

Adding support for nonce spaces when making requests to obtain a nonce for a passport wallet. SessionActivity requests utilise this to obtain nonces in a different space to all other nonce request per passport wallet.

A nonce space is a Sequence concept. A simple mental model for nonce spaces is a map, where the space is the key used to lookup the next nonce for the SCW. A nonce is stored in a 256 bit value. The first 160 bits are used for the space, and the remaining 96 for the nonce value. The original getNonce implementation used a space of 0, this means in effect the numbers are 0,1,2,3…. so on and so forth.

In the case of SessionActivity the space is 1, so the sequence becomes 79228162514264337593543950336, 79228162514264337593543950337, 79228162514264337593543950338… so on and so forth.

Important to note that means the maximum number of transactions each SCW can execute before there is a nonce conflict due to session activity is 79228162514264337593543950336, which I think is pretty safe. We could also pick a larger abtriray number for the session activity space, but I think that just complicates things. I think it is simpler for us to just increase the spaces sequentially as we find new use cases for spaces.


pano-skylakis (dismissed):
Nice one - Shall we start a confluence doc to track nonce spaces?

I think we just have 3 in use at the moment, is that correct?

* Default
* Session activity
* Minting API

haydenfowler (approved):

packages/passport/sdk/src/zkEvm/walletHelpers.ts
```diff
@@ -55,19 +56,47 @@ export const encodedTransactions = (
...
       smartContractWalletAddress,
       walletContracts.mainModule.abi,
       rpcProvider,
     );
-    const result = await contract.nonce();
+    const space: BigNumber = coerceNonceSpace(nonceSpace); // Default nonce space is 0
+    const result = await contract.readNonce(space);
     if (result instanceof BigNumber) {
-      return result;
+      return encodeNonce(space, result);
```
shirren, line 99: We also need to encode the space into the nonce returned. My bad I thought the returned nonce was already encoded.

pano-skylakis: I thought it was already encoded as well 🤔

pano-skylakis (approved):

imx-mikhala (approved):

PR #2512, opened 2025-01-28
Exclude tokens not bridged via native bridge from widget

luads:
# Summary

## Security


alejoloaiza (approved):
lgtm

alejoloaiza (approved):

PR #2527, opened 2025-02-14
[ID-3255] IR2 transactions being rejected with no user's input

rodrigo-fournier-immutable:
This PR aims to solve an issue where concurrent transactions trigger the closeWindow method through a backend flow, disrupting the expected behavior of the frontend flow.

The solution found was to identify the names of the transactions that call the validateEVMTransaction method. If the transaction is SessionActivity (a backend transaction that should not close the confirmation screen), the application will properly react by avoiding the method that closes the screen.


packages/passport/sdk/src/guardian/index.ts
```diff
@@ -245,7 +247,9 @@ export default class GuardianClient {
           'Transaction rejected by user',
         );
       }
-    } else {
+      // This verification is meant to ensure that it originates from zkEvmProvider#callSessionActivity
+      // and since it's a background transaction should not close the confirmation screen window.
+    } else if (BigNumber.from(1).eq(nonceSpace ?? 0)) {
```
imx-mikhala, line 252: For this logic, we want to not fall into this `else if` when nonceSpace equals 1 so we need to negate this check

`else if (!BigNumner.from(1).eq(...)`

packages/passport/sdk/src/zkEvm/transactionHelpers.ts
```diff
@@ -194,6 +198,7 @@ export const prepareAndSignTransaction = async ({
       chainId: getEip155ChainId(chainId),
       nonce: convertBigNumberishToString(nonce),
       metaTransactions,
+      nonceSpace,
```
haydenfowler, line 201: Callers of `validateEMTransaction` likely wouldn't expect the `nonceSpace` argument to change anything other than the nonce space that is used for the transaction, so this may cause confusion or lead to unexpected results in the `nonceSpace` changes at some point in the future.

With this in mind, we may be better off explicitly adding an optional argument to `sendTransaction` (that is then passed to `prepareAndSignTransaction` & `validateEVMTransaction`) to indicate that the popup shouldn't be closed, e.g. `isBackgroundTransaction` (I'm open to naming suggestions here 😅). What do you think?

rodrigo-fournier-immutable: My first implementation was to create a new parameter called transactionName to indicate when the origin was sessionActivity, but I wasn’t entirely satisfied with it and shared my thoughts with Mikhala. We decided to use nonceSpace instead, but I agree with the arguments that were presented. I’m somewhat resistant to adding too many parameters to functions, but a parameter with the suggested name clearly defines its purpose, which makes me much more satisfied.

packages/passport/sdk/src/guardian/index.test.ts
```diff
@@ -231,6 +231,110 @@ describe('Guardian', () => {
...
+        metaTransactions: [
+          {
+            data: transactionRequest.data,
+            revertOnError: true,
+            to: mockUserZkEvm.zkEvm.ethAddress,
+            value: '0x00',
+            nonce: 1,
+          },
+        ],
+        isBackgroundTransaction: false, // not passing this parameter should result in the same behavior cause default value is false
```
imx-mikhala, line 307: Should we just remove passing this field in the test so it uses the default? That way if for some reason the default is modified, this test will fail

rodrigo-fournier-immutable: Maybe I can create a new test to use the default value, what do you think?

imx-mikhala (approved):

PR #2528, opened 2025-02-14
chore: Track events on different swap form interactions

jeetparikh:
# Summary
Track different events on swap form interactions. There were no events on the to and from fields and input changes - so it was hard to know the drop off as there only final event on the swap button. 
Also added an event for the "unable to swap" modal popups for any kind of swap error

### DIFF
--- a/packages/checkout/widgets-lib/src/widgets/swap/components/SwapForm.tsx
+++ b/packages/checkout/widgets-lib/src/widgets/swap/components/SwapForm.tsx
@@ -228,7 +228,7 @@ export function SwapForm({ data, theme, cancelAutoProceed }: SwapFromProps) {
         toToken,
         fromAmount,
         toAmount,
-        error: error.message,
+        error: 'message' in error ? error.message : error,
       },
     });
   }, [track, fromToken, toToken, fromAmount, toAmount]);
--- a/packages/checkout/widgets-lib/src/widgets/swap/components/SwapForm.tsx
+++ b/packages/checkout/widgets-lib/src/widgets/swap/components/SwapForm.tsx
@@ -215,6 +215,17 @@ export function SwapForm({ data, theme, cancelAutoProceed }: SwapFromProps) {
     network,
   ]);
 
+  const openUnableToSwapDrawer = () => {
+    setShowNotEnoughImxDrawer(false);
+    setShowUnableToSwapDrawer(true);
+    track({
+      userJourney: UserJourney.SWAP,
+      screen: 'SwapCoins',
+      control: 'UnableToSwapDrawer',
+      controlType: 'Button',
+    });
+  };
+
   const tokensOptionsTo = useMemo(() => allowedTokens
     .map(
       (token) => ({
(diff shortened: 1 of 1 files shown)


keithbro-imx (approved):

packages/checkout/widgets-lib/src/widgets/swap/components/SwapForm.tsx
```diff
@@ -215,6 +215,17 @@ export function SwapForm({ data, theme, cancelAutoProceed }: SwapFromProps) {
     network,
   ]);
 
+  const openUnableToSwapDrawer = () => {
```
keithbro-imx, line 218: Worth a useMemo?

packages/checkout/widgets-lib/src/widgets/swap/components/SwapForm.tsx
```diff
@@ -215,6 +215,17 @@ export function SwapForm({ data, theme, cancelAutoProceed }: SwapFromProps) {
...
+    setShowNotEnoughImxDrawer(false);
+    setShowUnableToSwapDrawer(true);
+    track({
```
keithbro-imx, line 221: I would use `page` here instead of `track` as this is something the user sees rather than something that they did. There is no "control" for example

https://segment.com/docs/connections/spec/page/
https://segment.com/docs/connections/spec/track/

keithbro-imx: How easy would it be to include the token pair here, and potentially also the error message? The goal here is we want to be able to know which token pairs don't have liquidity that the users are trying to interact with

packages/checkout/widgets-lib/src/widgets/swap/components/SwapForm.tsx
jeetparikh: added.. usually don't default to using one

packages/checkout/widgets-lib/src/widgets/swap/components/SwapForm.tsx
jeetparikh: > I would use page here instead of track as this is something the user sees rather than something that they did. There is no "control" for example
I don't quite agree with the part there is no control. They are seeing something (error) as a result of their action of trying to swap by putting in the input tokens? `Page` event is only when you land on a page as a result of navigation

jeetparikh: added the necessary information in extras

pano-skylakis (dismissed):

packages/checkout/widgets-lib/src/widgets/swap/components/SwapForm.tsx
```diff
@@ -215,6 +215,24 @@ export function SwapForm({ data, theme, cancelAutoProceed }: SwapFromProps) {
...
+      userJourney: UserJourney.SWAP,
+      screen: 'SwapCoins',
+      control: 'UnableToSwapDrawer',
+      controlType: 'Button',
+      extras: {
+        fromToken,
+        toToken,
+        fromAmount,
+        toAmount,
+        error: error.message,
```
keithbro-imx, line 231:
```suggestion
        error: 'message' in error ? error.message : error,
```

keithbro-imx (dismissed):

keithbro-imx (approved):

pano-skylakis (approved):

PR #2529, opened 2025-02-17
DOCS: Added missing typedoc comments for methods | DVR-305

darrenmelvison1:
## Description
Added a lot of missing typedoc comments for methods, including: method description, params description, return description, etc.

JIRA: https://immutable.atlassian.net/browse/DVR-305


zaidarain1 (approved):

packages/passport/sdk/src/Passport.ts
```diff
@@ -135,22 +135,29 @@ export class Passport {
...
-  } = {
-    announceProvider: true,
-  }): Provider {
+  /**
+   * Connects to EVM and optionally announces the provider.
+   * @param {Object} options - Configuration options
+   * @param {boolean} options.announceProvider - Whether to announce the provider via EIP-6963 for wallet discovery (defaults to true)
+   * @returns {Provider} The EVM provider instance
+   */
+  public connectEvm(options: { announceProvider: boolean }): Provider {
```
zaidarain1, line 160: Ah, you've removed the default assignment of announceprovider: true, resulting in the build failing as the function params are now required. I think builds should be passing once this is fixed

zaidarain1 (changes requested):

zaidarain1 (approved):

packages/checkout/sdk/src/sdk.ts
```diff
@@ -322,14 +326,19 @@ export class Checkout {
 
   /**
    * Subscribes to changes in the injected providers.
-   * @param listener - The listener to be called when the injected providers change.
+   * @param {(providers: EIP6963ProviderDetail[]) => void} listener - The listener to be called when the injected providers change.
+   * @returns {void} - A function to unsubscribe the listener.
```
imx-mikhala, line 330: Should this be ` @returns {() => void}` or similar? Technically saying this returns void is not accurate, but instead returns a function that returns void

darrenmelvison1: Yes, you're right. I've changed it!

imx-mikhala (approved):

shineli1984 (dismissed):

mdymalla (approved):

frankisawesome (approved):

frankisawesome (approved):

shineli1984 (approved):

imx-mikhala (approved):

mdymalla (approved):

PR #2533, opened 2025-02-20
chore(passport): Check isLoggedIn before calling loginWithOIDC

imx-mikhala:

# Summary

- Check isLoggedIn before calling loginWithOIDC to prevent unnecessary calls to the Magic OIDC endpoint

## Security


packages/passport/sdk/src/magic/magicAdapter.ts
```diff
@@ -51,10 +51,13 @@ export default class MagicAdapter {
...
-          providerId: this.config.magicProviderId,
-        });
+        const isUserLoggedIn = await magicClient.user.isLoggedIn();
+        if (!isUserLoggedIn) {
+          await magicClient.openid.loginWithOIDC({
+            jwt: idToken,
+            providerId: this.config.magicProviderId,
+          });
+        }
         flow.addEvent('endLoginWithOIDC');
```
shirren, line 61: @imx-mikhala shouldn't the event then also be included inside of the `!isUserLoggedIn` predicate?

packages/passport/sdk/src/magic/magicAdapter.test.ts
```diff
@@ -102,6 +105,22 @@ describe('MagicWallet', () => {
       expect(magicProvider).toEqual(rpcProvider);
     });
 
+    it('should not call loginWithOIDC if isLoggedIn is true', async () => {
```
shirren, line 108: 🚀

packages/passport/sdk/src/magic/magicAdapter.ts
imx-mikhala: Hm we use this event to indicate when loginWithOIDC hangs (so we see a magicLoginStart event and no follow up loginWIthOIDC which indicates the call didn't complete). If we move into the if, it might be harder to debug if a user was stuck at loginWithOIDC (as we would no longer emit this event if they are logged in, but we won't know the difference between a user already being logged in with Magic or a hanging call)

We also don't want to add another event for this as we already had to reduce the number of events being emitted.

Let me know if you have any thoughts

shirren: That makes sense, guess if we can't add semantically more correct events then totally acceptable.

shirren (approved):

PR #2543, opened 2025-02-21
feat(passport): Redirect based login

rodrigo-fournier-immutable:
This PR aims to add support for redirect based login, to solve the issues described in [this](https://immutable.atlassian.net/browse/ID-3265) task.

### DIFF
--- a/packages/passport/sdk-sample-app/src/context/ImmutableProvider.tsx
+++ b/packages/passport/sdk-sample-app/src/context/ImmutableProvider.tsx
@@ -12,9 +12,9 @@ import { Passport, PassportModuleConfiguration } from '@imtbl/passport';
 import { Environment, ImmutableConfiguration, ModuleConfiguration } from '@imtbl/config';
 import {
   AUDIENCE,
-  LOGIN_REDIRECT_URI,
-  LOGOUT_REDIRECT_URI,
+  POPUP_REDIRECT_URI,
   REDIRECT_URI,
+  LOGOUT_REDIRECT_URI,
   SILENT_LOGOUT_REDIRECT_URI,
   LOGOUT_MODE,
   SCOPE,
@@ -90,9 +90,9 @@ const getPassportConfig = (environment: EnvironmentNames): PassportModuleConfigu
   const sharedConfigurationValues = {
     scope: SCOPE,
     audience: AUDIENCE,
+    popupRedirectUri: POPUP_REDIRECT_URI,
     redirectUri: REDIRECT_URI,
     logoutMode: LOGOUT_MODE,
-    loginRedirectUri: LOGIN_REDIRECT_URI,
     logoutRedirectUri: LOGOUT_MODE === 'silent'
       ? SILENT_LOGOUT_REDIRECT_URI
       : LOGOUT_REDIRECT_URI,
--- a/packages/passport/sdk-sample-app/src/context/ImmutableProvider.tsx
+++ b/packages/passport/sdk-sample-app/src/context/ImmutableProvider.tsx
@@ -12,6 +12,7 @@ import { Passport, PassportModuleConfiguration } from '@imtbl/passport';
 import { Environment, ImmutableConfiguration, ModuleConfiguration } from '@imtbl/config';
 import {
   AUDIENCE,
+  LOGIN_REDIRECT_URI,
   LOGOUT_REDIRECT_URI,
   REDIRECT_URI,
   SILENT_LOGOUT_REDIRECT_URI,
@@ -91,6 +92,7 @@ const getPassportConfig = (environment: EnvironmentNames): PassportModuleConfigu
     audience: AUDIENCE,
     redirectUri: REDIRECT_URI,
     logoutMode: LOGOUT_MODE,
+    loginRedirectUri: LOGIN_REDIRECT_URI,
     logoutRedirectUri: LOGOUT_MODE === 'silent'
       ? SILENT_LOGOUT_REDIRECT_URI
       : LOGOUT_REDIRECT_URI,
--- a/packages/passport/sdk/src/Passport.ts
+++ b/packages/passport/sdk/src/Passport.ts
@@ -198,7 +198,7 @@ export class Passport {
     useCachedSession?: boolean;
     anonymousId?: string;
     useSilentLogin?: boolean;
-    enableRedirectFlow?: boolean;
+    useRedirectFlow?: boolean;
   }): Promise<UserProfile | null> {
     return withMetricsAsync(async () => {
       const { useCachedSession = false, useSilentLogin } = options || {};
@@ -219,7 +219,7 @@ export class Passport {
       if (!user && useSilentLogin) {
         user = await this.authManager.forceUserRefresh();
       } else if (!user && !useCachedSession) {
-        if (options?.enableRedirectFlow) {
+        if (options?.useRedirectFlow) {
           await this.authManager.loginWithRedirect(options?.anonymousId);
         } else {
           user = await this.authManager.login(options?.anonymousId);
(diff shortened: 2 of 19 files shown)


examples/passport/telegram-mini-app/app/utils.ts
```diff
@@ -8,6 +8,8 @@ import { Environment, ImmutableConfiguration } from "@imtbl/sdk/config";
 		clientId: process.env.NEXT_PUBLIC_CLIENT_ID || "<CLIENT_ID>",
 		// The redirect URI set in the application created in Hub
 		redirectUri: process.env.NEXT_PUBLIC_REDIRECT_URI || "<REDIRECT_URI>",
+		// The login redirect URI set in the application created in Hub
+		loginRedirectUri: process.env.NEXT_PUBLIC_LOGIN_REDIRECT_URI || "<LOGIN_REDIRECT_URI>",
```
imx-mikhala, line 12: can you remove the changes made to the `telegram-mini-app` folder? This folder is an example of how to use passport in a telegram environment and as part of the documentation we state to use the `loginWithDeviceFlow` method which will use the `redirectUri` field

packages/passport/sdk-sample-app/src/context/ImmutableProvider.tsx
```diff
@@ -91,6 +92,7 @@ const getPassportConfig = (environment: EnvironmentNames): PassportModuleConfigu
     audience: AUDIENCE,
     redirectUri: REDIRECT_URI,
     logoutMode: LOGOUT_MODE,
+    loginRedirectUri: LOGIN_REDIRECT_URI,
```
imx-mikhala, line 95: I think we should name this something else. Its currently not clear the difference between `redirectUri` and `loginRedirectUri`. I'll have a think of other names after reviewing the rest of the PR, but let me know if you have any ideas.

examples/passport/telegram-mini-app/app/utils.ts
rodrigo-fournier-immutable: Sure, I was unsure if this was actually relevant, but I can remove my last commit since it was solely focused on that.

packages/passport/sdk-sample-app/src/context/ImmutableProvider.tsx
rodrigo-fournier-immutable: I was thinking about this over the weekend. Based on what we saw in the OIDC UserManagerSettings, we could align the naming by renaming what we currently call redirectUri to popupRedirectUri and loginRedirectUri to redirectUri. What do you think?

imx-mikhala: Although that would be ideal this is a breaking change as consumers are already using the redirectUri for their popup redirects and so they expect this to work for the popup flow.

We could make `redirectUri` work for the popup if `popupRedirectUri` is missing though, which might solve this.. 
Basically if both popupRedirectUri and redirectUri are both provided, then the popup will favour the popupRedirectUri. I'll have a think about it a bit more as that approach might make the code more confusing

packages/passport/sdk/src/authManager.ts
```diff
@@ -263,11 +276,19 @@ export default class AuthManager {
...
+        await this.userManager.signinCallback();
+        return undefined;
+      }
+
+      const oidcUser = await this.userManager.signinRedirectCallback();
+      if (!oidcUser) {
+        throw new Error('Failed to obtain user');
+      }
+      return AuthManager.mapOidcUserToDomainModel(oidcUser);
+    }, PassportErrorType.AUTHENTICATION_ERROR);
```
imx-mikhala, line 291: I think we should just simplify this, and not have the `enableRedirectFlow` flag.

The docs actually suggest to use `signinCallback` for redirects https://authts.github.io/oidc-client-ts/classes/UserManager.html#signinRedirectCallback

We can do the following instead:

```
public async loginCallback(): Promise<void | User> {
    return withPassportError<void | User>(async () => {
      const oidcUser = await this.userManager.signinCallback();
      if (!oidcUser) {
        return;
      }
      console.log('oidcUser', oidcUser);
      return AuthManager.mapOidcUserToDomainModel(oidcUser);
    }, PassportErrorType.AUTHENTICATION_ERROR);
  }
```

rodrigo-fournier-immutable: Refactored

imx-mikhala: Thanks, yeah thats fine

imx-mikhala: Wonder if we should make the function return `User | undefined` then to be consistent. Since it was only returning void before I dont think it would break anything to update the return

packages/passport/sdk/src/Passport.ts
```diff
@@ -236,8 +241,16 @@ export class Passport {
    * Handles the login callback.
    * @returns {Promise<void>} A promise that resolves when the callback is processed
    */
-  public async loginCallback(): Promise<void> {
-    return withMetricsAsync(() => this.authManager.loginCallback(), 'loginCallback');
+  public async loginCallback(enableRedirectFlow?: boolean): Promise<void> {
```
imx-mikhala, line 244: We can remove the `enableRedirectFlow` flag here. user should be undefined in the non redirect flow anyway

packages/passport/sdk/src/authManager.ts
rodrigo-fournier-immutable: Done

packages/passport/sdk/src/Passport.ts
rodrigo-fournier-immutable: Done

packages/passport/sdk/src/Passport.ts
```diff
@@ -198,6 +198,7 @@ export class Passport {
     useCachedSession?: boolean;
     anonymousId?: string;
     useSilentLogin?: boolean;
+    enableRedirectFlow?: boolean;
```
imx-mikhala, line 201: Lets rename this to `useRedirectFlow` as its a bit more consistent with the other params

packages/passport/sdk/src/authManager.ts
```diff
@@ -63,8 +63,8 @@ const getAuthConfiguration = (config: PassportConfiguration): UserManagerSetting
 
   const baseConfiguration: UserManagerSettings = {
     authority: authenticationDomain,
-    redirect_uri: oidcConfiguration.redirectUri,
-    popup_redirect_uri: oidcConfiguration.redirectUri,
+    redirect_uri: oidcConfiguration.redirectUri || '',
+    popup_redirect_uri: oidcConfiguration.popupRedirectUri,
```
imx-mikhala, line 67: Can we make this

`popup_redirect_uri: oidcConfiguration.popupRedirectUri || oidcConfiguration.redirectUri` so it defaults to the previous value? During testing I ran into this issue when the param was passed but empty string, so I think we should just set it to redirectUri when popupRedirectUri is not set

`AUTHENTICATION_ERROR - No popup_redirect_uri configured`

packages/passport/sdk/src/authManager.ts
```diff
@@ -63,8 +63,8 @@ const getAuthConfiguration = (config: PassportConfiguration): UserManagerSetting
 
   const baseConfiguration: UserManagerSettings = {
     authority: authenticationDomain,
-    redirect_uri: oidcConfiguration.redirectUri,
-    popup_redirect_uri: oidcConfiguration.redirectUri,
+    redirect_uri: oidcConfiguration.redirectUri || '',
```
imx-mikhala, line 66: Dont thiink we need the || here as its not optional

PR #2547, opened 2025-02-24
fix: Fixed SDK parser

darrenmelvison1:
## Description

FIxed the SDK parser script to ensure that no unknown methods are introduced to the JSON files, also methods to only be nested inside their respective classes with all of their params having the correct type (except for those that returns an Interface/Enum, they'll be flagged as unknown for now).

For now it's just for classes and methods.

JIRA: https://immutable.atlassian.net/browse/DVR-311


ImmutableAustinG (approved):
Changes look good - no issues that I can see

PR #2548, opened 2025-02-24
build(checkout): GFI-96: Track select token clicks

keithbro-imx:

- [x] PR is titled with conventional commit style naming: `type(scope): message`. For example: `feat(passport): my new feature`

# Summary

We track when the user chooses a token - event for that is checkoutSwapSwapCoins_SelectFromSelect and checkoutSwapSwapCoins_SelectToSelect, but we don't track the interaction before that which is opening the select menu. This PR tracks that.

## Security

### DIFF
--- a/packages/checkout/widgets-lib/src/widgets/bridge/components/BridgeForm.tsx
+++ b/packages/checkout/widgets-lib/src/widgets/bridge/components/BridgeForm.tsx
@@ -326,6 +326,9 @@ export function BridgeForm(props: BridgeFormProps) {
         {(!defaultTokenAddress || !isTokenBalancesLoading) && (
           <Box sx={formInputsContainerStyles}>
             <SelectForm
+              userJourney={UserJourney.BRIDGE}
+              screen="TokenAmount"
+              control="FromToken"
               testId="bridge-token"
               options={tokensOptions}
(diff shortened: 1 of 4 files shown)


packages/checkout/widgets-lib/src/widgets/bridge/components/BridgeForm.tsx
```diff
@@ -326,6 +326,9 @@ export function BridgeForm(props: BridgeFormProps) {
         {(!defaultTokenAddress || !isTokenBalancesLoading) && (
           <Box sx={formInputsContainerStyles}>
             <SelectForm
+              userJourney={UserJourney.BRIDGE}
+              screen="TokenAmount"
```
jeetparikh, line 330: is the screen type right here?

jeetparikh (approved):

shirren (approved):

packages/checkout/widgets-lib/src/widgets/bridge/components/BridgeForm.tsx
keithbro-imx: Answered in Slack, but for completeness, this matches the screen value used elsewhere in this file.

PR #2550, opened 2025-02-24
build(checkout): Track spending approval separate from swap approval

keithbro-imx:

- [x] PR is titled with conventional commit style naming: `type(scope): message`. For example: `feat(passport): my new feature`

# Summary

Currently we track these two screens with the same event, making it difficult to represent the user flow accurately in a funnel.

image
image

Use different events for each.

## Security


pano-skylakis (approved):
Might need to give Stefan a heads up! Not sure where this is tracked currently

jeetparikh (approved):

PR #2566, opened 2025-02-27
feat(checkout): Add a reverse button to the swap form to reverse the buy and sell tokens

jeetparikh:
- [✅  ] PR is titled with conventional commit style naming: `type(scope): message`. For example: `feat(passport): my new feature`

# Summary
Add a reverse button to the swap form to reverse the buy and sell tokens


packages/checkout/widgets-lib/src/widgets/swap/components/SwapForm.tsx
```diff
@@ -509,6 +510,7 @@ export function SwapForm({ data, theme, cancelAutoProceed }: SwapFromProps) {
         resetQuote();
         return;
       }
+
       (async () => await fetchQuote())();
```
keithbro-imx, line 514: wosaldisden?

keithbro-imx (approved):
Looks good. As discussed I found two things:

1. the button doesn't work exactly the same as quickswap/uniswap. In those implementations, the amount you entered always moves to the other side, whereas here the buy amount always becomes the sell amount.
2. If you click the button before the quote is ready you lose your amounts because 0 gets moved up.

shirren (approved):

PR #2573, opened 2025-03-05
feat(checkout): Not clear form values on error and show errors directly

jeetparikh:
# Summary
No longer clears all form values on quote error drawer closing. Quote error is shown directly below the form and not in the drawer.


packages/checkout/widgets-lib/rollup.config.js
```diff
@@ -10,54 +10,37 @@ const PRODUCTION = 'production';
```
jeetparikh, line 10: some build error. suggested by Zaid to change this

packages/checkout/widgets-lib/src/widgets/swap/components/SwapForm.tsx
```diff
@@ -217,9 +216,8 @@ export function SwapForm({ data, theme, cancelAutoProceed }: SwapFromProps) {
...
   ]);
 
-  const openUnableToSwapDrawer = useCallback((error: any) => {
-    setShowNotEnoughImxDrawer(false);
-    setShowUnableToSwapDrawer(true);
+  const onQuoteError = useCallback((error: any) => {
+    setQuoteError(error.message);
     track({
       userJourney: UserJourney.SWAP,
       screen: 'SwapCoins',
```
jeetparikh, line 223: I wonder if this track event control should still be called `UnableToSwapDrawer` or we call it different and update the looker dashboard to include both?

keithbro-imx (approved):

zaidarain1 (approved):

PR #2583, opened 2025-03-13
fix(checkout): Fix checkout / bridge regressions in v2

keithbro-imx:
# Summary

Fixes a couple of bugs in Checkout that surfaced in the Bridge widget.

* If your EOA wallet is not on the correct provider, you won't see any balances when trying to bridge.
* Sometimes, you get an error when choosing Passport from the list because we convert it to a WrappedBrowserProvider too early in the process.

## Security

### DIFF
--- a/packages/checkout/sdk/src/sdk.ts
+++ b/packages/checkout/sdk/src/sdk.ts
@@ -366,7 +366,7 @@ export class Checkout {
   public async getAllBalances(
     params: GetAllBalancesParams,
   ): Promise<GetAllBalancesResult> {
-    const browserProvider = !params.provider || (params.provider instanceof WrappedBrowserProvider)
+    const browserProvider = (!params.provider || ('ethereumProvider' in params.provider))
       ? params.provider
       : new WrappedBrowserProvider(params.provider);
 
--- a/packages/checkout/sdk/src/sdk.ts
+++ b/packages/checkout/sdk/src/sdk.ts
@@ -366,19 +366,9 @@ export class Checkout {
   public async getAllBalances(
     params: GetAllBalancesParams,
   ): Promise<GetAllBalancesResult> {
-    if (!params.provider) {
-      return balances.getAllBalances(
-        this.config,
-        params.provider,
-        params.walletAddress,
-        params.chainId,
-      );
-    }
-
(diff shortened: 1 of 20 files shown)


packages/checkout/sdk/src/sdk.ts
```diff
@@ -366,19 +366,9 @@ export class Checkout {
...
-        params.walletAddress,
-        params.chainId,
-      );
-    }
-
-    const browserProvider = await provider.validateProvider(
-      this.config,
-      params.provider,
-    );
+    const browserProvider = !params.provider || (params.provider instanceof WrappedBrowserProvider)
```
keithbro-imx, line 369: Fix 1: If the network on the provider is not an allowed one, we still try to go ahead with fetching balances from blockscout, using `params.chainId` to determine the network.

Note that there is an existing behaviour in v1 too, where if the provider's network != params.chainID AND Blockscout is unavailable then we will use the network of the provider to fetch balances, ignoring params chain ID.

packages/checkout/widgets-lib/src/lib/hooks/useInjectedProviders.ts
```diff
@@ -52,13 +49,10 @@ const processProviders = async (
   // Add passport from checkout config if not from injected providers
   if (checkout?.passport
     && priorityWalletRdns.includes(WalletProviderRdns.PASSPORT)
-    && !filteredProviders.some((provider) => provider.info.rdns === WalletProviderRdns.PASSPORT)) {
-    if (!passportBrowserProvider) {
-      // eslint-disable-next-line max-len
-      passportBrowserProvider = new WrappedBrowserProvider(await checkout.passport.connectEvm());
-    }
-    // eslint-disable-next-line max-len
-    filteredProviders.unshift(getPassportProviderDetail(passportBrowserProvider.provider as unknown as EIP1193Provider));
```
keithbro-imx, line 61: Fix two: `passportBrowserProvider.provider` is a WrappedBrowserProvider which is not what we want to put in to filteredProviders. Don't use type assertions unless you absolutely have to!!

zaidarain1 (approved):

w3njah (approved):

jeetparikh (approved):

PR #2599, opened 2025-03-26
fix: DVR-330 | Changed the mechanism on example app's sidebar_orders for consistency 

darrenmelvison1:
## Description

Changed how sidebar_orders for example apps are assigned to the metadata.json file. This fix is to ensure that updates to the order can be done via the parser instead of having to run the prompt.


proletesseract (approved):
LGTM

zaidarain1 (approved):

PR #2604, opened 2025-03-31
fix(checkout): Allow all decimals for max transfer

keithbro-imx:
# Summary

Two changes:

1. When clicking "max" - it puts in the full amount so that dust is not left in the wallet.
2. Once transfer completes, the balances are refreshed so that they reflect the transfer that just happened.

## Security

### DIFF
--- a/packages/checkout/widgets-lib/src/widgets/transfer/TransferWidget.tsx
+++ b/packages/checkout/widgets-lib/src/widgets/transfer/TransferWidget.tsx
@@ -81,23 +81,11 @@ function TransferWidgetInner(props: TransferWidgetInputs) {
     };
 
     x();
-  }, [checkout]);
+  }, [checkout, provider, cryptoFiatDispatch, props.amount, props.tokenAddress, props.toAddress, viewState.type]);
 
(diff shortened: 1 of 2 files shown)


packages/checkout/widgets-lib/src/widgets/transfer/TransferWidget.tsx
```diff
@@ -81,23 +81,11 @@ function TransferWidgetInner(props: TransferWidgetInputs) {
     };
 
     x();
-  }, [checkout]);
+  }, [checkout, provider, cryptoFiatDispatch, props.amount, props.tokenAddress, props.toAddress, viewState.type]);
```
jeetparikh, line 84: wow this were all missing

jeetparikh (approved):

packages/checkout/widgets-lib/src/widgets/transfer/TransferWidget.tsx
keithbro-imx: I'll make a ticket to enable the react hook linting

PR #2607, opened 2025-03-31
chore(checkout): Additional overrides to facilitate the Immutable dev env

keithbro-imx:
# Summary
Allows overrides to be passed to the checkout constructor so that it can be used in envs other than sandbox and prod.

## Security


packages/checkout/sdk/src/balances/balances.ts
```diff
@@ -238,20 +238,19 @@ export const getBalances = async (
...
-    // For some reason isNativeToken always returns undefined.
-    // We have spent way too much time figuring out why this is happening.
-    // That we have given up -- keep it as it is for now.
-    if (!token.address || isMatchingAddress(token.address, NATIVE)) resp.value.token.address = NATIVE;
-    return resp.value;
-  });
 
-  return { balances };
+  return balanceResults.reduce((acc, resp) => {
+    if (resp.status !== 'fulfilled' || resp.value.balance === 0n) return acc;
```
keithbro-imx, line 243: The change here is to filter out any zero balances. When fetching balances from the RPC it returns all tokens, even zero balance ones. So we need to filter these out to maintain the same behaviour as when we use blockscout.

packages/checkout/sdk/src/config/config.ts
```diff
@@ -91,18 +110,23 @@ export class CheckoutConfiguration {
     this.isBridgeEnabled = config.bridge?.enable ?? DEFAULT_BRIDGE_ENABLED;
     this.publishableKey = config.publishableKey ?? '<no-publishable-key>';
 
-    this.networkMap = networkMap(this.isProduction, this.isDevelopment);
+    this.networkMap = config.overrides?.networkMap ?? networkMap(this.isProduction, this.isDevelopment);
```
keithbro-imx, line 113: These changes are mostly about reading from the overrides

packages/checkout/widgets-sample-app/src/utils/passport.ts
```diff
@@ -1,10 +1,13 @@
 import { Passport } from "@imtbl/passport";
 import { Environment } from "@imtbl/config";
+import * as x from "@imtbl/x-client";
```
keithbro-imx, line 3: Not needed.

allan-almeida-imtbl (approved):

PR #2610, opened 2025-04-02
fix(StarkEx): Revert StarkEx Ethers v6

zaidarain1:


packages/checkout/sdk/tsup.config.js
```diff
@@ -33,7 +33,7 @@ export default defineConfig((options) => {
             Buffer: true,
             process: true,
           },
-          modules: ['crypto', 'buffer', 'process']
```
keithbro-imx, line 36: 👋 What does this do?

zaidarain1: It let's the polyfill plugin handle the polyfills, for some reason crypto isn't getting polyfilled for a specific package even if it is being specified here, removing this line fixes that issue

PR #2620, opened 2025-04-14
fix(checkout): Fetch crypto to USD conversion via our own subgraph

jeetparikh:
We use our own subgraph to get accurate USD pricing for crypto on zkEVM as compared to coingecko API that favours L1 pools.

This PR only changes the Swap Widget to use this new API


packages/checkout/widgets-lib/src/lib/hooks/useCryptoUSDConversion.ts
```diff
@@ -0,0 +1,32 @@
...
+  useEffect(() => {
+    async function fetchConversions() {
+      if (!environment) return;
+
+      try {
+        const data = await getCryptoToUSDConversion(environment);
+        setConversions(data);
+        setError(null);
+      } catch (err) {
+        setError(err instanceof Error ? err : new Error('Failed to fetch conversions'));
```
keithbro-imx, line 19:
```suggestion
        setError(err instanceof Error ? err : new Error('Failed to fetch conversions: '+err));
```

keithbro-imx (dismissed):

keithbro-imx (approved):

PR #2622, opened 2025-04-16
fix(checkout): Show Price Surge screen when slippage exceeded

keithbro-imx:
# Summary

Show Price Surge screen when slippage is exceeded. This behaviour regressed in v2.

## Security

### DIFF
--- a/packages/checkout/widgets-lib/src/widgets/swap/views/ApproveERC20Onboarding.tsx
+++ b/packages/checkout/widgets-lib/src/widgets/swap/views/ApproveERC20Onboarding.tsx
@@ -2,8 +2,9 @@ import { Box } from '@biom3/react';
 import {
   useCallback, useContext, useEffect, useMemo, useState,
 } from 'react';
-import { CheckoutErrorType, TokenInfo } from '@imtbl/checkout-sdk';
+import { CheckoutError, CheckoutErrorType, TokenInfo } from '@imtbl/checkout-sdk';
 import { useTranslation } from 'react-i18next';
+import { TransactionRequest } from 'ethers';
 import { SimpleLayout } from '../../../components/SimpleLayout/SimpleLayout';
 import { HeaderNavigation } from '../../../components/Header/HeaderNavigation';
 import { sendSwapWidgetCloseEvent } from '../SwapWidgetEvents';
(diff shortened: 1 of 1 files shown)


packages/checkout/widgets-lib/src/widgets/swap/views/ApproveERC20Onboarding.tsx
```diff
@@ -96,8 +97,10 @@ export function ApproveERC20Onboarding({ data }: ApproveERC20Props) {
     });
   }, [viewDispatch]);
 
-  const handleExceptions = (err, swapFormData:PrefilledSwapForm) => {
-    if (err.type === CheckoutErrorType.UNPREDICTABLE_GAS_LIMIT) {
+  const handleExceptions = (err: any, swapFormData: PrefilledSwapForm) => {
+    if (err instanceof CheckoutError
+      && err.type === CheckoutErrorType.TRANSACTION_FAILED
+      && err.message.startsWith('execution reverted: "STF"')) {
```
keithbro-imx, line 103: Not great but the best we can do. The ethers error is not easy to parse.

packages/checkout/widgets-lib/src/widgets/swap/views/ApproveERC20Onboarding.tsx
```diff
@@ -138,7 +141,7 @@ export function ApproveERC20Onboarding({ data }: ApproveERC20Props) {
     });
   };
 
-  const prepareTransaction = (transaction, isGasFree = false) => ({
+  const prepareTransaction = (transaction: TransactionRequest, isGasFree = false) => ({
```
keithbro-imx, line 144: Adding the type to this. Surprised this isn't enforced.

packages/checkout/widgets-lib/src/widgets/swap/views/ApproveERC20Onboarding.tsx
```diff
@@ -164,7 +167,7 @@ export function ApproveERC20Onboarding({ data }: ApproveERC20Props) {
       showErrorView();
       return;
     }
-    if (actionDisabled) return;
+    if (actionDisabled || !data.approveTransaction) return;
```
keithbro-imx, line 170: After adding the type we need to handle this case.

packages/checkout/widgets-lib/src/widgets/swap/views/ApproveERC20Onboarding.tsx
```diff
@@ -199,7 +202,7 @@ export function ApproveERC20Onboarding({ data }: ApproveERC20Props) {
         setRejectedSpending(true);
         return;
       }
-      handleExceptions(err, data.swapFormInfo as PrefilledSwapForm);
+      handleExceptions(err, data.swapFormInfo);
```
keithbro-imx, line 205: Remove unnecessary type assertion

jeetparikh (approved):

PR #2623, opened 2025-04-16
chore(checkout): Add pair info to price surge event

keithbro-imx:
# Summary

Adds the pair (token addresses) to the price surge event for data tracking.

## Security


packages/checkout/widgets-lib/src/widgets/swap/SwapWidget.tsx
```diff
@@ -353,6 +353,7 @@ export default function SwapWidget({
                 page({
                   userJourney: UserJourney.SWAP,
                   screen: 'PriceSurge',
+                  extras: viewState.view.data,
```
keithbro-imx, line 356: Contains: 

```
  fromAmount: string;
  fromTokenAddress: string;
  toTokenAddress: string;
  toAmount?: string;
```

jeetparikh (approved):

PR #2646, opened 2025-06-03
feat(checkout): Return price impact as part of the quote

keithbro-imx:

# Summary

Return price impact of the trade as part of the quote. 

Note that this is the price impact as calculated by comparing the midprice of the route and the execution price of the quote. It does not look at or consider the price impact on any of the pools individually.

## Security

### DIFF
--- a/packages/checkout/widgets-lib/src/widgets/swap/functions/swapFees.test.ts
+++ b/packages/checkout/widgets-lib/src/widgets/swap/functions/swapFees.test.ts
@@ -24,6 +24,7 @@ describe('formatSwapFees', () => {
         amountWithMaxSlippage: {} as Amount,
         slippage: 0,
         fees: [],
+        priceImpact: 0,
       } as Quote,
       swap: {
         gasFeeEstimate: {


packages/internal/dex/sdk/src/lib/getQuotesForRoutes.ts
```diff
@@ -19,14 +20,51 @@ export type QuoteResult = {
...
 };
 
+export function parseQuoteResult(
+  quoteResult: Result,
+  route: Route<Token, Token>,
+  amountSpecified: CoinAmount<ERC20>,
+  tradeType: TradeType,
+) {
+  const quoteAmount = quoteResult[amountIndex];
+  if (typeof quoteAmount !== 'bigint') throw new Error('Expected BigNumber');
```
jeetparikh, line 33: could say `expect bignumber but got `${typepof quoteAmont}` - helps with debugging

jeetparikh (approved):

packages/internal/dex/sdk/src/lib/getQuotesForRoutes.ts
keithbro-imx: true! this isn't new code fwiw, but i do agree.. i can update it in my next pr...

PR #2647, opened 2025-06-05
[ID-3716] feat: revoke refresh token on logout

ImmutableJeffrey:

# Summary

Revoke refresh token on logout

## Security

### DIFF
--- a/packages/passport/sdk/src/authManager.ts
+++ b/packages/passport/sdk/src/authManager.ts
@@ -71,11 +71,13 @@ const getAuthConfiguration = (config: PassportConfiguration): UserManagerSetting
       token_endpoint: `${authenticationDomain}/oauth/token`,
       userinfo_endpoint: `${authenticationDomain}/userinfo`,
       end_session_endpoint: endSessionEndpoint.toString(),
+      revocation_endpoint: `${authenticationDomain}/oauth/revoke`,
     },
     mergeClaims: true,
     automaticSilentRenew: false, // Disabled until https://github.com/authts/oidc-client-ts/issues/430 has been resolved
     scope: oidcConfiguration.scope,
     userStore,
+    revokeTokenTypes: ['refresh_token'],
     extraQueryParams: {
       ...config.extraQueryParams,
       ...(oidcConfiguration.audience ? { audience: oidcConfiguration.audience } : {}),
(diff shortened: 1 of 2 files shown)


packages/passport/sdk/src/authManager.ts
```diff
@@ -435,16 +435,44 @@ export default class AuthManager {
     return response.data;
   }
 
+  private async revokeRefreshToken(refreshToken: string): Promise<void> {
```
nattb8, line 438: oidc-client-ts has revokeTokens function, does that work?

packages/passport/sdk/src/types.ts
```diff
@@ -56,6 +56,7 @@ export type PassportMetadata = {
 
 export interface OidcConfiguration {
   clientId: string;
+  clientSecret?: string;
```
haydenfowler, line 59: We don't expose this to partners, but even if we did I don't think that this can be exposed on the client-side

packages/passport/sdk/src/authManager.ts
```diff
@@ -436,15 +438,22 @@ export default class AuthManager {
...
-    );
+    return withPassportError<void>(async () => {
+      const user = await this.userManager.getUser();
+      if (!user) {
+        return;
+      }
+
+      if (user.refresh_token) {
+        await this.userManager.revokeTokens(['refresh_token']);
+      }
```
haydenfowler, line 449: Refresh tokens are needed for the registration process, so it's probably safe to assume that all users will have a refresh token 🤔 It also looks like the library we're using won't do anything if the refresh token isn't present: [here](https://github.com/authts/oidc-client-ts/blob/676b86f849773ba54d1e857a3a84f07aa5e768c4/src/UserManager.ts#L734). 

We can probably simplify this by removing the `getUser` call and the `refresh_token` check.

packages/passport/sdk/src/authManager.ts
```diff
@@ -436,15 +438,22 @@ export default class AuthManager {
...
+
+      if (this.logoutMode === 'silent') {
+        await this.userManager.signoutSilent();
+      } else {
+        await this.userManager.signoutRedirect();
+      }
```
haydenfowler, line 455: We could improve performance a bit by parallelising the requests when `signoutSilent` is being used, e.g:
```suggestion
      if (this.logoutMode === 'silent') {
        await Promise.all([
          this.userManager.revokeTokens(['refresh_token']),
          this.userManager.signoutSilent(),
        ]);
      } else {
        await this.userManager.revokeTokens(['refresh_token']);
        await this.userManager.signoutRedirect();
      }
```

PR #2649, opened 2025-06-05
feat(checkout): Handle network switch when using walletconnect

jeetparikh:
WalletConnect switch wasn't working when on the wrong network. This now forces the 'Switch network' screen requiring user to interact in order to switch the network.


packages/checkout/widgets-lib/src/widgets/connect/components/WalletList.tsx
```diff
@@ -218,20 +218,14 @@ export function WalletList(props: WalletListProps) {
       const browserProvider = new WrappedBrowserProvider(ethereumProvider);
       selectBrowserProvider(browserProvider, 'walletconnect');
```
keithbro-imx, line 219: Do we need to do this?

keithbro-imx: Never mind, I see why we do

PR #2657, opened 2025-07-04
fix(passport): update access token expiry check

CodeSchwert:

# Summary

[ID-3828](https://immutable.atlassian.net/browse/ID-3828)

- Force token refresh if access token is going to expire (within 30 secs).

## Security

[ID-3828]: https://immutable.atlassian.net/browse/ID-3828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/passport/sdk/src/utils/token.ts
+++ b/packages/passport/sdk/src/utils/token.ts
@@ -2,7 +2,7 @@ import jwt_decode from 'jwt-decode';
 import {
   User as OidcUser,
 } from 'oidc-client-ts';
-import { IdTokenPayload } from '../types';
+import { IdTokenPayload, TokenPayload } from '../types';
 
 export function isIdTokenExpired(idToken: string | undefined): boolean {
   if (!idToken) {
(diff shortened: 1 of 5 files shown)


nattb8 (approved):

packages/passport/sdk/src/utils/token.ts
```diff
@@ -14,10 +14,21 @@ export function isIdTokenExpired(idToken: string | undefined): boolean {
   return decodedToken.exp < now;
 }
 
-export function isTokenExpired(oidcUser: OidcUser): boolean {
-  const { id_token: idToken, expired } = oidcUser;
+export function isAccessTokenExpiredOrExpiring(oidcUser: OidcUser): boolean {
+  const { id_token: idToken, expired, expires_in } = oidcUser;
```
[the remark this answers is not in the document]
CodeSchwert: `expired` is the based off `expires_in`, both relate to the access token expiry: https://github.com/authts/oidc-client-ts/blob/fcb067dc91c180537009edef21c7995a300e20fc/src/User.ts#L96

packages/passport/sdk/src/utils/token.ts
```diff
@@ -14,10 +14,32 @@ export function isIdTokenExpired(idToken: string | undefined): boolean {
...
+    // Check if access token is expired or expiring in 30 seconds or less
+    if (decodedAccessToken.exp <= now + 30) {
+      return true;
+    }
+  } catch (error) {
+    // If we can't decode the access token, assume it's invalid
+    return true;
+  }
+
   return isIdTokenExpired(idToken);
```
haydenfowler, line 44: Do you think it's worth us applying the same 30s expiration check to the `idToken`?

haydenfowler: IMO it's better to be safe than sorry here.

The SDK originally had the assumption that the ID & access token lifetimes would be the same, and things broke when that was no longer the case... We can assume that the ID token lifetime will always be >= the access token lifetime, but that may change at some point in the future.

packages/passport/sdk/src/utils/token.ts
```diff
@@ -2,22 +2,40 @@ import jwt_decode from 'jwt-decode';
...
+  }
 }
 
-export function isTokenExpired(oidcUser: OidcUser): boolean {
-  const { id_token: idToken, expired } = oidcUser;
-  if (expired) {
+export function isAccessTokenExpiredOrExpiring(oidcUser: OidcUser): boolean {
+  const { id_token: idToken, access_token: accessToken } = oidcUser;
+
+  // Handle missing tokens - assume they need to login again
```
haydenfowler, line 29:
```suggestion
  // Handle missing tokens - assume they need to login again
```

Returning `true` doesn't trigger a login

packages/passport/sdk/src/utils/token.ts
```diff
@@ -2,22 +2,40 @@ import jwt_decode from 'jwt-decode';
...
   }
-  return isIdTokenExpired(idToken);
+
+  // Check if access token is expired or expiring
+  if (isTokenExpiredOrExpiring(accessToken)) {
+    return true;
+  }
+
+  // Check if ID token is expired or expiring
+  return isTokenExpiredOrExpiring(idToken);
```
haydenfowler, line 40: This can be simplified

```suggestion
  // Check if access token is expired or expiring
  return isTokenExpiredOrExpiring(accessToken) || isTokenExpiredOrExpiring(idToken);
```

haydenfowler (approved):

PR #2666, opened 2025-07-18
fix(checkout-sdk): Unrecognized chain error handling

allan-almeida-imtbl:
# Summary

Fixes the error handling case upon receving a Metamask error when switching to an unrecognized network is detected, in order to add that network and then switch to it. 

The expected error shape has been added to with an `||`, to safely handle both error shapes if they may occur.

Allow again users who do not have zkEVM network in their wallet to connect more smoothly.

## Fixed

- Metamask error handling for unrecogized chain error
- Connect flow for users who have not yet added zkEVM network to their wallet

### DIFF
--- a/packages/checkout/sdk/src/network/network.ts
+++ b/packages/checkout/sdk/src/network/network.ts
@@ -136,7 +136,8 @@ export async function getNetworkInfo(
  * You get "Error: could not coalesce error....".
  * This is a workaround to check if the error is an unrecognised chain error.
  * */
-const isUnrecognisedChainError = (err: any) => err.error?.data?.originalError?.code === UNRECOGNISED_CHAIN_ERROR_CODE;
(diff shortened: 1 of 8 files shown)


packages/checkout/sdk/src/network/network.ts
```diff
@@ -136,7 +136,8 @@ export async function getNetworkInfo(
  * You get "Error: could not coalesce error....".
  * This is a workaround to check if the error is an unrecognised chain error.
  * */
-const isUnrecognisedChainError = (err: any) => err.error?.data?.originalError?.code === UNRECOGNISED_CHAIN_ERROR_CODE;
+const isUnrecognisedChainError = (err: any) => err.error?.data?.originalError?.code === UNRECOGNISED_CHAIN_ERROR_CODE
+  || err.error?.code === UNRECOGNISED_CHAIN_ERROR_CODE;
```
allan-almeida-imtbl, line 140: added as part of OR condition to handle both error shapes

keithbro-imx (approved):

PR #2670, opened 2025-07-22
Migrate to v2 sanction API and add token info

fang377:

# Summary

## Security

### DIFF
--- a/packages/checkout/sdk/src/riskAssessment/riskAssessment.ts
+++ b/packages/checkout/sdk/src/riskAssessment/riskAssessment.ts
@@ -22,9 +22,17 @@ export type AssessmentResult = {
   };
 };
 
+// New type for v2 request items
+type SanctionsCheckV2RequestItem = {
+  address?: string;
+  amount?: string;
+  token_addr?: string;
+};
+
 export const fetchRiskAssessment = async (
   addresses: string[],
   config: CheckoutConfiguration,
+  tokenData?: Array<{ address: string; tokenAddr?: string; amount?: string }>,
 ): Promise<AssessmentResult> => {
   const result = Object.fromEntries(
     addresses.map((address) => [address.toLowerCase(), { sanctioned: false }]),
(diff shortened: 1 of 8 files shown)


packages/checkout/sdk/src/riskAssessment/riskAssessment.ts
```diff
@@ -41,11 +49,25 @@ export const fetchRiskAssessment = async (
   try {
     const riskLevels = riskConfig?.levels.map((l) => l.toLowerCase()) ?? [];
 
+    // Prepare v2 request payload
+    const requestPayload: SanctionsCheckV2RequestItem[] = addresses.map((address) => {
+      const item: SanctionsCheckV2RequestItem = { address };
+
+      // Add token and amount data if available
```
keithbro-imx, line 56: In what scenario would the token and amount data not be available? 

Should we still send through a request to the sanctions service without this information?

fang377: yea the request should still be sent even when token and amount data are not available, 

> In what scenario would the token and amount data not be available?

like the sale context (where token/amount are determined later in the flow), early-stage widget initialization (before user selects token/amount), and generic address validation (where we just want to check if an address is sanctioned regardless of specific transaction details) etc

packages/checkout/sdk/src/riskAssessment/riskAssessment.ts
```diff
@@ -41,11 +49,27 @@ export const fetchRiskAssessment = async (
...
+    // Prepare v2 request payload
+    const requestPayload: SanctionsCheckV2RequestItem[] = addresses.map((address) => {
+      const item: SanctionsCheckV2RequestItem = {
+        address,
+        token_addr: '',
+        amount: '0',
+      };
+
+      // Add token and amount data
+      const tokenInfo = tokenData.find((t) => t.address.toLowerCase() === address.toLowerCase());
```
keithbro-imx, line 61: Looks like if tokenInfo can't be found, we'll send

```
        token_addr: '',
        amount: '0',
```

to the API.. what will happen? Will this cause an error for the user?

packages/checkout/sdk/src/riskAssessment/riskAssessment.ts
```diff
@@ -22,9 +22,17 @@ export type AssessmentResult = {
...
+type SanctionsCheckV2RequestItem = {
+  address: string;
+  amount: string;
+  token_addr: string;
+};
+
 export const fetchRiskAssessment = async (
   addresses: string[],
   config: CheckoutConfiguration,
+  tokenData: Array<{ address: string; tokenAddr: string; amount: string }>,
```
keithbro-imx, line 35: Is `address` here the same as the address is `addresses` - can we remove one of them so that the caller doesn't have the provide the same info twice?

packages/checkout/widgets-lib/src/widgets/add-tokens/views/AddTokens.tsx
```diff
@@ -425,7 +425,15 @@ export function AddTokens({
   const sendRequestOnRampEvent = async () => {
     if (
       toAddress
-      && (await checkSanctionedAddresses([toAddress], checkout.config))
+      && (await checkSanctionedAddresses(
+        [toAddress],
+        checkout.config,
+        [{
+          address: toAddress,
+          tokenAddr: selectedToken?.address || '',
```
keithbro-imx, line 433: Should we be checking that a token has been selected before calling this - rather than going ahead with the empty string?

packages/checkout/sdk/src/riskAssessment/riskAssessment.ts
fang377: the token_addr and amount are not used in determining the risk level actually so there should not be any impact, but let me update the logic here to only call api when token info's valid

PR #2692, opened 2025-09-04
feat(checkout): update swap widget to use sanctions v2 check

allan-almeida-imtbl:
# Summary

- Added seperate call to V2 sanctions API in risk assessment lib
- Updated swaps widget to call new sanctions check as part of the Swap flow 
  - Now that the check is being consumed as soon as it is done, there is no need to store the result in state and we can therefore remove

### DIFF
--- a/packages/checkout/sdk/src/riskAssessment/common.ts
+++ b/packages/checkout/sdk/src/riskAssessment/common.ts
@@ -1,0 +1,29 @@
+export type RiskAssessmentResponse = {
+  address: string;
+  risk: RiskAssessmentLevel;
+  risk_reason: string;
+};
+
+export enum RiskAssessmentLevel {
+  LOW = 'Low',
+  MEDIUM = 'Medium',
+  HIGH = 'High',
+  SEVERE = 'Severe',
+}
+
+export type AssessmentResult = {
+  [address: string]: {
+    sanctioned: boolean;
+  };
+};
+
+export const isAddressSanctioned = (
+  riskAssessment: AssessmentResult,
+  address?: string,
+): boolean => {
+  if (address) {
+    return riskAssessment[address.toLowerCase()].sanctioned;
+  }
+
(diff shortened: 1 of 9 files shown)


packages/checkout/widgets-lib/src/widgets/swap/components/SwapForm.tsx
```diff
@@ -855,6 +847,36 @@ export function SwapForm({ data, theme, cancelAutoProceed }: SwapFromProps) {
...
+    if (!address) {
+      return;
+    }
+
+    // As this is post form data validation, the fallback values is unecessary and only to satisfy type correctess.
+    // Ideally once the form data is validated it should be converted into a not null type to reflect its validity.
+    const riskAssessmentData = [{
+      address,
+      tokenAddr: data?.fromTokenAddress || '',
+      amount: data?.fromAmount || '',
```
keithbro-imx, line 862: Is this a formatted value or a wei value?

packages/checkout/sdk/src/riskAssessment/common.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export type AssessmentResult = {
+  [address: string]: {
+    sanctioned: boolean;
+  };
+};
+
+export const isAddressSanctioned = (
+  riskAssessment: AssessmentResult,
+  address?: string,
```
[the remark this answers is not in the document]
allan-almeida-imtbl: updated to seperate dedicated functions as discussed 👍

packages/checkout/widgets-lib/src/widgets/swap/components/SwapForm.tsx
allan-almeida-imtbl: checked and updated to use wei amount

keithbro-imx (approved):

keithbro-imx (approved):

PR #2695, opened 2025-09-04
GFI-1059: Include amount in add tokens check

keithbro-imx:

# Summary

## Security

### DIFF
--- a/packages/checkout/widgets-lib/src/functions/checkSanctionedAddresses.ts
+++ b/packages/checkout/widgets-lib/src/functions/checkSanctionedAddresses.ts
@@ -1,13 +1,24 @@
 import {
   CheckoutConfiguration,
-  fetchRiskAssessment,
+  fetchRiskAssessmentV2,
   isAddressSanctioned,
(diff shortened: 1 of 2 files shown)


packages/checkout/widgets-lib/src/functions/checkSanctionedAddresses.ts
```diff
@@ -1,13 +1,24 @@
 import {
   CheckoutConfiguration,
-  fetchRiskAssessment,
-  isAddressSanctioned,
+  fetchRiskAssessmentV2,
+  resultHasSanctionedWallets,
 } from '@imtbl/checkout-sdk';
 
+type TokenAmount = {
```
allan-almeida-imtbl, line 7: not a blocker, but you could do something like 

import { ...,  AssessmentData } from '@imtbl/checkout-sdk';

 Omit<AssessmentData, 'address'>

to avoid creating this intermediary type

allan-almeida-imtbl (approved):

PR #2715, opened 2025-09-24
feat(checkout-widgets): Allow blocklistWalletRdns to be used with non-Connect widgets

keithbro-imx:

# Summary

## Security

### DIFF
--- a/packages/checkout/widgets-lib/src/components/ConnectLoader/ConnectLoader.tsx
+++ b/packages/checkout/widgets-lib/src/components/ConnectLoader/ConnectLoader.tsx
@@ -42,6 +42,7 @@ export interface ConnectLoaderParams {
   checkout: Checkout;
   allowedChains: ChainId[];
   isCheckNetworkEnabled?: boolean;
+  blocklistWalletRdns: string[];
 }
 
 export function ConnectLoader({
@@ -60,6 +61,7 @@ export function ConnectLoader({
     allowedChains,
     browserProvider,
     isCheckNetworkEnabled,
+    blocklistWalletRdns,
   } = params;
 
   const { t } = useTranslation();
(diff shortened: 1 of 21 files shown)


packages/checkout/widgets-lib/src/components/ConnectLoader/ConnectLoader.tsx
```diff
@@ -37,11 +37,12 @@ export interface ConnectLoaderProps {
 
 export interface ConnectLoaderParams {
   targetChainId: ChainId;
-  walletProviderName?: WalletProviderName;
-  browserProvider?: WrappedBrowserProvider;
+  walletProviderName: WalletProviderName | undefined;
```
keithbro-imx, line 40: IMO preferable in this case because it makes it explicit that we are not passing a value through (for example in packages/checkout/widgets-lib/src/widgets/immutable-commerce/functions/getConnectLoaderParams.ts)

packages/checkout/widgets-lib/src/widgets/connect/ConnectWidget.tsx
```diff
@@ -219,7 +220,8 @@ export default function ConnectWidget({
               targetWalletRdns={targetWalletRdns}
               targetChainId={targetChain}
               allowedChains={allowedChains ?? [targetChain]}
-              blocklistWalletRdns={blocklistWalletRdns}
+              allowlistWalletRdns={allowlistWalletRdns}
+              blocklistWalletRdns={blocklistWalletRdns ?? []}
```
keithbro-imx, line 224: Default to [] here so that we reduce complexity by not having to deal with undefined lower in the stack

packages/checkout/widgets-lib/src/widgets/sale/SaleWidgetRoot.tsx
```diff
@@ -124,6 +124,8 @@ export class Sale extends Base<WidgetType.SALE> {
       browserProvider: this.browserProvider,
       checkout: this.checkout,
       allowedChains: [this.checkout.config.l2ChainId],
+      allowlistWalletRdns: undefined,
+      walletProviderName: undefined,
```
keithbro-imx, line 128: Interesting that we're not passing along `this.parameters.walletProviderName,` like in the other widget roots, not sure what the implication of that is...

PR #2723, opened 2025-09-26
feat(checkout-widgets): Allow hide header on On-Ramp widget

priyaleedman-imm:
Jira: https://immutable.atlassian.net/browse/GFI-1196

**Default**
Screenshot 2025-09-26 at 3 45 12 pm

**Removed**
Screenshot 2025-09-26 at 3 46 42 pm


packages/checkout/widgets-lib/src/widgets/on-ramp/views/OnRampMain.tsx
```diff
@@ -55,6 +56,7 @@ export function OnRampMain({
   showMenu,
   customTitle,
   customSubTitle,
+  showHeader,
```
keithbro-imx, line 59: Would be a bit better to do the defaulting here but not a blocker

keithbro-imx (approved):

PR #2736, opened 2025-10-28
chore(ci): use github secrets for nx access token

mw-w:
# Summary
Remove nx access token and use access token in github secrets, environment variable supersedes nx.json

## Added 
- NX Access Token secret variable in workflows that use NX

## Removed
- NX Access Token from nx.json

## Security
- Rotate NX access token after merge


timm088 (approved):
:+1: Good with this, matches NX best practices.

timm088 (approved):

PR #2837, opened 2026-04-08
feat(audience): interactive demo page and CDN bundle

ImmutableJeffrey:
## What this PR does

Adds two things needed to ship `@imtbl/audience` for studios that embed it with a `<script>` tag:

1. **A single-file build of the SDK** at `dist/cdn/imtbl-audience.global.js`. Studios drop one `<script>` tag into their HTML, then use `ImmutableAudience.Audience.init({...})` to start tracking. No bundler, no `npm install`.
2. **An interactive demo page** that loads that single-file build and exercises every public method on the `Audience` class against the real dev / sandbox backend. Live event log, one button per method, test keys baked in. Fastest way to sanity-check SDK changes end-to-end.

## How to try it

```sh
cd packages/audience/sdk-sample-app
pnpm dev
```

Open http://localhost:3456/. Test publishable keys and a 10-step "what to try" walkthrough are in the sample-app's README.

## Addressing the review feedback

@nattb8 asked two things in their review:

> Shouldn't the demo live outside the sdk? Copy passport - audience/sdk-sample-app or audience/demo. Also, this is included in the CDN bundle too. Don't think we want that.

**1. Demo moved out of the sdk package.** It now lives in its own private workspace package at `packages/audience/sdk-sample-app`, matching the `passport-sdk-sample-app` / `checkout-sdk-sample-app` / `dex-sdk-sample-app` convention. The `sdk` package is now just the shippable library — no demo files mixed in with the build artifacts.

**2. Demo was never bundled into the single-file SDK.** Verified this before doing the restructure: `src/cdn.ts` (the entry point for the single-file build) imports only `./sdk`, `./config`, and `@imtbl/audience-core` — nothing in the TypeScript source references the demo's HTML/JS/CSS, so the minified output doesn't contain any of it. And the sdk's `"files": ["dist"]` field already excluded the demo directory from the published npm tarball. Confirmed by running `pnpm pack` on the sdk package and listing the archive contents — it has only `dist/{browser,cdn,node,types}`, `LICENSE.md`, `README.md`, and `package.json`. No demo files, no sample-app files. The restructure in point 1 removes any ambiguity at the directory level regardless.

## What changed at a glance

```
packages/audience/sdk-sample-app/       NEW — the interactive demo, private workspace package
├── README.md                             how to run, test keys, walkthrough
├── index.html, demo.js, demo.css         the demo page itself
├── serve.mjs                             ~90-line Node static file server, no deps
└── package.json                          depends on @imtbl/audience via workspace link

packages/audience/sdk/
├── src/cdn.ts, cdn.test.ts             NEW — single-file SDK entry point + smoke test
├── tsup.cdn.js                         NEW — build config for the single-file build
├── README.md                           NEW — package docs (install, quickstart, API, errors)
└── package.json                          adds the single-file build step to `pnpm build`

pnpm-workspace.yaml                       registers the new sample-app package
pnpm-lock.yaml                            picks up the new workspace link (6 lines)
```

The branch is rebased onto current main, so it sits on top of #2838 (which introduced the `@imtbl/audience` publishing pipeline). The two build systems coexist cleanly: the existing `tsup.config.js` builds `dist/browser` + `dist/node` and the new `tsup.cdn.js` builds `dist/cdn`, without either touching the other's output.

Ten commits total — follow them in order if you want to see how the demo came together incrementally.

## Related tickets

- Closes **SDK-49** (Web SDK: CDN bundle, demo page, README)

## Test plan

- [x] `pnpm --filter @imtbl/audience-core --filter @imtbl/audience run lint` — clean on both packages
- [x] Same filter + `run typecheck` — clean
- [x] Same filter + `run test` — 113 core tests + 51 sdk tests all pass
- [x] `pnpm --filter @imtbl/audience run build` — produces all expected artifacts: browser ESM, node CJS+ESM, single-file web build (52 KB), and the rolled-up `.d.ts`
- [x] `pnpm --filter @imtbl/audience-sdk-sample-app run dev` — demo loads at http://localhost:3456/, every method exercisable, 200 responses from the sandbox audience API
- [x] `pnpm pack` on `packages/audience/sdk` — tarball is clean: only shipping artifacts, no demo files or sample-app files
- [ ] CI green on this branch


nattb8 (commented):
Shouldn't the demo live outside the sdk? Copy passport - audience/sdk-sample-app or audience/demo.

Also, this is included in the CDN bundle too. Don't think we want that.

packages/audience/sdk/src/cdn.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    (globalThis as unknown as { ImmutableAudience?: unknown }).ImmutableAudience = sentinel;
+
+    const warn = jest.spyOn(console, 'warn').mockImplementation(() => {});
+    await import('./cdn');
+
+    expect((globalThis as unknown as { ImmutableAudience?: unknown }).ImmutableAudience).toBe(sentinel);
+    expect(warn).toHaveBeenCalledWith(expect.stringContaining('loaded twice'));
+    warn.mockRestore();
+  });
+});
```
nattb8, line 56: The `as any` cast is needed because the inline type shapes `AudienceError` as `typeof Error`. If you type it as `typeof AudienceError` instead, you can construct it without the cast and the compiler will catch a broken constructor signature.

packages/audience/sdk/src/cdn.test.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  it('attaches the SDK surface to window.ImmutableAudience', async () => {
+    await import('./cdn');
+    const g = (globalThis as unknown as {
+      ImmutableAudience?: {
+        Audience: { init: Function };
+        AudienceError: typeof Error;
+        IdentityType: Record<string, string>;
+        version: string;
+      };
```
nattb8: Using `typeof Audience` (imported from `./sdk`) here would turn this into a compile-time check that the CDN export shape matches the real class.

nattb8 (commented):
Can we split sample app and CDN into separate PRs please?

PR #2838, opened 2026-04-09
build(audience): make @imtbl/audience publishable on npm

ImmutableJeffrey:
## Summary

Makes `@imtbl/audience` installable via `npm install @imtbl/audience` from the npm registry and includes it in the monorepo's publish workflow. Unblocks SDK-66 (Publish WebSDK on npm) and SDK-63 (integrate Web SDK into Play).

## The two problems this PR solves

**1. The package wasn't installable outside the monorepo.** The built output referenced `@imtbl/audience-core` as a dependency, but audience-core is a private internal package that's never published to npm. Any consumer running `npm install @imtbl/audience` would hit a missing dependency error.

**2. The publish workflow wasn't including audience at all.** `.github/workflows/publish.yaml` runs `pnpm -r pack:root` to collect packages for publishing. Every other shipped `@imtbl/*` package defines a `pack:root` script. `packages/audience/sdk` did not, so the workflow silently skipped it.

## What the commits do

- **`tsup.config.js`** — local build config that bundles all `@imtbl/*` workspace dependencies (audience-core + transitive metrics) directly into the output. This makes the package fully self-contained — no external `@imtbl/*` packages needed at runtime.
- **`rollup.dts.config.js`** + `rollup-plugin-dts` — bundles TypeScript type declarations into a single self-contained file. Without this, consumers would get unresolved type references to `@imtbl/audience-core`. Leftover per-file `.d.ts` files are cleaned up after bundling.
- **`scripts/prepack.mjs` / `scripts/postpack.mjs`** — removes `@imtbl/audience-core` from `package.json` before packaging (since it's bundled inline and doesn't exist on npm), then restores it afterwards for local development. This pattern is unique to `@imtbl/audience` because it's the only published package that depends on a private workspace package.
- **`package.json` — `pack:root` script** — registers audience with the publish workflow, matching every other shipped `@imtbl/*` package.
- **`package.json` — explicit devDependencies** — `esbuild-plugin-replace` and `esbuild-plugins-node-modules-polyfill` declared explicitly rather than relying on root hoisting (matches `@imtbl/checkout-sdk` pattern).
- **`.eslintignore`** — exempts the new build configs from eslint.
- **`.gitignore`** — adds `*.prepack-backup` to prevent accidental commit if packaging fails mid-way.

## Verified locally

- Built output contains no references to `@imtbl/*` packages — fully self-contained
- Type declarations contain no references to `@imtbl/*` packages
- Package installs cleanly in a fresh project via `npm install ./imtbl-audience-0.0.0.tgz` — both runtime and types resolve correctly
- `pnpm --filter @imtbl/audience pack:root` produces the package at the repo root on the first try

## What this PR still does NOT do (flagged for verification before first publish)

**Verify nx release assigns the correct version to @imtbl/audience.** The publish workflow seeds versions from `@imtbl/metrics` for packages in `@imtbl/sdk`'s dependency tree. `@imtbl/audience` is not in that tree, so it starts at `0.0.0`. nx release *should* bump it to match the rest (since `projectsRelationship` is `fixed`) — but **this hasn't been independently verified**.

**Recommended first publish flow:**
1. Merge this PR
2. Trigger `publish.yaml` via `workflow_dispatch` with `dry_run: true` and `release_type: prerelease`
3. Inspect the logs: confirm `@imtbl/audience` is versioned correctly and included in the package list
4. If dry run looks good, re-run with `dry_run: false`
5. Verify `npm install @imtbl/audience@<version>` works from a fresh project

## Tickets

- **SDK-66** (Publish WebSDK on npm) — this PR is the last code change needed
- **SDK-63** (Integrate Web SDK into Play/Game Page) — Game Page needs `npm install @imtbl/audience`; this PR makes that possible

## Test plan

- [x] `pnpm --filter @imtbl/audience build` — ESM + CJS + browser + bundled types all produced
- [x] `pnpm --filter @imtbl/audience pack:root` — package produced at repo root
- [x] Clean install in fresh `/tmp` project — runtime and types resolve
- [x] No bare `@imtbl/*` references in built output
- [ ] CI green
- [ ] `workflow_dispatch --dry_run` to verify versioning and inclusion in publish

## Relationship to other open PRs

Independent of #2836 (foundation: errors + IdentityType) and #2837 (demo + CDN bundle). No file collisions; only adjacent-line overlap in `packages/audience/sdk/package.json`. Git auto-merges; whichever lands second regenerates `pnpm-lock.yaml`.

### DIFF
--- a/packages/audience/sdk/tsup.config.js
+++ b/packages/audience/sdk/tsup.config.js
@@ -1,0 +1,74 @@
+// @ts-check
+// Local tsup config for @imtbl/audience.
+//
+// Extends the monorepo's root tsup config but overrides `noExternal` so that
+// every `@imtbl/*` workspace dep (audience-core, its transitive metrics dep)
+// is inlined into the built bundle. This makes the package installable as a
+// standalone tarball without the pnpm `workspace:*` protocol.
+import { defineConfig } from 'tsup';
+import { nodeModulesPolyfillPlugin } from 'esbuild-plugins-node-modules-polyfill';
+import { replace } from 'esbuild-plugin-replace';
+import pkg from './package.json' assert { type: 'json' };
+
+const IMTBL_WORKSPACE = /^@imtbl\//;
+
+export default defineConfig((options) => {
+  if (options.watch) {
+    return {
+      entry: ['src/index.ts'],
+      outDir: 'dist/browser',
+      format: 'esm',
+      target: 'es2022',
+      platform: 'browser',
+      bundle: true,
+      noExternal: [IMTBL_WORKSPACE],
+      esbuildPlugins: [
+        nodeModulesPolyfillPlugin({
+          globals: { Buffer: true, process: true },
+          modules: ['crypto', 'buffer', 'process'],
+        }),
+        replace({
+          __SDK_VERSION__: pkg.version === '0.0.0' ? '2.0.0' : pkg.version,
+        }),
+      ],
+    };
+  }
+
+  return [
+    // Browser ESM bundle
+    {
+      entry: ['src/index.ts'],
+      outDir: 'dist/browser',
+      platform: 'browser',
+      format: 'esm',
+      target: 'es2022',
+      minify: true,
+      bundle: true,
+      noExternal: [IMTBL_WORKSPACE, '@uniswap/swap-router-contracts'],
+      treeshake: true,
+      esbuildPlugins: [
+        nodeModulesPolyfillPlugin({
+          globals: { Buffer: true, process: true },
+          modules: ['crypto', 'buffer', 'process'],
+        }),
+        replace({ __SDK_VERSION__: pkg.version }),
+      ],
+    },
+
+    // Node CJS + ESM bundle
+    {
+      entry: ['src/index.ts'],
+      outDir: 'dist/node',
+      platform: 'node',
+      format: ['esm', 'cjs'],
+      target: 'es2022',
+      minify: true,
+      bundle: true,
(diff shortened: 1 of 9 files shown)


packages/audience/sdk/tsup.config.js
```diff
@@ -0,0 +1,74 @@
...
+// Extends the monorepo's root tsup config but overrides `noExternal` so that
+// every `@imtbl/*` workspace dep (audience-core, its transitive metrics dep)
+// is inlined into the built bundle. This makes the package installable as a
+// standalone tarball without the pnpm `workspace:*` protocol.
+import { defineConfig } from 'tsup';
+import { nodeModulesPolyfillPlugin } from 'esbuild-plugins-node-modules-polyfill';
+import { replace } from 'esbuild-plugin-replace';
+import pkg from './package.json' with { type: 'json' };
+
+const IMTBL_WORKSPACE = /^@imtbl\//;
```
nattb8: These two lists are supposed to describe the same set of packages but they don't match. `tsup.config.js` uses a broad regex that covers the entire `@imtbl` scope, while `prepack.mjs` explicitly names just `audience-core` and `metrics`. Right now `@imtbl/pixel` isn't a dependency of the SDK so it won't be bundled, but if it ever gets added accidentally (including via a transitive dep), tsup will pull it in silently while prepack won't strip it from `package.json`, and `npm install` will break. Worth making `tsup.config.js` use the same explicit list as prepack.

nattb8 (approved):

rodrigo-fournier-immutable (approved):

PR #2841, opened 2026-04-09
feat(audience-core): unified AudienceError surface + fix partial-success silent drop

ImmutableJeffrey:
## Summary

Adds a unified `AudienceError` surface in `@imtbl/audience-core` so every audience surface (web, pixel, future Unity / Unreal) reports failures through the same type. Also fixes a silent data-loss bug in `MessageQueue.flush` where backend partial-rejections were silently dropped.

### Part A — Error surface

- New `AudienceError` class in `core/src/errors.ts` — extends Error with `code: AudienceErrorCode`, `status`, `endpoint`, `responseBody`, `cause`.
- `AudienceErrorCode` closed union: `'FLUSH_FAILED' | 'CONSENT_SYNC_FAILED' | 'NETWORK_ERROR' | 'VALIDATION_REJECTED'`.
- `toAudienceError(err, source, count?)` helper centralises the mapping from `TransportError` to `AudienceError`, removing the need for per-consumer mapping logic. Kept internal — not re-exported from the public index since consumer code never holds a `TransportError`.
- `invokeOnError(onError, err)` helper centralises the swallow-and-continue semantics shared by `MessageQueue` and `createConsentManager` so the internal state machines can't wedge on a throwing handler.
- `MessageQueue` accepts `onError?: (err: AudienceError) => void` in options. Fires after failed flushes. Callback exceptions are swallowed via `invokeOnError`.
- `createConsentManager` accepts `onError?` as a new optional last parameter. Fires on consent PUT failure via `.then()` on the fire-and-forget chain.

### Part B — Partial-success bug fix

- `httpSend` now parses the 2xx response body. If it reports `rejected > 0`, returns `ok: false` with the parsed body — instead of pretending everything succeeded.
- The partial-rejection check uses a `'rejected' in body` type guard before casting, so malformed backend responses shaped as arrays can't slip through (`typeof [] === 'object'` would otherwise let them pass).
- `toAudienceError` maps 2xx failures to `'VALIDATION_REJECTED'` with message `"Backend rejected M of N messages"`.
- `MessageQueue.flush` drops the batch on `VALIDATION_REJECTED` (retrying won't fix what the backend deterministically rejected) and fires `onError`. Generic `FLUSH_FAILED` / `NETWORK_ERROR` still retain messages for retry.

## Breaking changes

`@imtbl/audience-core` is `private: true` so these are internal-only, but any code in `packages/audience/sdk` or `packages/pixel` that still imported the following from `@imtbl/audience-core` will need to update:

**Removed exports (dead code — nothing in the repo consumed them):**

- `ConsentUpdatePayload` (type) — moved fully internal; only `httpSend` constructs these.
- `getAnonymousId` — superseded by `getOrCreateAnonymousId`.
- `setCookie` — only the read/delete side is part of the public API now.
- `storage` namespace — internal-only persistence layer.
- `SESSION_MAX_AGE` — replaced by `SESSION_START` / `SESSION_END`.
- `truncateSource` — the single caller was inlined.
- `getOrCreateSessionId`, `getSessionId` — superseded by `getOrCreateSession` which returns the full session result.
- `clearAttribution` — attribution is now immutable per session; no clear path needed.

**Changed exports:**

- `TransportError` is now a runtime export (class), previously a type-only export. Instances have `status`, `endpoint`, `body`, and `cause` fields. This matches the squash-merged SDK-84 shape.
- `toAudienceError` is still used internally by queue + consent but is **not** re-exported from the public index — consumer code receives an already-mapped `AudienceError` from its `onError` callback and never needs the raw `TransportError` → `AudienceError` mapper.

**New exports:**

- `AudienceError` (class), `AudienceErrorCode` (type).

## Linear

- **SDK-85** — Audience SDK: unified AudienceError surface in core + fix partial-success silent drop

## Test plan

- [x] `pnpm --filter @imtbl/audience-core --filter @imtbl/audience --filter @imtbl/pixel test` — 113 core + 49 sdk + 64 pixel
- [x] `pnpm --filter @imtbl/audience-core --filter @imtbl/audience --filter @imtbl/pixel typecheck` — clean
- [x] `pnpm --filter @imtbl/audience-core --filter @imtbl/audience --filter @imtbl/pixel lint` — clean
- [ ] CI green

## What's next

One follow-up PR wires the new `onError` surface through `@imtbl/audience` and `@imtbl/pixel` (SDK-88).

### DIFF
--- a/packages/audience/core/src/index.ts
+++ b/packages/audience/core/src/index.ts
@@ -39,7 +39,7 @@ export { generateId, getTimestamp, isBrowser } from './utils';
 export type { HttpSend, TransportOptions } from './transport';
 export { httpSend } from './transport';
 export type { TransportResult, AudienceErrorCode } from './errors';
-export { TransportError, AudienceError, toAudienceError } from './errors';
+export { TransportError, AudienceError } from './errors';
 export { MessageQueue } from './queue';
 export { collectContext } from './context';
 export { isTimestampValid, isAliasValid, truncate } from './validation';
--- a/packages/audience/core/src/index.ts
+++ b/packages/audience/core/src/index.ts
@@ -43,7 +43,12 @@ export { generateId, getTimestamp, isBrowser } from './utils';
 
 export type { HttpSend, TransportOptions } from './transport';
 export { httpSend } from './transport';
-export type { TransportError, TransportResult } from './errors';
+export type {
+  TransportError,
+  TransportResult,
+  AudienceErrorCode,
+} from './errors';
+export { AudienceError, toAudienceError } from './errors';
 export { MessageQueue } from './queue';
 export { collectContext } from './context';
 export {
--- a/packages/audience/core/src/index.ts
+++ b/packages/audience/core/src/index.ts
@@ -13,18 +13,14 @@ export type {
   BatchPayload,
   ConsentLevel,
   ConsentStatus,
-  ConsentUpdatePayload,
 } from './types';
 export { IdentityType } from './types';
 
 export {
   getOrCreateAnonymousId,
-  getAnonymousId,
   getCookie,
-  setCookie,
   deleteCookie,
 } from './cookie';
-export * as storage from './storage';
 
 export {
   getBaseUrl,
@@ -34,7 +30,6 @@ export {
   FLUSH_SIZE,
   COOKIE_NAME,
   SESSION_COOKIE,
-  SESSION_MAX_AGE,
   SESSION_START,
   SESSION_END,
 } from './config';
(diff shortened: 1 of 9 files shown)


packages/audience/core/src/index.ts
```diff
@@ -13,18 +13,14 @@ export type {
   BatchPayload,
   ConsentLevel,
   ConsentStatus,
-  ConsentUpdatePayload,
 } from './types';
 export { IdentityType } from './types';
 
 export {
   getOrCreateAnonymousId,
-  getAnonymousId,
```
nattb8: Should mention in the PR that these are breaking changes? so many exports removed

nattb8 (approved):

PR #2842, opened 2026-04-09
feat(audience): automatic CMP consent detection (GCM v2 + IAB TCF)

bkbooth:
## Summary

- Add `consentMode: 'auto'` init option to the pixel for automatic consent detection from existing CMPs on the page
- Implement Google Consent Mode v2 detection (reads `window.dataLayer`, intercepts `dataLayer.push` for real-time updates)
- Implement IAB TCF v2 detection (reads `window.__tcfapi`, registers `addEventListener` for consent changes)
- CMP detection module lives in `@imtbl/audience-core` so the web SDK can reuse it
- Polls up to 3 times over ~2.4s for async CMP loading, then listens for mid-session consent changes
- Bundle size: **4.77 KB gzipped** (well under 10KB budget)

### Consent mapping

| CMP signal | Pixel consent level |
|---|---|
| `analytics_storage: 'denied'` / no Purpose 1 | `none` |
| `analytics_storage: 'granted'` + `ad_storage: 'denied'` / Purpose 1 only | `anonymous` |
| `analytics_storage: 'granted'` + `ad_storage: 'granted'` / Purpose 1 + 3/4/5 | `full` |

### De-scoped

Direct CMP integrations (OneTrust, Cookiebot) moved out of scope — GCM + TCF cover the vast majority of deployments since both OneTrust and Cookiebot fire Google Consent Mode updates. Context preserved on [SDK-87](https://linear.app/imtbl/issue/SDK-87) for a future issue.

## Test plan

- [x] 29 CMP unit tests (GCM mapping, TCF mapping, priority chain, polling, cleanup, consent change callbacks)
- [x] 8 pixel integration tests for `consentMode: 'auto'` (deferred page view, CMP upgrade/downgrade, teardown)
- [x] All 192 tests passing (119 core + 73 pixel)
- [x] Bundle size verified: 4.77 KB gzipped (max 10 KB)

Refs: SDK-87

### DIFF
--- a/packages/audience/core/src/cmp.ts
+++ b/packages/audience/core/src/cmp.ts
@@ -168,6 +168,7 @@ function mapTcfConsent(data: TcfData): ConsentLevel {
  */
 function detectTcf(onUpdate: ConsentCallback): CmpDetector | null {
   const win = window as unknown as Record<string, unknown>;
+  // eslint-disable-next-line no-underscore-dangle
   const tcfapi = win.__tcfapi as TcfApi | undefined;
   if (typeof tcfapi !== 'function') return null;
 
--- a/packages/audience/core/src/cmp.ts
+++ b/packages/audience/core/src/cmp.ts
@@ -232,11 +232,15 @@ export function detectCmp(onUpdate: ConsentCallback): CmpDetector | null {
  * This function tries detection immediately, then polls up to MAX_POLLS times.
  * Once a CMP is found, polling stops and the callback is invoked for future changes.
  *
+ * If no CMP is found after all polling attempts, `onTimeout` is called so callers
+ * can distinguish "CMP said none" from "no CMP found at all".
+ *
  * Returns a cleanup function that stops polling and tears down CMP listeners.
  */
 export function startCmpDetection(
   onUpdate: ConsentCallback,
   onDetected: (detector: CmpDetector) => void,
+  onTimeout?: () => void,
 ): () => void {
   // Try immediately
   let detector = detectCmp(onUpdate);
@@ -259,6 +263,7 @@ export function startCmpDetection(
 
     if (pollCount >= MAX_POLLS) {
       clearInterval(timer);
+      onTimeout?.();
     }
   }, POLL_INTERVAL);
 
--- a/packages/audience/core/src/cmp.ts
+++ b/packages/audience/core/src/cmp.ts
@@ -1,0 +1,268 @@
+/**
+ * CMP (Consent Management Platform) auto-detection.
+ *
+ * Detects consent state from:
+ *   1. Google Consent Mode v2 (via `window.dataLayer`)
+ *   2. IAB TCF v2 (via `window.__tcfapi`)
+ *
+ * Priority: GCM → TCF → fall back to 'none'.
+ * Once a CMP is detected, registers a listener for ongoing consent changes.
+ */
+import type { ConsentLevel } from './types';
+
+// -- Types for external CMP globals ------------------------------------------
+
+interface GcmConsentParams {
+  analytics_storage?: 'granted' | 'denied';
+  ad_storage?: 'granted' | 'denied';
+  [key: string]: unknown;
+}
+
+interface TcfData {
+  gdprApplies?: boolean;
+  purpose?: { consents?: Record<number, boolean> };
+  listenerId?: number;
+  eventStatus?: string;
+}
+
+type TcfCallback = (data: TcfData, success: boolean) => void;
+
+type TcfApi = (
+  command: string,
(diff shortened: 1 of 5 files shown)


packages/audience/core/src/cmp.ts
```diff
@@ -0,0 +1,269 @@
...
+    detector = detectCmp(onUpdate);
+
+    if (detector) {
+      clearInterval(timer);
+      onDetected(detector);
+      return;
+    }
+
+    if (pollCount >= MAX_POLLS) {
+      clearInterval(timer);
```
nattb8, line 265: If nothing is detected after 3 tries, customers have to explicitly call setConsent() right? How do they know if it failed to detect? Wondering if this is making it easier but also complicated

Maybe we just need a onFailed callback so we know the difference between "detected `none` from CMP" and "can't find CMP"

packages/audience/pixel/src/pixel.ts
```diff
@@ -113,6 +128,12 @@ export class Pixel {
...
     this.queue.start();
 
+    // If consentMode is 'auto', start CMP detection.
+    // The pixel begins at 'none' and upgrades when a CMP is found.
+    if (isAutoConsent) {
+      this.startCmpDetection();
+    }
+
     // Auto-fire page view if consent allows
     if (this.consent.level !== 'none') {
```
nattb8: 🔴 **Duplicate page view when CMP is already on the page**

When a CMP is available immediately (e.g. GTM has already set `dataLayer` consent defaults before the pixel loads — which is the common case), `startCmpDetection()` calls `onDetected` **synchronously**. That upgrades consent and fires `this.page()` inside `startCmpDetection()`. Then `init()` falls through to this check, consent is no longer `none`, and a second page view fires.

This won't be caught by the current tests because `mockStartCmpDetection` captures the callbacks without calling them during `init()`.

Fix — make the two paths mutually exclusive:
```ts
if (isAutoConsent) {
  this.startCmpDetection();
} else if (this.consent.level !== 'none') {
  this.page();
}
```

packages/audience/pixel/src/pixel.ts
```diff
@@ -189,6 +214,31 @@ export class Pixel {
...
+    const onCmpUpdate = (level: ConsentLevel): void => {
+      if (!this.isReady()) return;
+      this.consent!.setLevel(level);
+
+      // If this is the first consent upgrade from 'none', fire the initial
+      // page view that was deferred at init.
+      if (level !== 'none' && !this.sessionId) {
+        this.page();
+      }
+    };
```
nattb8, line 233: 🟡 **`!this.sessionId` is a fragile guard for the deferred page view**

Using `!this.sessionId` as a proxy for "initial page view hasn't fired yet" has two problems.

First, it's indirect — if `sessionId` is set through any other path before the CMP callback fires, the deferred page view is silently suppressed.

Second, and more importantly: `setConsent()` is public, but it only calls `this.consent.setLevel()` — it doesn't go through `onCmpUpdate`, so the deferred page view logic never runs. If CMP detection fails and the caller manually calls `setConsent('anonymous')` as a fallback, consent upgrades and events start flowing, but the initial page view is permanently lost.

A dedicated flag would make the intent explicit and cover both paths:
```ts
private initialPageViewFired = false;

// in onCmpUpdate:
if (level !== 'none' && !this.initialPageViewFired) {
  this.initialPageViewFired = true;
  this.page();
}

// in setConsent():
setConsent(level: ConsentLevel): void {
  if (!this.isReady()) return;
  this.consent!.setLevel(level);
  if (level !== 'none' && !this.initialPageViewFired) {
    this.initialPageViewFired = true;
    this.page();
  }
}
```

packages/audience/pixel/src/pixel.ts
```diff
@@ -38,6 +40,8 @@ export interface PixelInitOptions {
   key: string;
   environment?: Environment;
   consent?: ConsentLevel;
+  /** Set to 'auto' to auto-detect consent from CMPs (Google Consent Mode, IAB TCF). */
+  consentMode?: 'auto' | ConsentLevel;
```
nattb8: 🟡 **`consentMode` mixes a mode selector with a static value**

`'auto' | ConsentLevel` means `consentMode: 'full'` is valid, but so is `consent: 'full'`. Two fields that accept the same values with one silently winning is confusing for callers, and will only get worse if this pattern lands in the web SDK too.

The useful new behaviour here is just `'auto'`. Consider narrowing the type to `consentMode?: 'auto'` and keeping `consent` as the only way to set a static level — or deprecating `consent` explicitly if the intent is to replace it.

nattb8 (approved):

PR #2848, opened 2026-04-11
refactor(audience): unify consent level checks via capability-query functions

ImmutableJeffrey:
## Summary

- Add two helper functions to `@imtbl/audience-core`:
  - `canTrack(level)` — true if the level allows event recording
  - `canIdentify(level)` — true if the level allows identity features (identify, alias, userId, email hashing)
- Export both from `@imtbl/audience-core` and `@imtbl/audience`
- Replace `=== 'none'` / `=== 'full'` comparisons in `@imtbl/audience` (`sdk.ts`) and `@imtbl/pixel` (`pixel.ts`, `autocapture.ts`) with calls to these helpers
- Retype `SnippetOptions.consent` from `'none' | 'anonymous' | 'full'` to `ConsentLevel`
- Rename Pixel's private `canTrack()` method to `isTrackingAllowed()`

LInear: [SDK-122](https://linear.app/imtbl/issue/SDK-122/audience-sdk-unify-consent-level-checks)

## Tests

- core: 150 tests pass (6 new)
- sdk: 49 tests pass
- pixel: 78 tests pass
- `pnpm typecheck` and `pnpm lint` clean across all three packages

### DIFF
--- a/packages/audience/pixel/src/pixel.ts
+++ b/packages/audience/pixel/src/pixel.ts
@@ -22,6 +22,7 @@ import {
   collectAttribution,
   getOrCreateSession,
   createConsentManager,
+  canTrack as consentAllowsTracking,
   startCmpDetection,
 } from '@imtbl/audience-core';
 import { setupAutocapture } from './autocapture';
@@ -124,7 +125,7 @@ export class Pixel {
     if (isAutoConsent) {
       // CMP detection will fire the deferred page view when consent upgrades.
       this.startCmpDetection();
-    } else if (this.consent.level !== 'none') {
+    } else if (consentAllowsTracking(this.consent.level)) {
       // Static consent — fire page view immediately.
       this.initialPageViewFired = true;
       this.page();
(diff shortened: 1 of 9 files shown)


packages/audience/sdk/src/sdk.ts
```diff
@@ -144,12 +146,12 @@ export class Audience {
 
   /** True when consent is 'none' — SDK should not enqueue anything. */
```
nattb8: Update this comment, as it delegates the check to `canTrack()`
```suggestion
  /** True when the current consent level does not permit tracking. */
```

packages/audience/pixel/src/pixel.ts
```diff
@@ -22,6 +22,7 @@ import {
   collectAttribution,
   getOrCreateSession,
   createConsentManager,
+  canTrack as consentAllowsTracking,
```
nattb8: Might be better to rename pixel's private `canTrack()` to something else e.g. `isTrackingAllowed()`, then you don't need to rename this and can be consistent everywhere.

nattb8 (approved):

PR #2850, opened 2026-04-13
feat(audience): typed track() surface for v1 event catalogue (SDK-116)

ImmutableJeffrey:
# Summary

Adds pre-defined events per the [Web SDK Event Reference (v1)](https://immutable.atlassian.net/wiki/spaces/~84417537/pages/3862560801/Web+SDK+Event+Reference+v1). 

Linear: [SDK-116](https://linear.app/imtbl/issue/SDK-116/pre-defined-events-v1).

## Added

- Rewrites `sdk.track()` as a generic `track<E extends string>` with a `PropsFor<E>` conditional type mapping each of 11 event names to its property interface
- Exports `AudienceEvents`, 11 `*Properties` interfaces, `ProgressionStatus`, `ResourceFlow`
- Adds `sdk.test-d.ts` with compile-time negative assertions
- Adds runtime regression tests for 5 of the new events

## Demo

https://github.com/user-attachments/assets/201ee92e-c23d-447c-989b-e68589c2a05a

## Test plan

- [x] `pnpm lint && pnpm typecheck && pnpm test && pnpm typegen` — 54/54 passing

### DIFF
--- a/packages/audience/sdk/src/events.ts
+++ b/packages/audience/sdk/src/events.ts
@@ -133,6 +133,8 @@ interface EventPropsMap {
   link_clicked: LinkClickedProperties;
 }
 
+export type AudienceEventName = keyof EventPropsMap;
+
 /**
  * Event name → property type. Falls back to `Record<string, unknown>` for
  * unknown names. Used by `sdk.track()` to type-check property shapes at the
--- a/packages/audience/sdk/src/events.ts
+++ b/packages/audience/sdk/src/events.ts
@@ -1,0 +1,149 @@
+export const AudienceEvents = {
+  SIGN_UP: 'sign_up',
+  SIGN_IN: 'sign_in',
+  WISHLIST_ADD: 'wishlist_add',
+  WISHLIST_REMOVE: 'wishlist_remove',
+  PURCHASE: 'purchase',
+
+  GAME_LAUNCH: 'game_launch',
+  PROGRESSION: 'progression',
+  RESOURCE: 'resource',
+
+  EMAIL_ACQUIRED: 'email_acquired',
+  GAME_PAGE_VIEWED: 'game_page_viewed',
+  LINK_CLICKED: 'link_clicked',
+} as const;
+
+export interface SignUpProperties {
+  /** Optional. */
+  method?: string;
+}
+
+export interface SignInProperties {
+  /** Optional. */
+  method?: string;
+}
+
+export interface WishlistAddProperties {
+  /** Required. */
+  gameId: string;
+  /** Optional. */
+  source?: string;
+  /** Optional. */
+  platform?: string;
+}
+
+export interface WishlistRemoveProperties {
+  /** Required. */
+  gameId: string;
+}
+
+export interface PurchaseProperties {
+  /** Required. */
+  currency: string;
+  /** Required. */
+  value: number;
+  /** Optional. */
+  itemId?: string;
+  /** Optional. */
+  itemName?: string;
+  /** Optional. */
+  quantity?: number;
+  /** Optional. */
+  transactionId?: string;
+}
+
+export interface GameLaunchProperties {
+  /** Optional. */
+  platform?: string;
+  /** Optional. */
+  version?: string;
+  /** Optional. */
+  buildId?: string;
+}
+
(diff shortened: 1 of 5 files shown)


packages/audience/sdk/src/events.ts
```diff
@@ -0,0 +1,103 @@
...
+  purchase: PurchaseProperties;
+  game_launch: GameLaunchProperties;
+  progression: ProgressionProperties;
+  resource: ResourceProperties;
+  email_acquired: EmailAcquiredProperties;
+  game_page_viewed: GamePageViewedProperties;
+  link_clicked: LinkClickedProperties;
+}
+
+export type PropsFor<E extends string> =
```
bkbooth, line 150: This is some serious types kung fu, can you add a comment explainer of the intent here so it doesn't get unintentionally broken in the future?

packages/audience/sdk/src/sdk.ts
```diff
@@ -240,15 +241,21 @@ export class Audience {
...
     if (this.isTrackingDisabled()) return;
     getOrCreateSession(this.cookieDomain);
 
+    const [properties] = args;
     this.enqueue('track', {
       ...this.baseMessage(),
       type: 'track',
       eventName: truncate(event),
-      properties,
+      properties: properties as Record<string, unknown> | undefined,
```
bkbooth, line 255: Minor nitpick from Claude:

> `properties as Record<string, unknown> | undefined` — the cast is safe but could be avoided with a small refactor to enqueue's own typing. Minor, not blocking.

bkbooth (dismissed):

packages/audience/sdk/src/events.ts
```diff
@@ -0,0 +1,103 @@
...
+  version?: string;
+  buildId?: string;
+}
+
+export type ProgressionStatus = 'start' | 'complete' | 'fail';
+
+export interface ProgressionProperties {
+  status: ProgressionStatus;
+  world?: string;
+  level?: string;
```
nattb8, line 73: Need to update https://immutable.atlassian.net/wiki/spaces/~84417537/pages/3862560801/Web+SDK+Event+Reference+v1 that level is a string, not just a number
Could be '5' or '5-A'

packages/audience/sdk/src/sdk.ts
```diff
@@ -240,15 +241,21 @@ export class Audience {
    * Pass the event name and any properties you want to analyse later.
```
nattb8, line 236: Need to update this. Track also takes enforces typed property for predefined events too

packages/audience/sdk/src/index.ts
```diff
@@ -1,4 +1,20 @@
...
+  ProgressionProperties,
+  ProgressionStatus,
+  PurchaseProperties,
+  ResourceFlow,
+  ResourceProperties,
+  SignInProperties,
+  SignUpProperties,
+  WishlistAddProperties,
+  WishlistRemoveProperties,
+} from './events';
```
nattb8, line 21: > `PropsFor<E>` is exported from `events.ts` but is not re-exported from `index.ts`. Consumers who want to write a typed wrapper (e.g. `function logPurchase(props: PropsFor<'purchase'>) {... }`) cannot import it without reaching into the internal path. Worth adding to the public surface.

packages/audience/sdk/src/sdk.ts
```diff
@@ -240,15 +241,21 @@ export class Audience {
...
    * No-op when consent is 'none'.
    */
-  track(event: string, properties?: Record<string, unknown>): void {
+  track<E extends string>(
```
nattb8: From our design docs we want auto complete:

> The single-generic approach works correctly but does not give IDE autocomplete for known event names — typing `sdk.track('` will not surface `'purchase'`, `'progression'`, etc. as suggestions. The overload pattern from the reviewer's design (`track<E extends AudienceEvent>(event: E, ...): void; track(event: string & {}, ...): void;`) gives this for free. The `string & {}` trick on the fallback overload stops bare `string` from hiding the literal suggestions. Worth considering if developer experience is a priority for this SDK.

nattb8 (approved):

PR #2853, opened 2026-04-14
feat(audience): CDN bundle entry point and tsup build (SDK-115)

ImmutableJeffrey:
# Summary

Adds a single-file IIFE build of `@imtbl/audience` at `dist/cdn/imtbl-audience.global.js` so studios can embed via `<script>` and call `ImmutableAudience.Audience.init({...})` — no bundler, no `npm install` required.

Linear: [SDK-115](https://linear.app/imtbl/issue/SDK-115/cdn-bundle).

### DIFF
--- a/packages/audience/sdk/tsup.config.js
+++ b/packages/audience/sdk/tsup.config.js
@@ -29,7 +29,7 @@ export default defineConfig((options) => {
           modules: ['crypto', 'buffer', 'process'],
         }),
         replace({
-          __SDK_VERSION__: pkg.version === '0.0.0' ? '2.0.0' : pkg.version,
+          __SDK_VERSION__: pkg.version === '0.0.0' ? '0.0.0-local' : pkg.version,
         }),
       ],
     };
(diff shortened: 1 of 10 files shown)


packages/audience/sdk/tsup.config.js
```diff
@@ -29,7 +29,7 @@ export default defineConfig((options) => {
           modules: ['crypto', 'buffer', 'process'],
         }),
         replace({
-          __SDK_VERSION__: pkg.version === '0.0.0' ? '2.0.0' : pkg.version,
+          __SDK_VERSION__: pkg.version === '0.0.0' ? '0.0.0-local' : pkg.version,
```
nattb8, line 32: This is correct — hardcoding `'2.0.0'` as the local dev fallback was clearly wrong — but it changes the version string reported by all local builds of the existing npm package, not just the new CDN bundle. Worth a quick check that nothing in the codebase compares against `'2.0.0'` at runtime before the next publish.

nattb8 (approved):

JCSanPedro (approved):

PR #2854, opened 2026-04-15
feat(orderbook): Add createTraitBid, getTraitBid and listTraitBid functions for orderbook

JCSanPedro:

# Summary

Regenerates the orderbook api client, and adds `createTraitBid`, `getTraitBid` and `listTraitBids` utility functions for working with trait bids for the orderbook sdk. 

## Security


packages/orderbook/src/orderbook.ts
```diff
@@ -566,6 +601,48 @@ export class Orderbook {
...
+    buy,
+    orderStart,
+    orderExpiry,
+  }: PrepareTraitBidParams): Promise<PrepareTraitBidResponse> {
+    return this.seaport.prepareSeaportOrder(
+      makerAddress,
+      sell,
+      buy,
+      true,
+      orderStart || new Date(),
```
naveen-imtb, line 624: In what scenario will the startDate be `undefined`? 🤔  Do we need to default it? Same with `orderExpiry` - Are they not [mandatory fields](https://docs.immutable.com/api-reference/orders/create-a-trait-bid)?

JCSanPedro: Yes they are mandatory fields for the API. If the caller of this SDK function doesn't define `start` and `expiry` values, then we opt to use sensible defaults here.

naveen-imtb (approved):

PR #2864, opened 2026-04-16
feat(audience): sessionId on all events, UTMs on sign_up and link_clicked (SDK-140)

ImmutableJeffrey:
## Summary

Adds `sessionId` to all `page()` and `track()` event properties. Adds UTM attribution (parsed fresh from the current URL) to `sign_up` and `link_clicked` track events only.

Renames `collectAttribution` → `collectSessionAttribution` and introduces `collectPageAttribution` (no session cache). Extracts shared URL-parsing logic into a `buildAttribution` helper. Removes a dead empty-check guard in `page()`. Fixes `@imtbl/metrics` resolution in the SDK jest config.

Linear: [SDK-140](https://linear.app/imtbl/issue/SDK-140/attach-sessionid-to-all-events-utms-to-sign-up-and-link-clicked)

### DIFF
--- a/packages/audience/sdk/jest.config.ts
+++ b/packages/audience/sdk/jest.config.ts
@@ -9,6 +9,10 @@ const config: Config = {
   },
   moduleNameMapper: {
     '^@imtbl/audience-core$': '<rootDir>/../core/src/index.ts',
+    // The core source code imports other @imtbl packages (e.g. metrics).
+    // This tells jest where to find them so we can run tests without
+    // building the whole monorepo first.
+    '^@imtbl/(.*)$': '<rootDir>/../../../node_modules/@imtbl/$1/src',
   },
 };
 
--- a/packages/audience/sdk/src/sdk.ts
+++ b/packages/audience/sdk/src/sdk.ts
@@ -20,7 +20,8 @@ import {
   isTimestampValid,
   truncate,
   collectContext,
-  collectAttribution,
+  collectSessionAttribution,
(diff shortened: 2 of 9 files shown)


packages/audience/sdk/src/sdk.ts
```diff
@@ -44,6 +45,10 @@ import {
 export class Audience {
   private static liveInstances = 0;
 
+  private static readonly UTM_EVENTS: ReadonlySet<string> = new Set([
```
nattb8: Nit: doesn't have to live in the class, can just be top level constant.

packages/audience/sdk/jest.config.ts
```diff
@@ -9,6 +9,7 @@ const config: Config = {
   },
   moduleNameMapper: {
     '^@imtbl/audience-core$': '<rootDir>/../core/src/index.ts',
+    '^@imtbl/(.*)$': '<rootDir>/../../../node_modules/@imtbl/$1/src',
```
nattb8, line 15: Why did? Add a short comment?

packages/audience/sdk/src/sdk.ts
```diff
@@ -44,6 +45,10 @@ import {
...
+    'sign_up', 'link_clicked',
```
nattb8: Need to call out in JSdocs + public docs that if you use track() with these events, UTM is included?

nattb8 (approved):

PR #2881, opened 2026-06-09
feat(audience-sdk): add getAnonymousId() and imtbl_aid URL param support

nattb8:
## Summary

When a user clicks a social login option inside an in-app browser (e.g. Facebook), they get booted to the OS browser to complete OAuth — and the Audience SDK mints a fresh anonymous ID in that new context, orphaning all pre-login attribution from the in-app session. This PR fixes that by giving callers the tools to carry the anonymous ID across the browser switch.

- **`getAnonymousId()`** — exposes the current anonymous ID as a public method so integrators can append it to OAuth redirect URLs as `?imtbl_aid=<id>`
- **`?imtbl_aid=` on init** — the SDK reads this param on startup, adopts it as the canonical anonymous ID (writing it to the cookie), and strips it from the URL via `history.replaceState`

Mirrors the same pattern Segment uses (`?ajs_aid=`) and what the `play` repo already implements for Segment. No changes needed to the Pixel — it shares the same cookie and inherits the corrected ID automatically.

Closes [SDK-478](https://linear.app/imtbl/issue/SDK-478)

## Test plan

- [ ] New SDK unit tests cover: `imtbl_aid` adoption, URL stripping, `getAnonymousId()` return value matching outgoing event `anonymousId`, and that `imtbl_aid` is ignored at `none` consent
- [ ] All existing SDK tests pass (59 total)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/packages/audience/sdk/src/sdk.ts
+++ b/packages/audience/sdk/src/sdk.ts
@@ -12,6 +12,7 @@ import {
   MessageQueue,
   httpSend,
   getOrCreateAnonymousId,
+  adoptAnonymousId,
   getCookie,
   deleteCookie,
   generateId,
@@ -26,6 +27,7 @@ import {
   createConsentManager,
   canTrack,
   canIdentify,
+  isBrowser,
   SESSION_START,
   SESSION_END,
 } from '@imtbl/audience-core';
@@ -74,6 +76,18 @@ export class Audience {
 
   private destroyed = false;
 
+  private static readAndStripAidParam(): string | null {
+    if (!isBrowser()) return null;
+    const params = new URLSearchParams(window.location.search);
+    const aid = params.get('imtbl_aid');
+    if (!aid) return null;
+    params.delete('imtbl_aid');
+    const search = params.toString();
+    const newUrl = `${window.location.pathname}${search ? `?${search}` : ''}${window.location.hash}`;
+    window.history.replaceState(null, '', newUrl);
+    return aid;
+  }
+
   private constructor(config: AudienceConfig) {
     const {
       cookieDomain,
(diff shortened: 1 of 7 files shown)


packages/audience/sdk/src/sdk.ts
```diff
@@ -74,6 +76,18 @@ export class Audience {
...
 
+  private static readAndStripAidParam(): string | null {
+    if (!isBrowser()) return null;
+    const params = new URLSearchParams(window.location.search);
+    const aid = params.get('imtbl_aid');
+    if (!aid) return null;
+    params.delete('imtbl_aid');
+    const search = params.toString();
+    const newUrl = `${window.location.pathname}${search ? `?${search}` : ''}${window.location.hash}`;
+    window.history.replaceState(null, '', newUrl);
```
bkbooth: I think this should be `window.history.replaceState(window.history.state, '', newUrl);`

`null` clobbers any SPA router state. Not sure if that's intentional though.

packages/audience/sdk/src/sdk.ts
```diff
@@ -74,6 +76,18 @@ export class Audience {
 
   private destroyed = false;
 
+  private static readAndStripAidParam(): string | null {
+    if (!isBrowser()) return null;
+    const params = new URLSearchParams(window.location.search);
+    const aid = params.get('imtbl_aid');
```
bkbooth, line 84: Do we need to do any validation of `imtbl_aid`? Probably small scope for damage, but not sure if it's negligible?

bkbooth (approved):

PR #2892, opened 2026-07-01
feat(audience-core): wire up GPC/DNT compliance in consent manager

nattb8:
The original `detectDoNotTrack()` had a broken ternary: both branches resolved to `'none'`, so GPC and DNT signals had no effect. This PR fixes the wiring and enforces privacy signals at every consent decision point.

**Core (`@imtbl/audience-core`)**
- `createConsentManager` forces consent to `'none'` at init if GPC or DNT is active, regardless of the configured `initialLevel`
- `setLevel` re-checks on every call and caps any attempted upgrade to `'none'` while either signal is active
- GPC is evaluated before DNT because it carries legal weight under CCPA/CPRA

**Web SDK (`@imtbl/audience`)**
- `setConsent` caps the incoming level before running side effects (cookie creation, queue start), not after — prevents tracking infrastructure spinning up for a level GPC would block
- Fires `gpc_consent_overridden` to New Relic via `@imtbl/metrics` with `{ signal, requestedLevel, context }`
  - `context: 'init'` when the configured level is overridden at SDK construction
  - `context: 'runtime'` on subsequent `setConsent` calls
- Logs a `console.warn` to the customer when `debug: true`

**Pixel (`@imtbl/pixel`)**
- No internal telemetry (pixel stubs out `@imtbl/metrics` to keep bundle size down)
- Adds `debug?: boolean` (default `false`) to `PixelInitOptions`: logs a `console.warn` if GPC/DNT is active and a non-`'none'` consent level was passed explicitly
- Warning is suppressed in auto-consent mode since auto mode already resolves to `'none'` independent of GPC

### DIFF
--- a/packages/audience/core/src/consent.test.ts
+++ b/packages/audience/core/src/consent.test.ts
@@ -1,8 +1,13 @@
+import { track } from '@imtbl/metrics';
 import {
-  createConsentManager, canTrack, canIdentify,
+  createConsentManager, canTrack, canIdentify, detectDoNotTrack,
 } from './consent';
 import type { HttpSend } from './transport';
 import { TransportError } from './errors';
+
+jest.mock('@imtbl/metrics', () => ({
+  track: jest.fn(),
+}));
 
 function createMockSend() {
   return jest.fn<ReturnType<HttpSend>, Parameters<HttpSend>>().mockResolvedValue({ ok: true });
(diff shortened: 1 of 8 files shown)


JCSanPedro (dismissed):

JCSanPedro (dismissed):

nattb8:
Superseded by new PR on renamed branch.

PR #2894, opened 2026-07-01
feat(audience): stamp consentLevel on pixel and web SDK messages [SDK-565]

JCSanPedro:
## What

- Adds an optional `consentLevel` field to the shared `BaseMessage` type in `@imtbl/audience-core`.
- **Pixel**: stamps the current consent level onto every message in `Pixel.buildBase()`.
- **Web SDK**: stamps the current consent level onto every message in `Audience.baseMessage()`.
- On a `full → anonymous` consent downgrade, queued messages have their stamped `consentLevel` rewritten to `anonymous` (alongside the existing `userId` strip), so no queued event still reports `full` after the user withdraws consent.

## Why

Consent level is being plumbed end-to-end through the pipeline (SDK → audience ingest → attribution → postbacks) as part of [SDK-449](https://linear.app/imtbl/issue/SDK-449). Attribution currently *infers* consent from the presence of a `userId`, which cannot distinguish **full-but-unidentified** traffic (full consent granted before/without `identify()`, e.g. all pixel traffic) from genuinely anonymous traffic. Sending the decided level explicitly lets the backend record provenance rather than guess.

The audience-service ingest/derivation side ([SDK-554](https://linear.app/imtbl/issue/SDK-554)) has already landed and accepts the field as optional, so this is a non-breaking, forward-compatible change.

Both client surfaces are included together because the change touches the shared `@imtbl/audience-core` (the `BaseMessage` type and the shared consent-downgrade transform), so shipping them in one PR keeps core and its consumers consistent.

Notes:
- Every enqueue path is consent-guarded (pixel `buildBase`, web `baseMessage` via `isTrackingDisabled`/`canIdentify`), so the stamped level is always a decided, sendable value (`anonymous`/`full`) — never `none`/`not_set`.
- The pixel never calls `identify()`, so it emits `full` consent with **no** `userId` — exactly the case the explicit field exists to capture. The web SDK carries `userId` once `identify()` is called under `full` consent.

## Tests

- `pixel.test.ts`: stamps `anonymous` on all emitted messages; stamps `full` with `userId` undefined; reflects an upgraded level on later events.
- `sdk.test.ts`: stamps `anonymous` on all emitted messages; stamps `full` carrying `userId` once identified; downgrade from `full → anonymous` rewrites `consentLevel` to `anonymous` and strips `userId`.
- `consent.test.ts`: downgrade rewrites `consentLevel` to `anonymous` (including messages that lacked one, e.g. restored from a pre-`consentLevel` build).

### DIFF
--- a/packages/audience/core/src/consent.ts
+++ b/packages/audience/core/src/consent.ts
@@ -85,14 +85,16 @@ export function createConsentManager(
           // Purge all queued messages
           queue.purge(() => true);
         } else if (next === 'anonymous') {
-          // Remove identify/alias messages and strip userId from the rest
+          // Remove identify/alias messages, strip userId from the rest, and
+          // downgrade the stamped consentLevel.
           queue.purge((msg: Message) => msg.type === 'identify' || msg.type === 'alias');
           queue.transform((msg: Message) => {
-            if ('userId' in msg) {
-              const { userId, ...rest } = msg;
+            const downgraded = { ...msg, consentLevel: 'anonymous' as const } as Message;
+            if ('userId' in downgraded) {
+              const { userId, ...rest } = downgraded;
               return rest as Message;
             }
-            return msg;
+            return downgraded;
           });
         }
       }
--- a/packages/audience/core/src/consent.ts
+++ b/packages/audience/core/src/consent.ts
@@ -86,10 +86,15 @@ export function createConsentManager(
           queue.purge(() => true);
         } else if (next === 'anonymous') {
           // Remove identify/alias messages, strip userId from the rest, and
-          // downgrade the stamped consentLevel.
+          // downgrade the stamped consentLevel. The rewrite only touches
+          // messages that already carry a consentLevel (the pixel stamps every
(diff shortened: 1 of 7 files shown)


packages/audience/core/src/consent.ts
```diff
@@ -121,13 +121,17 @@ export function createConsentManager(
...
           queue.purge(() => true);
         } else if (effective === 'anonymous') {
+          // Remove identify/alias messages, strip userId from the rest, and
+          // downgrade the stamped consentLevel to 'anonymous' — the user
+          // withdrew full consent, so queued events must not still report 'full'.
           queue.purge((msg: Message) => msg.type === 'identify' || msg.type === 'alias');
           queue.transform((msg: Message) => {
-            if ('userId' in msg) {
-              const { userId, ...rest } = msg;
+            const downgraded = { ...msg, consentLevel: 'anonymous' as const } as Message;
```
nattb8: `as Message` looks avoidable

nattb8 (dismissed):

nattb8 (approved):

PR #2896, opened 2026-07-08
feat(audience): avoid mutating or purging queued messages on consent change [SDK-647]

JCSanPedro:
## Summary

Stops the web SDK and pixel from retro-actively mutating already-queued events when the client consent level changes. Under GDPR the applicable consent is the one in force at capture time, so events keep the `consentLevel`/`userId` they were recorded with; consent now only gates what is collected going forward.

Both surfaces share the consent state machine in `@imtbl/audience-core`, so the fix lands in one place.

- `core/src/consent.ts`: removed the downgrade block in `setLevel`. A consent change no longer:
  - purges the entire queue (downgrade to `none`),
  - purges `identify`/`alias` messages, or
  - rewrites `consentLevel` to `anonymous` / strips `userId` (downgrade to `anonymous`).
  Consent changes now only update the current level and notify the backend. Also dropped the now-dead `queue` parameter (and unused imports) from `createConsentManager`.
- `core/src/queue.ts`: removed `purge()` and `transform()` from `MessageQueue` - they existed solely for consent-driven mutation and have no remaining callers.
- Updated the two `createConsentManager` call sites (`sdk.ts`, `pixel.ts`).
- Updated unit/integration tests and the pixel/sample-app docs describing the old behavior.

## Context

[SDK-647](https://linear.app/imtbl/issue/SDK-647/avoid-mutating-consent-level-in-client-side-message-queues). Complements the consent-level provenance pipeline (SDK-449/SDK-565): each message already carries the consent level it was captured under, which the client-side rewrite was destroying.

## Behavioral note for reviewers

`setConsent('none')` still stops the queue and deletes cookies, but no longer clears the queue. Events captured before the downgrade now persist (and flush if consent is later re-enabled) instead of being dropped. This is the intended data-retention change - flagging for privacy visibility.

Unity is intentionally out of scope for this PR (web + pixel first).

## Test plan

- [x] `core` unit tests (178 passing), incl. rewritten consent tests
- [x] `sdk` tests (86 passing) - integration tests now assert queued events retain capture-time `consentLevel` + `userId` across a downgrade
- [x] `pixel` tests (120 passing)
- [x] `eslint --max-warnings=0` + `tsc` typecheck clean on all three packages

### DIFF
--- a/packages/audience/core/src/queue.ts
+++ b/packages/audience/core/src/queue.ts
@@ -207,18 +207,6 @@ export class MessageQueue {
     this.persist();
   }
 
-  /** Remove all messages matching a predicate. */
-  purge(predicate: (msg: Message) => boolean): void {
-    this.messages = this.messages.filter((m) => !predicate(m));
-    this.persist();
-  }
-
-  /** Transform messages in place (e.g., strip userId on consent downgrade). */
-  transform(fn: (msg: Message) => Message): void {
-    this.messages = this.messages.map(fn);
-    this.persist();
-  }
-
   get length(): number {
     return this.messages.length;
   }
--- a/packages/audience/core/src/consent.test.ts
+++ b/packages/audience/core/src/consent.test.ts
@@ -13,21 +13,6 @@ function createMockSend() {
   return jest.fn<ReturnType<HttpSend>, Parameters<HttpSend>>().mockResolvedValue({ ok: true });
 }
 
-function createMockQueue() {
-  return {
-    purge: jest.fn(),
-    transform: jest.fn(),
-    enqueue: jest.fn(),
-    flush: jest.fn(),
-    flushUnload: jest.fn(),
-    start: jest.fn(),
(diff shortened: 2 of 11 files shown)


packages/audience/core/src/queue.ts
```diff
@@ -207,18 +207,6 @@ export class MessageQueue {
...
   }
 
-  /** Remove all messages matching a predicate. */
-  purge(predicate: (msg: Message) => boolean): void {
-    this.messages = this.messages.filter((m) => !predicate(m));
-    this.persist();
-  }
-
-  /** Transform messages in place (e.g., strip userId on consent downgrade). */
-  transform(fn: (msg: Message) => Message): void {
```
JCSanPedro, line 217: Removing these because they will become unused, and also mutating existing events is something we will generally avoid.

The above `purge` might be useful for a different case but will add back in if needed.

nattb8 (approved):

PR #2899, opened 2026-07-09
feat(audience): warn and drop identify()/alias() calls with a malformed passport id

nattb8:
## Summary
Studios sometimes call `identify()` or `alias()` with `identityType: 'passport'` but pass their own internal user id instead of a real Passport id, and today nothing catches it, so bad data silently flows into the audience pipeline. This rejects the call (no event sent, `userId` untouched) and warns unconditionally (not gated behind `debug`) when the id doesn't match the shape Passport actually issues (`connection|id` or a bare UUID), so studios see the mistake immediately during local testing instead of finding it in the data warehouse later.

## Test plan
- [x] Unit tests for `isPassportIdValid` (core) covering valid pipe-form ids, valid UUIDs, whitespace padding, and invalid shapes
- [x] Unit tests for `identify()`/`alias()`: malformed passport id warns and drops the call; warning fires without `debug` enabled; valid ids and non-passport identity types never warn
- [x] Unit tests confirming a whitespace-padded valid passport id is stored/sent trimmed, not with the padding intact (the id that's validated and the id that's stored/sent must match)
- [x] Sample-app UI affordance ("Fill invalid Passport ID example") and a fix so the sample app's success/drop mirroring reflects the real SDK outcome instead of always showing "ok"
- [x] `pnpm jest` passes for `packages/audience/core` and `packages/audience/sdk`
- [x] `eslint` passes on all changed files
- [x] `tsc --noEmit` passes for `packages/audience/sdk`

### DIFF
--- a/packages/audience/core/src/validation.ts
+++ b/packages/audience/core/src/validation.ts
@@ -16,6 +16,18 @@ export function isTimestampValid(eventTimestamp: string): boolean {
   return ts <= now + MAX_FUTURE_MS && ts >= now - MAX_PAST_MS;
 }
 
+const PASSPORT_ID_RE = /^[^|]+\|[^|]+$/;
+const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+
+/**
+ * Validate that an ID looks like a Passport ID: either `<connection>|<id>`
+ * (e.g. `email|abc123`, `google-oauth2|456`) or a bare UUID.
+ */
+export function isPassportIdValid(id: string): boolean {
+  const trimmed = id.trim();
+  return PASSPORT_ID_RE.test(trimmed) || UUID_RE.test(trimmed);
+}
+
 /**
  * Validate that alias from and to are not the same identity.
  */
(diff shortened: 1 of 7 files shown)


shineli1984 (approved):

packages/audience/core/src/validation.ts
```diff
@@ -16,6 +16,18 @@ export function isTimestampValid(eventTimestamp: string): boolean {
...
 }
 
+const PASSPORT_ID_RE = /^[^|]+\|[^|]+$/;
+const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+
+/**
+ * Validate that an ID looks like a Passport ID: either `<connection>|<id>`
+ * (e.g. `email|abc123`, `google-oauth2|456`) or a bare UUID.
+ */
+export function isPassportIdValid(id: string): boolean {
```
JCSanPedro, line 26: Something we should keep in mind, when we migrate from Auth0 -> Ory Kratos, the user id format will change. Will become UUIDs I think.

Hmm actually, probably need to consult with @brayansdt the implications of some of these things when we migrate.

PR #2919, opened 2026-07-26
feat(audience): add paid traffic network detection [SDK-710]

arsenal942:
https://linear.app/imtbl/issue/SDK-710/extract-metatiktok-paid-traffic-detection-into-ts-immutable-sdk

Extracts the hardcoded Meta/TikTok UTM and click-ID checks from play's pixel.ts into a shared getAttributionNetwork() classifier, plus isPaidMeta/isPaidTikTok/isPaidGoogle/isPaidReddit/isPaidX helpers. Mirrors platform-services' server-side attribution network taxonomy so client- and server-classified traffic agree on the same names. Also adds twclid to the tracked click-ID params, since X had no click-ID fallback (the same gap TikTok had before ttclid was already tracked).

### DIFF
--- a/packages/audience/core/src/attribution.ts
+++ b/packages/audience/core/src/attribution.ts
@@ -14,6 +14,7 @@ const CLICK_ID_PARAMS = [
   'rdt_cid',
   'msclkid',
   'li_fat_id',
+  'twclid',
 ] as const;
 
 const STORAGE_KEY = '__imtbl_attribution';
@@ -31,6 +32,7 @@ export interface Attribution {
   rdt_cid?: string;
   msclkid?: string;
   li_fat_id?: string;
+  twclid?: string;
   referral_code?: string;
   referrer?: string;
   landing_page?: string;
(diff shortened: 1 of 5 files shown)


packages/audience/core/src/attribution.ts
```diff
@@ -31,6 +32,7 @@ export interface Attribution {
   rdt_cid?: string;
   msclkid?: string;
   li_fat_id?: string;
+  twclid?: string;
```
nattb8, line 35: Need to add to docs too https://docs.immutable.com/docs/products/audience/data-dictionary#attribution-signals

nattb8 (dismissed):

packages/audience/core/src/attribution.ts
arsenal942: @nattb8 Do I need to do this? If so, is that all in this repo?

JCSanPedro (approved):

packages/audience/core/src/attribution.ts
arsenal942: @nattb8 I updated readme.

nattb8 (approved):

packages/audience/core/src/attribution.ts
nattb8: https://github.com/immutable/documentation/pull/142

PR #2932, opened 2026-07-29
feat(audience): add trackConversion for pixel/CAPI dedup

JCSanPedro:
## Summary

Adds a conversion event id primitive to `@imtbl/audience` so a browser pixel event and its server-side Conversions API (CAPI) counterpart can be deduplicated on a **shared id**. Today Game Page fires browser pixels client-side and postbacks fires CAPI events server-side for the same `sign_up`, with no common id, so ad networks double-count the conversion.

This is piece 1 of 3 (SDK-719). Follow-ups: SDK-720 (Play fires vendor pixels with the shared id) and SDK-721 (postbacks reads the shared id + event-name mapping).

## Changes

- **`Audience.trackConversion(event, properties?)`** - behaves like `track()`, and additionally: classifies the visit's network from the session-cached first-touch attribution (the same signals that drive server-side attribution), mints a shared id for dedup-capable networks (Meta, TikTok, Reddit), stamps it onto the event under the reserved `_imtbl_conversion_id` / `_imtbl_conversion_network` properties, and returns `{ eventId, network }`. The caller passes `eventId` to the browser pixel (e.g. Meta `fbq('track', 'CompleteRegistration', {}, { eventID })`).
- **`Audience.getAttributionNetwork()`** instance method - classifies from the client's cached snapshot so client- and server-side attribution agree even after the landing URL's query params are gone.
- **Standalone `getAttributionNetwork(attribution?)`** - retained and made optional-arg: classifies from a passed snapshot, else falls back to the live URL, for consumers that have a pixel but no SDK instance.
- **Reserved property contract** in `sdk/src/conversion.ts` (`_imtbl_conversion_id`, `_imtbl_conversion_network`, `DEDUP_CAPABLE_NETWORKS`). These are a cross-service contract with the postbacks builders (SDK-721).

### Breaking change

Removes `isPaidMeta` / `isPaidTikTok` / `isPaidGoogle` / `isPaidReddit` / `isPaidX`. The only consumer is Game Page (Play), which we control and which is updated in SDK-720 to use `getAttributionNetwork()`.

## Test plan

- [x] `packages/audience/core` unit tests pass (attribution classification, snapshot + live-URL paths)
- [x] `packages/audience/sdk` unit tests pass (trackConversion id minting, reserved props, consent gating, instance `getAttributionNetwork`)
- [x] Typecheck passes (incl. `sdk.test-d.ts` compile-time assertions)
- [ ] Validated end-to-end once SDK-720 (Play) consumes the prerelease alpha build

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -68,7 +68,9 @@
       "!packages/game-bridge",
       "!@imtbl-alpha/**"
     ],
-    "releaseTagPattern": "{version}",
+    "releaseTag": {
+      "pattern": "{version}"
+    },
     "changelog": {
       "workspaceChangelog": {
         "file": false,
--- a/packages/audience/core/src/attribution.test.ts
+++ b/packages/audience/core/src/attribution.test.ts
@@ -3,11 +3,6 @@ import {
   collectPageAttribution,
   clearAttribution,
   getAttributionNetwork,
-  isPaidMeta,
-  isPaidTikTok,
-  isPaidGoogle,
-  isPaidReddit,
-  isPaidX,
 } from './attribution';
 
 const STORAGE_KEY = '__imtbl_attribution';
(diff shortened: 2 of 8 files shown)


nx.json
```diff
@@ -68,7 +68,9 @@
       "!packages/game-bridge",
       "!@imtbl-alpha/**"
     ],
-    "releaseTagPattern": "{version}",
+    "releaseTag": {
+      "pattern": "{version}"
+    },
```
JCSanPedro, line 73: fixes pre release publish

nattb8 (approved):

PR #2946, opened 2026-08-12
fix(auth): retry silent token refresh on transient failures

allan-almeida-imtbl:
**TL;DR:** Illuvium players getting 401s mid-session isn't a CORS or backend problem — it's the Passport SDK destroying valid sessions when a token refresh hiccups. Tracked in [BLOC-522](https://linear.app/imtbl/issue/BLOC-522/passport-silent-token-refresh-logs-players-out-on-transient-failures).

## Symptom

Logged-in players in Unreal games start getting 401 Unauthorized mid-session, clustered in EU afternoon (peak) hours. Client logs show a CORS error on `counterfactual-address` — that's a red herring: it's the re-login cascade after the session dies, not the cause. (`Origin: null` is normal for the Unreal SDK's embedded browser bridge, and the endpoint's CORS config is fine — verified it returns 204 + `allow-origin: *`.)

## Root cause

Since ts-immutable-sdk **v1.76.8** (#2479, Dec 2024), a failed silent token refresh against Auth0 **deletes the stored user by default** — only timeouts are exempt, no retry, no backoff. So one transient failure (429 rate limit, 5xx, network blip) mid-session = permanent logout, even though the player's refresh token is still valid. On Unreal this ships in the bundled game-bridge from **unreal-immutable-sdk v1.5.0** (Feb 2025) onward.

The EU-afternoon clustering fits peak concurrency pushing refresh volume into Auth0 tenant rate limits — each throttled refresh becomes a logout instead of a retry. (Confirmation pending: Auth0 tenant logs, `fertft`/`api_limit` events during that window.)

## Fix

Mirrors the retry treatment #2931 gave the login token exchange, applied to the silent-refresh path (`refreshTokenAndUpdatePromise`):

- `signinSilentWithRetry` retries transient failures up to 3 times (4 attempts over ~3–6s of jittered backoff) with full-jitter backoff; tracks `silentRefreshRecovered` when a retry saves a session
- The stored user is only removed on a definitive OAuth rejection — an allowlist of codes from [RFC 6749 §5.2](https://datatracker.ietf.org/doc/html/rfc6749#section-5.2) and [OIDC Core §3.1.2.6](https://openid.net/specs/openid-connect-core-1_0.html#AuthError) (`invalid_grant`, `login_required`, …). Those are removed immediately with no retry — unchanged semantics, so the invalid-refresh-token hammering #2479 guarded against stays fixed
- Everything else is transient and **keeps the user** on retry exhaustion, so the next `getAccessToken` call tries again and the session self-heals once the outage/rate limit clears. Notably this includes rate-limit responses (`too_many_requests`) — a subtlety because oidc-client-ts throws `ErrorResponse` for any non-OK token response with an `error` body field (429s and 5xxs included), so a transient-code denylist would have misclassified the suspected production trigger as permanent

## Test plan

- [x] `Auth.test.ts`: 21/21 passing — retry-then-recover; retry-exhaustion keeping the user for network / `server_error` / `too_many_requests` / unknown errors; no-retry + immediate removal for `invalid_grant` and `login_required`
- [ ] Rebuild game-bridge and verify on Unreal (follow-up: cut alpha for the affected customer)

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -62,7 +62,12 @@ const parseJsonSafely = (text: string): unknown => {
 // from the token endpoint) would otherwise destroy a session backed by a still-valid
 // refresh token. Retry with jittered backoff, and only remove the stored user when the
 // authorization server definitively rejects the refresh token.
-const SILENT_REFRESH_MAX_RETRIES = 2;
+// 4 attempts spread over ~3-6s of jittered backoff. Kept deliberately short: the
+// refresh blocks getAccessToken callers, and rate-limit windows are per-second, so
+// ~1s spacing already lands retries in a fresh bucket. Outages longer than this are
+// covered across cycles — exhaustion keeps the user, so the next getAccessToken
+// call starts a new cycle.
+const SILENT_REFRESH_MAX_RETRIES = 3;
 const SILENT_REFRESH_RETRY_DELAY_MS = 1000;
 
 function refreshRetryDelay(ms: number): Promise<void> {
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -870,6 +870,13 @@ export class Auth {
             errorMessage = `${errorMessage}: ${err}`;
           }
 
+          // Terminal refresh failures were previously invisible (client-side
+          // logger.warn only), which made fleet-wide incidents impossible to see.
+          trackError('passport', 'silentRefresh', err instanceof Error ? err : new Error(errorMessage), {
+            userRemoved: removeUser,
+            ...(err instanceof ErrorResponse && err.error ? { oauthErrorCode: err.error } : {}),
+          });
+
           if (removeUser) {
             // Emit USER_REMOVED event BEFORE removing user so consumers can react
             // (e.g., auth-next-client can clear the NextAuth session)
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -58,6 +58,34 @@ const parseJsonSafely = (text: string): unknown => {
   }
 };
 
+// A transient failure on the silent refresh (network error, timeout, 5xx or rate limit
+// from the token endpoint) would otherwise destroy a session backed by a still-valid
+// refresh token. Retry with jittered backoff, and only remove the stored user when the
+// authorization server definitively rejects the refresh token.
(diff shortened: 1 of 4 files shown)


packages/auth/src/Auth.ts
```diff
@@ -751,6 +808,26 @@ export class Auth {
...
+  private async signinSilentWithRetry(): Promise<OidcUser | null> {
+    const attempt = async (attemptNumber: number): Promise<OidcUser | null> => {
+      try {
+        const oidcUser = await this.userManager.signinSilent();
+        if (attemptNumber > 1) {
+          track('passport', 'silentRefreshRecovered', { attempt: attemptNumber });
+        }
+        return oidcUser;
+      } catch (error) {
+        if (isPermanentRefreshError(error) || attemptNumber > SILENT_REFRESH_MAX_RETRIES) {
```
ermyas: When silent login runs with no stored session, does it still go through the new retry path? That failure comes back as a plain Error and so wouldn't be identified as permanent refresh error? Should it fail immediately instead of being treated as transient?

packages/auth/src/Auth.ts
```diff
@@ -58,6 +58,63 @@ const parseJsonSafely = (text: string): unknown => {
...
+    setTimeout(resolve, ms);
+  });
+}
+
+// Full jitter on the backoff so that when the token endpoint fails for many clients at
+// once (peak-hour rate limiting), their retries spread out instead of synchronising.
+function refreshBackoffWithJitter(attemptNumber: number): number {
+  const base = SILENT_REFRESH_RETRY_DELAY_MS * attemptNumber;
+  return base * (0.5 + Math.random() * 0.5);
+}
```
ermyas: nit: guess these are identical to delay and backOffWithJitter in standalone.ts? any reason not to reuse?

packages/auth/src/Auth.ts
allan-almeida-imtbl: thanks, now updated to check for a stored user once at the start of the cycle and only enters the retry loop when a `refresh_token` exists

No refresh token → single attempt, fail fast, same as pre-PR behaviour for that branch

packages/auth/src/Auth.ts
allan-almeida-imtbl: no reason - refactored to be re-used across standalone.ts, thanks

ermyas (approved):

PR #2950, opened 2026-08-25
feat(auth): send OIDC login_hint for direct email login (PLT-1666)

brayansdt:
## What
On the hosted Ory login popup, after the user enters their email in the embedded prompt the popup **restarts at step 0** instead of resuming at the OTP screen (Auth0 opened straight on the code screen). [PLT-1666](https://linear.app/imtbl/issue/PLT-1666).

Root cause: the SDK forwards the email as the **custom** `email` query param on `/authorize`. On the Ory/Hydra path Hydra does not relay custom params to the login UI, so the branded page only ever sees `login_challenge` and starts over.

## Change
Additionally set the **standard OIDC `login_hint`** (= the email) on the popup `/authorize`, alongside the existing `email`/`direct` params. Hydra **does** relay `login_hint` onto the login request, and Kratos surfaces it on the login flow, so the branded page can pre-fill + auto-send the code. Mirrored in the standalone/NextAuth popup builder.

- `Auth.ts` `buildExtraQueryParams` (main popup path)
- `login/standalone.ts` `buildAuthorizationUrl` (standalone path)

Additive and Auth0-safe: the Auth0 path still consumes `email`; `login_hint` is just an extra standard param it ignores/pre-fills harmlessly.

## Paired change
Consumed by the branded page in passport-login (reads `oauth2_login_request.oidc_context.login_hint` off the flow and auto-sends the code). This SDK change is a no-op until that lands, and vice-versa.

## Scope
Email one-time-code method only. Social method resume (no OIDC-standard method hint through Hydra) is a separate follow-up.

## Tests
`Auth.test.ts` — new case asserts `login_hint` is forwarded for direct email login. Full suite green (20/20); `tsc`/eslint clean.

### DIFF
--- a/packages/auth/src/Auth.test.ts
+++ b/packages/auth/src/Auth.test.ts
@@ -504,6 +504,45 @@ describe('Auth', () => {
       );
     });
 
+    it('forwards the email as login_hint for direct email login (PLT-1666)', async () => {
+      const mockOidcUser = {
+        id_token: 'token',
+        access_token: 'access',
+        refresh_token: 'refresh',
+        expired: false,
+        profile: { sub: 'user-123', email: 'test@example.com', nickname: 'tester' },
+      };
+
+      (decodeJwtPayload as jest.Mock).mockReturnValue({
+        username: 'username123',
+        passport: undefined,
+      });
+
+      mockUserManager.signinPopup.mockResolvedValue(mockOidcUser);
+
+      const auth = Object.create(Auth.prototype) as Auth;
+      (auth as any).userManager = mockUserManager;
+      (auth as any).config = {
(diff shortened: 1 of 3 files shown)


brayansdt:
Closing in favour of an SDK-free fix. Rather than change the SDK (which needs every game integration to upgrade), we promote the SDK's custom `email` param to the standard OIDC `login_hint` at the CDN — in the passport-auth-route CloudFront Function on /authorize — matching that function's existing 'translate at the CDN so SDKs stay unchanged' design. The branded-page side (passport-login#237) is unchanged; it reads oauth2_login_request.oidc_context.login_hint either way. Infra PR to follow.

